PR #24, opened 2015-10-22
Added instructions for keyboard switching between editors

damiensawyer:
As Q&A'd at http://stackoverflow.com/questions/33271488/shortcut-key-for-switching-between-split-panes/33276179#33276179


gregvanl:
Thanks

PR #27, opened 2015-10-27
add quick-run snippet for oh-my-fish shell

Twista:


gregvanl:
Thank you for your suggestion but we're striving to keep the documentation as brief as possible.  If we see the oh-my-fish shell gaining popularity, we can revisit adding this.

PR #28, opened 2015-10-27
Simplify npm install command

scottaddie:
The "Yo Code - Adding Customizations for VS Code" section of the docs shows 2 separate `npm install` commands. This can be simplified by combining into a 1-liner.


gregvanl:
Thanks for your suggestion but we'll leave these as two commands for now.  We may add a little more description to call out that you need to install yo and also our generator.

PR #46, opened 2015-11-20
Improve Linux setup script to run in bg and hide output

Tyriar:
This enables the terminal to continue to be used alongside the instance(s) of
vscode.

### DIFF
--- a/docs/editor/setup.md
+++ b/docs/editor/setup.md
@@ -29,12 +29,21 @@ Now, you can simply type `code .` in any folder to start editing files in that f
 2. Make a new folder and extract `VSCode-linux-x64.zip` inside that folder.
 3. Double click on `Code` to run Visual Studio Code.
 
->**Tip:** If you want to run VS Code from the terminal, create the following link substituting `/path/to/vscode/Code` with the absolute path to the `Code` executable
+>**Tip:** If you want to run VS Code from the terminal, run the following to setup a script in `/usr/local/bin` to run Visual Studio Code in the background and hide its output. Replace `/path/to/vscode/Code` with the actual path to the executable.
+
+```
+VSCODE_PATH=/path/to/vscode/Code
+echo 'nohup '$VSCODE_PATH' "$@" > ~/.vscode/nohup.out 2>&1 &' | sudo tee /usr/local/bin/code > /dev/null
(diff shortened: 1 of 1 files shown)


joaomoreno:
Looks awesome. We're trying to move away from recommending people to write these things though and just do it ourselves (e.g. Code puts the launcher in the PATH somehow), though... so I'm more willing to let this stay as is until we do so.

Will reject this, but your intention is good and I'm sure we can use this knowledge to build the proper solution.

Tyriar:
That's fine, this could turn off Linux users from using vscode who don't know how to setup such a script as `Code` makes the terminal unable to used until it is closed and `Code &` will still pipe all output to the terminal, making it much less usable.

Something like this `nohup /path/to/Code "$@" > ~/.vscode/nohup.out 2>&1 &` would be a good starting point to put in the bin when an ubuntu package gets created.

docs/editor/setup.md
```diff
@@ -29,13 +29,22 @@ Now, you can simply type `code .` in any folder to start editing files in that f
 2. Make a new folder and extract `VSCode-linux-x64.zip` inside that folder.
 3. Double click on `Code` to run Visual Studio Code.
 
->**Tip:** If you want to run VS Code from the terminal, create the following link substituting `/path/to/vscode/Code` with the absolute path to the `Code` executable
+>**Tip:** If you want to run VS Code from the terminal, run the following to setup a script in `/usr/local/bin` to run Visual Studio Code in the background and hide its output. Replace `/path/to/vscode/Code` with the actual path to the executable.
 
 ```
-sudo ln -s /path/to/vscode/Code /usr/local/bin/code
+VSCODE_PATH=/path/to/vscode/Code
+echo 'nohup '$VSCODE_PATH' "$@" > ~/.vscode/nohup.out 2>&1 &' | sudo tee /usr/local/bin/code > /dev/null
```
hashhar: This works better and is simpler and can be added to bashrc or a standalone shellscript.

function code()
{
    nohup /opt/VSCode/Code "$@" > /dev/null 2>&1 &
    disown
}

Tyriar: Hey @hashnar, there will likely be a deb package as part of the next release. Auto updating and rpm package to follow. 

This part of the documentation will be removed soon as a result.

PR #54, opened 2015-11-23
Add GitHub Flavored Markdown FAQ

Beman:
It would be better to ask "What flavors of Markdown do you support?", but I don't know the answer to that.

--Beman Dawes, Boost.org


gregvanl:
Thanks.  I'm making a few other edits to the Markdown FAQ and I'll add yours.

PR #59, opened 2015-11-28
Fix src and lib directory verbiage

scottaddie:
The explanation of the code snippet in the "Run Babel inside VS Code" section is misleading. The text makes it seem as though there's a directory named "lib-directory" and another one named "src-directory".


gregvanl:
Thanks for keeping our docs clear and honest.

PR #65, opened 2015-12-02
ERROR:OmniSharp.Dnx.DnxPaths

GuidC0DE:
Added instruction to solve ERROR:OmniSharp.Dnx.DnxPaths in OmniSharp Log for OS X

### DIFF
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -104,6 +104,11 @@ brew update
 brew reinstall mono
 ```
 
+## OS X - ERROR:OmniSharp.Dnx.DnxPaths in OmniSharp Log
+
+Install DNX from homebrew, follow these instructions:
+https://github.com/aspnet/homebrew-dnx
+
 ## Linux - Azure VM Issues
 I getting a "Running without the SUID sandbox" error?
 


gregvanl:
Thank you for submitting a pull request.  I'm not familiar with the error and the instructions are a little too general.  Can you provide more information here to explain the issue and how your suggestion solves it?

GuidC0DE:
When I was setting up environment to develop Unity3D applications with VS Code, I got some errors from OmniSharp and intelli sense did not work. One of the error in OmniSharp Log was about DnxPaths, so I thought that maybe it will be good to have some info about this issue in one place here. But later after fixing another errors I found that DnxPaths does not affect intelli sense (previously I thought it does). It seems OmniSharp just wants to say that it will not load DNX assemblies that needed to develop ASP .NET 5 applications. But it is not required for Unity3D applications or some other .NET applications. Maybe OmniSharp should not report it as error, but as warning. So, as you said, it is not for general FAQ. So, now I think, it should not be included here. Sorry for this pull request.

PS: In the end I've found that my real problem was in OmniSharp engine that was unable to parse line of C# code that has correct syntax and compiled with no errors by mono.

gregvanl:
Thanks for the explanation.  We really appreciate people contributing to our documentation, so feel free to submit pull requests for doc bugs and improvements.

PR #91, opened 2015-12-21
Add instructions for disabling word wrap

bob-white:
Noticed that -1 actually disabled word wrap, thought it might be useful to actually inform others.
Should probably update the default settings file as well.


gregvanl:
Thank you for your pull request.  The markdown file userandworkspace.md just takes a snapshot of the default user settings so this change would need to be made in the VS Code sources.  You can open a issue or pull request against https://github.com/Microsoft/vscode and next release the markdown will be updated.  The VS Code dev team may not want this setting value publicized if it is just an implementation detail.

PR #94, opened 2015-12-22
Modified command line arguments

mioae:
Prefixed the command line arguments with a dot, '.', to open the file in
the root folder context. Improves usability by opening the root folder
instead of just the file.


gregvanl:
Thank you for your pull request.  We'd like to leave the example command line arguments as they are since they match the Unity Preferences dialog snapshot.  Also given the -r (reuse active window) switch is also passed in and VS Code was likely open to the project folder, the user shouldn't need to also pass the current folder.

PR #98, opened 2015-12-27
Clarify documentation for Ctrl-Tab switcher

sveiss:
Clarify the documentation for the Ctrl-Tab switcher to make it clear you need to hold down Ctrl to show, then hit Tab to select and release to open. This also removes an incorrect reference to Cmd-Tab on Macs instead of Ctrl (Cmd-Tab gets you the system application switcher).


gregvanl:
Thank you for your pull request.  Your suggestion makes the sequence much clearer.

PR #99, opened 2015-12-30
Fixing docs for previous files keyboard shortcut.

marshallswain:
The ctrl key appears to be the correct key (with the tab key) for toggling between previous files, even after the list of files is showing.


gregvanl:
Thank you for your pull request.  I just merged another doc change clarifying this sequence (pull request #98).

PR #100, opened 2015-12-30
Added additional side loading instructions

wadeanderson7:
When I read this I wasn't sure what qualified as the generated output. I added a sentence describing an additional way to package a side loaded app that should help developers if they are confused.


wadeanderson7:
@gregvanl let me know if this is the correct way to suggest a change.

gregvanl:
Yep, this is how to update the documentation.

PR #107, opened 2016-01-05
Updated landing page from extensions gallery.

wadeanderson7:
Users are not sure how to install extensions when they land on this page from the gallery. This commit adds more focus to installing from the product.

### DIFF
--- a/docs/editor/extension-gallery.md
+++ b/docs/editor/extension-gallery.md
@@ -17,7 +17,7 @@ You can browse and install extensions from within VS Code. Press `kb(workbench.a
 
 Pick `Extensions: Install Extension`.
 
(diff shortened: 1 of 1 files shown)


wadeanderson7:
@chrisdias @bgashler1 @egamma @gregvanl feel free to chime in. My intent was to really focus on the in product install and remove the focus of the online marketplace. I believe this is a better user experience and I'm open to feedback.

bgashler1:
@wadeanderson7 My only feedback is that we should have a link somewhere in the top section you proposed to the extension gallery:
```
Not sure what to install?
<a href="https://marketplace.visualstudio.com/#VSCode" target="_blank">
     View the gallery of available extensions
</a> 
```

wadeanderson7:
Good suggestion @bgashler1. @gregvanl and I have made the change. Check it out now.

PR #118, opened 2016-01-18
Update setup.md

rtorr:
Adding some instructions to allow you to use the `code .` command from fish shell. I made a gist a while back, thought this might be helpful (https://gist.github.com/rtorr/233bf6ffd81b6ecdf534)


rtorr:
@gregvanl it works with fish shell. No need to re-add :)

gregvanl:
Excellent.  Thanks for checking.

PR #145, opened 2016-02-06
Use typings instead of tsd to grab d.ts files

benfahrenheit:
TSD has been deprecated (see https://github.com/DefinitelyTyped/tsd/issues/269) in favor of further development on typings. Updating the node.js sample to install and use typings instead.

TSD still works for now, but may as well be future proof :)

### DIFF
--- a/docs/runtimes/nodejs.md
+++ b/docs/runtimes/nodejs.md
@@ -61,22 +61,21 @@ Also notice that VS Code knows that `msg` is a string based on the initializatio
 
 ![Your first Node Express App](images/nodejs/stringintellisense.png)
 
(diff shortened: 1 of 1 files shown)


egamma:
Thanks for the heads-up on this deprecation and PR!

There are several other places where we refer to TSD in the docs and we should be consistent, e.g, https://code.visualstudio.com/Docs/runtimes/nodejs. We also use TSD in samples and we as a time still have tsd in our brains and hands.

@gregvanl let's make this a concerted effort on this and let's wait with adopting this PR.

benfahrenheit:
Sounds good, thanks Erich!

gregvanl:
Hi Ben, I can't merge your pull request due to conflicts but I'll take your changes manually.  I'm also updating other uses of TSD in our docs.
Thanks

gregvanl:
Added Ben changes with https://github.com/Microsoft/vscode-docs/commit/ea22387b1b73eb607489aa919c6553967fecd060

PR #157, opened 2016-02-13
Update example-word-count.md

buzzfrog:
changed to package.json where the extension should be activated.


gregvanl:
Thanks for catching this as well.

PR #159, opened 2016-02-14
changed vsce cmd 'create' to 'create-publisher'

buzzfrog:


gregvanl:
Thanks for catching this.

PR #172, opened 2016-02-24
Update README links in Chrome Debugger post

auchenberg:


gregvanl:
I've made the same change along with some edits.

PR #176, opened 2016-02-24
Updated osx bash and added nvm note for osx and nvm users

vprasanth:
The original bash info didn't work for me until I added the export part, not sure if it was just for me, but I do remember other mac users having this issue. The nvm part I feel needs to be here or somewhere in general as its pretty helpful and on of the first issues you'd experience with VS Code if you are using nvm.

### DIFF
--- a/docs/editor/setup.md
+++ b/docs/editor/setup.md
@@ -22,11 +22,13 @@ Getting up and running with VS Code is quick and easy.  Follow the platform spec
 >
 >```bash
 >function code () { VSCODE_CWD="$PWD" open -n -b "com.microsoft.VSCode" --args $*; }
+>export -f code
 >```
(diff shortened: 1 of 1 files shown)


waderyan:
This is a good suggestion, however, what needed to be done was for the terminal to be restarted. 

We'd like to keep the simplicity of one line of code for now. 

The NVM suggestion is a good one. Let's move it to Q&A portion of the docs. I'm imagining we could have a "note for X" for a bunch of different uses that wouldn't scale well.

vprasanth:
Sounds good @waderyan, Although the restart part didn't work.

waderyan:
Really? Hmm... Where are you putting the code? Are you saving it in your .bashrc? 

I run this command after saving something to .bashrc and it works great. 

`. ~/.bashrc`

vprasanth:
@waderyan You were right! PEBCAK error.

waderyan:
Glad to hear.

PR #184, opened 2016-02-26
Add "Changing Settings" section

JanKok:
In a freshly installed Code on Linux, I see no Settings icon, and settings.json is not listed under WORKING FILES. I just want to clarify how to change the settings since there are many references to changing settings later in the document.

Also made a couple other changes to make the document match the Code menus.


gregvanl:
I've taken your suggestions and added them to https://github.com/Microsoft/vscode-docs/commit/a404600224bfba8b4c927c17f41e23087a023714.
Thanks.  Let me know if you agree with the final text and flow.

PR #186, opened 2016-02-27
sudo needed on Ubuntu

JanKok:


Tyriar:
This depends on how `npm` was installed. [We advocate](https://github.com/Microsoft/vscode/wiki/How-to-Contribute#installing-prerequisites) for installing `npm` in such a way that `sudo` is not required. Not doing so will probably lead you to others problems with node modules, such as Yeoman breaking.

Thanks for the PR!

JanKok:
You may be right that sudo should not be needed for running npm. Ok, but then what about users like me who get errors when we don't use sudo? What should we do? Should we just sudo and try to move on? Or should we read through and try to follow the two pages of Installing Prerequisites that you pointed me to, and maybe also the info at http://www.johnpapa.net/how-to-use-npm-global-without-sudo-on-osx/ (and that info is for OSX, would it work for Ubuntu Linux as well?)?

The easier it is to install, learn, and use Code, the more people will like it and use it. If it's too hard to get started, people will ignore Code and continue using other, existing tools.

I would like to make the installation of Code, and whatever else is required to get started, as simple as possible. Ideally, it should take a single click.

I'd also like to make it as easy as possible to learn to use Code. I wonder if it's necessary to install Node.js, NPM, and Express in order to learn to use Code. Code seems capable of compiling - or at least interpreting - files such as settings.json or launch.json. Can the Code introductory documentation be written to use just whatever Node.js support is already built in to Code?

waderyan:
The goal is to make VS Code a great experience. Users will have a good experience when they follow good technology standards and practices. Following [johnpapa's article](http://www.johnpapa.net/how-to-use-npm-global-without-sudo-on-osx/) is an essential step for having a good experience with Node. 

You can get started with VS Code outside of the Node ecosystem. PHP, Python, ASP, and many other technology stacks are supported. The docs focus on the Node stack because this is **the best** experience with support for intellisense, debugging, the works. More support for other stacks are in the works and the documentation will follow. 

> The easier it is to install, learn, and use Code, the more people will like it and use it. If it's too hard to get started, people will ignore Code and continue using other, existing tools.I would like to make the installation of Code, and whatever else is required to get started, as simple as possible. Ideally, it should take a single click.

We share all of these sentiments and are working hard to make it a reality. Unfortunately, its not as easy as it sounds, especially in a cross platform, exponential growth of technologies world. If you have specific suggestions please elaborate. 

> Can the Code introductory documentation be written to use just whatever Node.js support is already built in to Code?

I'm not sure what you mean by this. Can you elaborate?

JanKok:
I've never done web-related development, so I never had any exposure to .json files, Node.js, NPM, or Express before I installed Code and tried to read and follow the introductory documentation.

Node.js has been a huge stumbling block for me. I've been trying to read through the Code documentation in the order that it's presented. (I'm deliberately avoiding skipping around in the documentation so I can get a sense of whether the information is presented in a good order and I can give feedback when there are problems with the documentation.) Near the beginning of the debugging page, it says:

> It is helpful to first create a sample Node.js application before reading about debugging. Follow this guide to do a run-through with Node.js:

    Node.js Applications

But when I tried to go through that guide (the nodejs page), I had several problems: see https://github.com/Microsoft/vscode-docs/issues/187 .

Now, I got Node.js and Express installed and apparently working, but you are telling me that my NPM isn't installed properly and I should follow this procedure to fix NPM... And my reaction is, "Seriously? You expect your users to plough through several pages of procedures (which may or may not work!) in order to install NPM properly, in order to install Express properly, when some of us don't even care about Node.js all that much?"

On the other hand, Code looks interesting and useful, and the nodejs documentation page looks useful for getting familiar with Code, and of course I understand that Code in in the beta stage, so I'm willing to work with you to try to find a way that newcomers can follow the instructions on that nodejs page.

So anyway, what should we do about sudo? I suggest either:
1. After the line "npm install -g express-generator" in the nodejs doc, say something like "If you get 'permission denied' errors, please click here" which opens another page that advises about the risk of doing sudo and gives instructions on how to fix the NPM installation.
2. Provide a script to fix things up, so the user doesn't have to follow pages of instructions.

Do the NPM developers, and distributors of NPM such as Ubuntu, agree with John Papa's view? If so, maybe this issue will just go away after a while.

>> Can the Code introductory documentation be written to use just whatever Node.js support is already built in to Code?

> I'm not sure what you mean by this. Can you elaborate?

Please remember I know almost nothing about Node.js, but it's my impression that Code contains a Node.js interpreter or compiler so that it can read files like settings.json. Does Code contain enough Node.js capabilities so that users could do the exercises in the nodejs documentation page, using just Code's built-in Node.js capabilities?

waderyan:
These are good suggestions. 

Ping @gregvanl. @gregvanl is working on an end to end doc story for Node and could use some of this feedback.

Tyriar:
Unfortunately the root cause of the issue is due to confusion around how node and npm should be installed. I believe there are apt packages out there as well that stick npm's package cache in questionable locations. Even npm have [an article on how to fix it](https://docs.npmjs.com/getting-started/fixing-npm-permissions). 

Our documentation largely assumes they have been installed correctly. I'd prefer not to even have a link here because:

- Telling devs to fix their npm installs every time we say install a module would be messy and difficult to maintain
- Assuming a dependency is installed correctly is a reasonable assumption 
- Most Linux/OSX developers know they can just tack sudo on the end and it will probably work, in this case express will probably work just fine. *Some* modules may have issues when operating in a directory without permissions

JanKok:
@gregvanl
When I try to follow instructions (such as "npm install -g express-generator") and don't get exactly the results that I expect, I suspect any of several causes:
1. Instructions are not accurate or don't cover all cases (in this case I assumed "sudo" was missing).
2. The command (NPM) or package, etc. are broken.
3. My machine is messed up, perhaps due to difficulties with earlier instructions.
Adding a line like "If you get 'permission denied' errors, please click here" would immediately narrow down the problem, and also restore confidence that the instructions are accurate.

> Assuming a dependency is installed correctly is a reasonable assumption

Well, NPM wasn't installed properly on my Ubuntu system, and I believe this is one of the most popular, if not THE most popular Linux distro. And there are several articles about the incorrectly-installed NPM problem. So in this case it seems the assumption is not correct.

> Telling devs to fix their npm installs every time we say install a module would be messy and difficult to maintain

People who have NPM installed correctly can quickly skip over the instruction  "If you get 'permission denied' errors, please click here". But people who don't have it installed correctly should be informed about the situation, including "[you] can just tack sudo on the end and it will probably work, in this case express will probably work just fine. Some modules [such as Yeoman] may have issues when operating in a directory without permissions" and then give instructions for how to repair the installation.

The "if you get errors" message should go after the first npm command in the nodejs page, and perhaps in some other key places where npm is mentioned. I don't think it has to be said more than once per page, and perhaps not in in-depth pages that beginners are unlikely to wander into.

Tyriar:
The main point I was trying to make was that Code should work just fine here without `sudo`, and that the vast majority of developers running Linux know they can add `sudo` to run with elevated permissions when they encounter the error in question.

Issue #189, opened 2016-02-27
Errors when starting Code


JanKok:
I get these errors each time I start code from xterm in Ubuntu:

jan@i5:~/GM/mgm$ code .
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell

Other programs don't do that, so I assume this is a problem with Code. Would someone please clean up these warnings/errors?

Tyriar:
This will be fixed in the February release as `code` is no longer a symlink to the Electron binary. See https://github.com/Microsoft/vscode/issues/2859

JanKok:
@Tyriar I have the February ("stable") release and it's still giving that message.

Tyriar:
@JanKok it was actually pushed back to March where deb and rpm packages will be available for installation.

JanKok:
I just downloaded and installed VSCode v0.10.10 dated 2016-03-07 from the ZIP file, and it's still giving that message. Other things are working better.

PR #191, opened 2016-02-28
Add & to command lines when starting Code, and reorganize

JanKok:
I suggest some additional changes in this section:

When creating a new project or page, as in this example, would it be better to include a ".", i.e. "code . index.html style.css readme.md &" ?

In order to shorten the page and remove content less relevant to newcomers,

Delete this tip (it seems obvious from the example and not very important): >**Tip:** You can have as many file names as you want separated by spaces.

Move the ## Additional Command line arguments section to another page and provide a link here, e.g. "For more information about command line arguments, see ..."

### DIFF
--- a/docs/editor/codebasics.md
+++ b/docs/editor/codebasics.md
@@ -192,15 +192,23 @@ Then choose an encoding.
 You can launch VS Code from the command line to quickly open a file, folder, or project. Typically, you open VS Code within the context of a folder. We find the best way to do this is to simply type:
 
 ```
-code .
+code . &
 ```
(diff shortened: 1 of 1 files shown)


Tyriar:
`code` has been improved as per https://github.com/Microsoft/vscode-docs/issues/189

I'll close this but feel free to open a new PR for this (I don't normally review these):

> If you don't supply _any_ folder argument, Code will resume its last editing session, even if that session was in a different folder from the folder in which the `code` command is invoked.

Thanks!

PR #199, opened 2016-02-29
Absolute link

csholmq:
Blog link now works from Github as well.

### DIFF
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -401,4 +401,4 @@ Last but certainly not least, a big *__Thank You!__* to the following folks that
 
 ## VS Code Blog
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks, I'll add a .md to the second link so the Chrome Debugger extension blog post link works as well from GitHub.

gregvanl:
Actually there is another issue here.  Our repo uses /blog but the website is /blogs and this causes GitHub navigation pain.  I'll fix it and my revert your change.  We try to avoid absolute links so we can stay in the same branch on GitHub.

csholmq:
Not really following your logic but fair enough. It relative works then go for it.

gregvanl:
Just pushed a change.  Now /blogs goes to the GitHub /blogs root and will go to the website blogs location when published on our website.

csholmq:
:+1:

gregvanl:
Thanks.  Really appreciate you reading the 0.10.10 latest.md today.

PR #201, opened 2016-03-01
Updated the profile tip to use source command.

fallenprogrammr:
Updated the tip section to use the source command to load the updated profile settings and not resort to restarting the terminal. Also, added the profile file entry for z shell users.


gregvanl:
Hi Amit,
Thank you for reading our docs and submitting a pull request.  In the interest of keeping out doc set tight, we just show instructions in the bash shell and assume that users of other shells will be able to understand how to do similar operations in their preferred shell.

PR #211, opened 2016-03-07
Fix a typo in the window events section

shardulm94:
Change the word `das` to `has` in the description of
`onDidChangeTextEditorViewColumn` event.


gregvanl:
Thank you for the pull request.  This API reference is actually auto-generated from the release's vscode.d.ts typings file but I've taken your pull request since it will be some weeks before we move the API reference forward and generate a new topic.  I've fixed the comment in the vscode.d.ts file from which the API topic is generated.

shardulm94:
Thanks for the insight! Will keep this in mind for any future issues.

PR #223, opened 2016-03-10
Added note on seeing key for other OS

estei:
Added a bit on how to see the key bindings for another OS than the one you are using.


gregvanl:
Great addition.

Issue #236, opened 2016-03-17
Increase advised fs.inotify.max_user_watches value


Tyriar:
Context: https://github.com/Microsoft/vscode/issues/4252

Also note it's configurable range?

http://code.visualstudio.com/Docs/supporting/faq#_linux-faq

PR #237, opened 2016-03-17
Improve advice around fs.inotify.max_user_watches

### LINKED ISSUE #235: Advise Linux users how to load in the ENOSPC config
Context: https://github.com/Microsoft/vscode/issues/4252

http://code.visualstudio.com/Docs/supporting/faq#_linux-faq

```bash
sudo sysctl -p
```

Tyriar:
Fixes #235, #236

### DIFF
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -152,6 +152,8 @@ The limit can be increased to its maximum by editing `/etc/sysctl.conf` and add 
 fs.inotify.max_user_watches=524288
 ```
 
+The new value can then be loaded in by running `sudo sysctl -p`.
+
 While 524288 is the maximum number of files that can be watched, if you're in an environment that is particularly memory constrained you may wish to lower the number. Each file watch [takes up 540 bytes (32-bit) or ~1kB (64-bit)](http://stackoverflow.com/a/7091897/1156119), so assuming that all 524288 watches are consumed that results in an upperbound of around 256MB (32-bit) or 512MB (64-bit).
 
 ### I can't see Chinese characters in Ubuntu
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -143,7 +143,7 @@ sudo apt-get install gvfs-bin
 When you see this error, it indicates that the VS Code file watcher is running out of handles. To increase the limit open `/etc/sysctl.conf` and add this line to the end of the file:
 
 ```
-`fs.inotify.max_user_watches=16384`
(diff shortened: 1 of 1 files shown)


Tyriar:
@gregvanl let me know if wording looks good.

docs/supporting/faq.md
```diff
@@ -146,13 +146,13 @@ When you see this error, it indicates that the VS Code file watcher is running o
...
 
-The limit can be increased to its maximum by editing `/etc/sysctl.conf` and add this line to the end of the file:
+The limit can be increased to its maximum by editing `/etc/sysctl.conf` and adding this line to the end of the file:
 
 ```
 fs.inotify.max_user_watches=524288
 ```
 
-The new value can then be loaded in by running `sudo sysctl -p`.
+The new value can then be loaded in by running `sudo sysctl -p`. Note that ArchLinux works a little different, [view this page for advice](https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers).
```
gregvanl: "Note that [ArchLinux](https://www.archlinux.org) works a little differently, "

Tyriar: :apple: works a little different

docs/supporting/faq.md
```diff
@@ -140,11 +140,19 @@ sudo apt-get install gvfs-bin
...
 ```
-fs.inotify.max_user_watches=16384
+
+The limit can be increased to its maximum by editing `/etc/sysctl.conf` and add this line to the end of the file:
+
 ```
+fs.inotify.max_user_watches=524288
+```
+
+While 524288 is the maximum number of files that can be watched, if you're in an environment that is particularly memory constrained you may wish to lower the number. Each file watch [takes up 540 bytes (32-bit) or ~1kB (64-bit)](http://stackoverflow.com/a/7091897/1156119), so assuming that all 524288 watches are consumed that results in an upperbound of around 256MB (32-bit) or 512MB (64-bit).
```
gregvanl: "constrained, you may"

gregvanl:
Thanks Dan for writing this up.  I just add a few suggestions.

Tyriar:
Cheers :)

gregvanl:
Do you think we should link to https://www.archlinux.org?

Tyriar:
May as well, I notice it's also stylized as "Arch Linux". I'll fix that up.

gregvanl:
Cool, nice to link people directly instead of them having to type in search if they aren't familiar with a technology.

PR #239, opened 2016-03-17
Correcting invalid type value in configuration example.

thegaryroberts:
The example configuration shown contains an invalid type specifier that does not allow for a default value of null (the default value used in the example). Corrected this example to match the actual value used in the real typescript.tsdk setting definition.


thegaryroberts:
For reference: The related editor screenshot shows a green squiggly under the null value and you can see 1 warning in the status bar, probably this very error! I've came across this issue because package authors have been repeating the same mistake with defaults, causing their packages to generate warnings when installed.

gregvanl:
Thanks Gary.  I'll update the screenshot as well.

PR #262, opened 2016-04-09
A better introduction of Markdown support.

warnov:
This is the first time I use VS Code. And I found myself kind of lost at the beginning trying to figure out how is the support of Markdown in the editor, because the current instructions assume certain familiarity with the tool and also omit the fact that the file must be saved before visualizing it for the first time. Just giving a little more context at the beginning would be great.


gregvanl:
Thanks.  The markdown topic is going to get some updates shortly (for example the section on using gulp needs work) and your changes are a good start.

PR #265, opened 2016-04-12
Update css.md, Output directory must be set first

chris72x:
Error while trying the directions resulted in "Output directory must be set".  There were no directions on how to set the output directory so I figured it out and included it with these directions.

### DIFF
--- a/docs/languages/css.md
+++ b/docs/languages/css.md
@@ -113,7 +113,8 @@ The first example shows how to use configure tasks for TypeScript compilation.  
     "version": "0.1.0",
     "command": "node-sass",
     "isShellCommand": true,
-    "args": ["styles.scss", "styles.css"]
+    "args": ["styles.scss", "styles.css"],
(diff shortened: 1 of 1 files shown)


chris72x:
While attempting these directions, I received the error "No output directory set".  This might have been mentioned in some other directions but this was the first set of directions I followed so it didn't work.  I could not find directions by searching the internet or Microsoft's site, so I figured it out.

gregvanl:
Hi Chris,
When I try the walkthrough (on Windows), I haven't needed to set outDir to have either tool (node-sass or less) transpile to .css.  
Can you provide some details about your configuration? For example the OS platform and version, VS Code version and anything else you think might be unique or relevant.
Thanks
- Greg

chris72x:
I have tried the process on another computer and the instructions worked as written.  The computer that gave me an error regarding output directory tried to install a Windows update today, failed, and is now stuck in a restoring to previous version of Windows.

PR #311, opened 2016-05-19
Add focus prevention level tip on Ubuntu

Tyriar:


docs/supporting/faq.md
```diff
@@ -180,6 +180,21 @@ sudo yum update
 sudo dnf update
 ```
 
+### The code bin command does not bring the window to the foreground on Ubuntu
+
+When running `code .` on Ubuntu when vscode is already open in `.` will not bring VS Code into the foreground. This is a feature of the OS which can be disabled using `ccsm`.
```
gregvanl: VS Code

gregvanl: "Running `code .` on Ubuntu when VS Code is already open in the current directory will not bring VS Code into the foreground."

Tyriar: @gregvanl Would "The `code` bin command..." work?

PR #324, opened 2016-05-27
Fix link and link text to tsconfig.json

jvilk:
The previous link pointed to an old wiki page; I have updated the link to the new handbook page.

The previous link also mistakenly referred to `tsconfig.json` as `typescript.json`.


gregvanl:
Thanks!

PR #325, opened 2016-05-29
Update not working typings reference to react to a new working one

UnderNotic:


gregvanl:
Hi Peter, 

What part of this instruction isn't working for you?  The typings command using 'react-global' appears to install the correct typings files (under typings\globals\react-global).  Is the issue that the 'react-global' typings don't give you the correct IntelliSense?

UnderNotic:
Hey Gregvanl,

The problem is that "typings install dt~react-global --global" doesn't work for me, they have moved react typings to diffrent address on DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/react)

PR #326, opened 2016-05-31
Correct an invalid command name

scottaddie:
The JavaScript language docs reference a `Reload Java Script` command, which doesn't exist in Command Palette. This PR changes the command reference to `Reload JavaScript Project`, which does exist in Command Palette.


gregvanl:
Thanks!

PR #331, opened 2016-06-03
Change word completion to code completion

waderyan:


gregvanl:
The sentence is correct.  VS Code always offers word completions meaning word-based completions. For the cases where there is no language service for a programming language, VS Code does the poor man's IntelliSense (similar to Sublime and Atom) and gives completions from the available words in the file.  In many cases, such as variable names, this is helpful.

gregvanl:
Do you think this should be made clearer?  I'm not sure we want to dwell too much on describing word based completions in this paragraph since the point is to call out code intelligence completions.  Word-based completions is fairly obvious from using the product.

waderyan:
I see your point. I'm not sure what to do here... My first impression was "what is word completion?", thinking that "code completion" is more understandable for our audience. I thought it was a mistake so I made the PR.

PR #352, opened 2016-06-10
Update definition to declaration

ligaz:
The semantically correct term for TypeScript `.d.ts` files is declaration not definition.


gregvanl:
Hi Stefan, Thank you for the pull request and you're right but "type definition" is the commonly used name in the [repository](http://definitelytyped.org) and typings [tool](https://www.npmjs.com/package/typings). We should keep the same phrasing to avoid confusion and help with search results.

ligaz:
Hey @gregvanl, It looks [this is not the case](https://github.com/Microsoft/TypeScript-Handbook/commit/c5e45771e6c4c8f4ac43991bb704361175a789d0) in the TypeScript's handbook repository.

PR #363, opened 2016-06-22
Update codebasics.md

kencenerelli:
Changed menu bar item 'Files' to File, the correct menu bar item name.


kencenerelli:
Found a typo in the docs. Menu bar item should be File.

gregvanl:
Thanks!

PR #365, opened 2016-06-23
Explain that Show Explorer key binding also toggles focus

jmm:
There's no way to preview how the docs are going to render on the website, right?

There's no default key binding for hiding the explorer?


gregvanl:
Thanks.

PR #368, opened 2016-06-27
Remove highlighting of standard text

DAXaholic:
The word "activationEvent" is wrongly highlighted.
As it is just standard text and to align with the other usages of the same word within this document, I think the highlighting should be removed.


gregvanl:
Thanks! Much better.

PR #471, opened 2016-07-17
add Link to the javascript snippets file

philj0st:
I would have liked to be directed to the place where I can get some inspiration through seeing how the existing snippets are defined.

### DIFF
--- a/docs/customization/userdefinedsnippets.md
+++ b/docs/customization/userdefinedsnippets.md
@@ -34,7 +34,7 @@ You can define your own snippets for specific languages.  To open up a snippet f
 
 Snippets are defined in a JSON format and stored in a per user `(languageId).json` file. For example, Markdown snippets go in a `markdown.json` file.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi Philipp,
Thank you for your pull request and taking an interest in our documentation. Linking to more snippet examples is a great idea but I worry that linking directly to a vscode source code file may be fragile.  If the source code is restructured that link will be broken.  I'll enter an issue to provide more snippet examples or figure out a more general way to direct users to the vscode shipped snippets.
Thanks

philj0st:
Hey Greg,
I too thought that it was a little prone to break in the future but couldn't think of another way to refer to it except the path to the installed snippets which might be even more likely to change and is different from user to user. For example `/usr/share/code/resources/app/extensions/javascript/snippets/javascript.json`.

Yeah, somehow letting the reader know where to find more examples would have been appreciated at least by me.

Thank you :ok_hand:

PR #476, opened 2016-07-20
Added new docs images (revised)

bgashler1:
FYI @gregvanl 

These are the revised images based on feedback we discussed yesterday.


waderyan:
Looks great.

gregvanl:
@bgashler1 : Couple of comments on the tagging:
codebasics/hero.png
1. Shouldn't "Activity Bar" be "View Bar"?  It lets you change the view (or viewlet). It's called out as the "View Bar" in the Basic Layout paragraph so these need to match.
2. "Editor Groups" - I think "Editor" or "Editor Region" might be clearer.  It's called out as "Editor" in the Basic Layout section below.
3. "Viewlet" is called the "Side Bar" in the View menu commands (Toggle Side Bar, Move Side Bar).  We should be consistent.  It may be just me but I don't care for "viewlet".
debugging/debugging_hero.png
1. "Output Window" should be "Debug Console Panel".  The Output Panel is a different control.

bgashler1:
@gregvanl I just addressed your feedback.  Thanks for sharing.  The only thing I didn't change was Editor Groups.  I think we should leave it that way, because of terminology Ben, Steven and I came up with since we created the new tab experiences.

PS: I'm not sure why, but several components in VS Code have different names internally than publicly in our docs (e.g. viewlet is a side bar, activity bar is a view bar).  Is there a reason for this?  I know that many of these names are also used extensively in our source code, so it would be hard to change them to reflect the documentation.  It would be nice to standardize though.

PR #540, opened 2016-08-24
Markdown external links now open in new tab

bgashler1:
Fixes https://github.com/Microsoft/vscode-website/issues/464

@waderyan 
Two things to note:

1. To use this, we needed to upgrade our **markdown-it**, so it could support a `getAttr` method used by the new dependency I added.  Note that I set the required version to `"markdown-it": "^7.0.1"` (latest release).
2. Note that the dev dependencies for **markdown-it-external-links** says that it requires version `^6.0.4` of markdown-it as a devDependency [(see here)](https://david-dm.org/rotorz/markdown-it-external-links?type=dev).  Does this mean I can't use the latest markdown-it as I mentioned earlier?  Everything seems to work fine.

We should check that nothing renders in an unexpected way in the docs on PPE before we roll this out.

### DIFF
--- a/scripts/gulpfile.common.js
+++ b/scripts/gulpfile.common.js
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
  
 var markdownIt = require('markdown-it');
+var externalLinks = require('markdown-it-external-links');
 var highlightjs = require('highlight.js');
 var swig = require('swig');
 var slash = require('gulp-slash');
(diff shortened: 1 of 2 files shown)


bgashler1:
@gregvanl FYI since this concerns the docs, you might want to verify nothing is busted or rendering weird.

scripts/gulpfile.common.js
```diff
@@ -64,6 +65,10 @@ exports.mapFileToArticle = function(file) {
 
 exports.compileMarkdown = function(file, article) {
 	var md = markdownIt({ html: true, langPrefix: '' });
+	md.use(externalLinks, {
```
waderyan: This is exactly where this code was supposed to go. Great job.

waderyan:
Looks good Brad. 

> Note that the dev dependencies for markdown-it-external-links says that it requires version ^6.0.4 of markdown-it as a devDependency (see here). Does this mean I can't use the latest markdown-it as I mentioned earlier? Everything seems to work fine.

We should be fine.

waderyan:
@gregvanl merge this in at your convenience. We're not in a rush to get it out.

PR #565, opened 2016-09-09
Js roundup blog

waderyan:


waderyan:
@gregvanl are we still good to publish this tomorrow? Can you take an edit pass on it?

PR #587, opened 2016-09-18
Change description for languages

seanmcbreen:
The list as it was implied a short lost of languages so this change is designed to correct that and make the page easier to maintain.


gregvanl:
I merged this PR but I thought we wanted to list other languages to help with SEO so that users would find our topic when searching for languages like Coffee Script, Java, Powershell, etc.

PR #590, opened 2016-09-19
Java placeholder

adenner:
Redhat java is now supported https://marketplace.visualstudio.com/items?itemName=redhat.java
added a placeholder into documentation to reflect this


seanmcbreen:
Thanks @adenner we are very excited to see the Java work shipped.  

It's been a great partnership between our teams.  For now we are attempting to work out a way to list out more of the marketplace extensions so that the extension authors can include the docs in their readmes.  That way our docs will be accurate and for users they will know where to go for autorative information.

Do for now we won;t take this pull request but we will work to maaker Java visible and sucessful (we have listed in our featured extensions and are adding it to in-product recomendations).

Sean

PR #591, opened 2016-09-20
Fix GFM Question Since We're Using CommonMark

mjbvz:
I believe that the VS Code markdown editor now targets CommonMark markdown using the Markdown-it library, instead of GFM using the marked library. This fixes the question and answer about GFM.


gregvanl:
Thanks!

PR #642, opened 2016-10-14
Document that deactivate() should return a promise in case cleanup is…

daserge:
… async

See details in https://github.com/Microsoft/vscode-react-native/pull/316.


gregvanl:
Thanks!

PR #649, opened 2016-10-26
Keymaps and Formatters category.

waderyan:
I think this is the right process for adding to release notes...


gregvanl:
Feel free to update the vnext release notes without a pull request. We do several edit passes before they are published.

PR #654, opened 2016-10-29
Update unFoldRecursively key binding

laggingreflex:


gregvanl:
Good catch. Rather than take this PR, I've entered https://github.com/Microsoft/vscode/issues/14723 to change unFoldRecursively to unfoldRecursively to match other unfold commands (unfoldAll).
Thanks!

PR #655, opened 2016-10-30
extension-manifest Spell example does not use lower case name

sleutho:
Example name does not match the pattern of "^[^A-Z]+$".


gregvanl:
Thanks!

PR #658, opened 2016-11-01
Update linux.md

sudopluto:
Use apt instead of dpkg because dpkg doesn't handle dependencies automatically. In general, Ubuntu and Debian are moving away from apt-get to apt.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -17,7 +17,7 @@ MetaDescription: Get Visual Studio Code up and running on Linux.
 
 ```bash
  # For .deb
- sudo dpkg -i <file>.deb
+ sudo apt install ./<file>.deb
 
  # For .rpm (Fedora 21 and below)
  sudo yum install <file>.rpm


Tyriar:
Doesn't `apt` only work on debs sourced from repos, not local ones? The recommendation AFAIK to ensure dependencies is installed is to do a dpkg install followed by `sudo apt-get install -f` http://unix.stackexchange.com/a/159114

sudopluto:
apt can handle local debs:

image

Tyriar:
@sudopluto does this work on 32/64 bit Ubuntu 14.04 and Debian 7 too?

sudopluto:
@Tyriar I believe so, but I can't confirm. I think that apt, was introduced in debian, and then trickled down  into ubuntu by 14.04.

Tyriar:
It might be best to let [the clock run out on Ubuntu 12.04 LTS support](https://assets.ubuntu.com/v1/65d114f8-release-chart-desktop.png?w=800) at the very least then before switching to this recommendation since both work. Suggesting a call to `sudo apt-get install -f` would be useful though to install the deps.

gregvanl:
Add mention of "sudo apt-get install -f" with https://github.com/Microsoft/vscode-docs/commit/f3565093392f88bb7f33151cac212d11ecdf9a17

PR #659, opened 2016-11-01
Add note about unbound commands in keybindings.md.

ChristianAlexander:
Not all commands are listed in keybindings.md. This has led to at least one [feature request](https://github.com/Microsoft/vscode/issues/14266) for a command that already exists.

This change points users to the `Default Keyboard Shortcuts` file for more commands.


gregvanl:
Thanks!

PR #661, opened 2016-11-02
Fix "next" links in intro videos

Gpx:
The order of the videos in the first post was not the same as the TOC. Also, the *next* links were not following the correct order.


waderyan:
Looks good. Thank you.

PR #669, opened 2016-11-05
Replace the "LICENSE" file for "LICENSE.md"

eridem:


gregvanl:
Hi @eridem, 
Thank you for your pull request. The vsce tool accepts the filename extensions .md and .txt and also no extension.  When a file simply called "LICENSE" is found, the .txt extension is added.
So the documentation is correct as a "LICENSE" filename is valid. 
Plaintext files are easier to read while a Markdown file needs to be rendered to be viewed without the markup.

eridem:
Nice! I didn't know.
Keep the good work!

gregvanl:
Thanks. We really appreciate people taking an interest in our docs.

PR #678, opened 2016-11-14
Minor typo fixed

xekri:
Occurences => Occurrences


gregvanl:
It is also incorrect in the product in the Command Palette. I'll enter an issue to fix it there as well.
Thanks!

PR #681, opened 2016-11-16
Tree Explorer doc

octref:
@gregvanl Don't know much about the front-matter. Let me know if I did anything wrong.

@jrieken Can you take a look and let me know if anything is in accurate?


jrieken:
I think it's too early to write doc about this. For once because it is *proposed api* only and more importantly because things are still very much in flux. We will be busy updating this document all the time which I think is too much effort. 

I think we should move the disclaimer about proposed api and its restrictions into the `vscode.proposed.d.ts`-file because it applies to everything in there and should be very prominent. 

IMO the tutorial part of this should go into one of the samples.

octref:
@jrieken Sure, but would you mind if I add a section about "Proposed API" here?

image

I'll just explain the nature of the proposed API, and provide links to the `vscode.proposed.d.ts` and the sample using this API. Everything specific to the Tree Explorer API will be in `vscode-extension-samples/tree-explorer-sample/README.md`.

octref:
Closing in favor of #682.

Content here is moved to https://github.com/Microsoft/vscode-extension-samples/blob/master/tree-explorer-sample/USAGE.md

PR #686, opened 2016-11-21
Addition of new the system variable name $(workspaceRootFolderName) t…

visualperception:
…o documentaion


isidorn:
We need to document this variable, so this looks good to me.
Assigning to @gregvanl to do the final revision

gregvanl:
Thanks!
I'll add the same variable to the debugging.md launch.json configuration section. I may also change "solidus" to "trailing slash" as "solidus" is not in common use.

visualperception:
note that it has no leading and no trailing slash so "no slashes"

PR #688, opened 2016-11-25
Update docs to reflect current include path implementation

s-martin:
Currently only absolute paths are supported, see https://github.com/Microsoft/vscode-cpptools/issues/366. This should be reflected in the current documentation.


gregvanl:
@sean-McManus Do you want to take this doc update now or will this be fixed shortly in the extension?

sean-mcmanus:
Relative paths work on Windows, but they don't work on Linux. I don't know if you want to document that buggy behavior or not.

gregvanl:
Closing this pull request per Sean's comment.

PR #715, opened 2016-12-12
Chromium SUID sandbox error can be ignored

Tyriar:
It's being removed in Chromium so it's probably not worth the effort to fix and validate it


gregvanl:
Thanks

PR #721, opened 2016-12-13
Fix #708 - add basic node2 description

### LINKED ISSUE #708: Add section on node2 to debugging docs
It has only been mentioned in release notes in the past. It would be good to call it out briefly under "debugging Node.js".

roblourens:
I need to add the rest of the details from the readme - https://github.com/Microsoft/vscode-node-debug2 - but will start a new issue for that


roblourens:
Actually, I think it would confuse people by adding node2-specific info to the same page as node-debug docs. I think we should only cover the default node debug experience (plus this blurb about node2) so I'll leave it with this.

PR #744, opened 2016-12-13
Fix #713 - node sourcemap troubleshooting

### LINKED ISSUE #713: Add info on Node debug troubleshooting
Look at common node-specific issues and add a few notes about them. Especially sourcemap issues and breakpoints not binding. There is very little official info out there, beyond github issues.

Example from chrome-debug: https://github.com/Microsoft/vscode-chrome-debug#troubleshooting

roblourens:
Please temporarily forget everything you know about sourcemaps, read this, and tell me if it could be helpful.

It might be too much information. On one hand, detailed info on how this process works may help some users help themselves. On the other hand, I don't want to scare users away by making this seem more complicated than it is. Most of the time, debugging with source maps just works. The rest of the time, it's usually one of the first 3 bullet points. 

The important ideas in the paragraphs before the bulleted list are
* We have to get the .js first before we can map .js <-> .ts
* The paths in the source map file have to be correct

If this should be a detailed explanation of everything that happens, it could be expanded. If it should be more of a basic troubleshooting list, it could be cut.

I'm not mentioning the `trace` option since it's not quite official and we should decide how to standardize between node and node2 before documenting it.

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -330,14 +330,15 @@ When you build your `app.ts` file in TypeScript with source maps enabled, it eit
 Finally, the debug adapter searches for the full path of `app.ts` in this resulting list of `.ts` files. If there's a match, it has found the source map file to use when mapping `app.ts` to `app.js`. If there is no match, then it can't bind the breakpoint, and it will turn gray.
 
 Here are some things to try when your breakpoints turn gray.
-* First and most important: do you have `"sourceMaps": true` in your launch config?
+* Do you have `"sourceMaps": true` in your launch config?
 * Did you build with source maps enabled? Are there `.js.map` files, or inlined source maps in your `.js` files?
 * Did you set the `outFiles` property in your launch config? It should be a glob pattern for an absolute path that matches your `.js` files.
 * Try the new experimental `node2` debug adapter. It can handle some more of the more complex source map cases.
 * Are the `sourceRoot` and `sources` properties in your source map correct? Can they be combined to get the correct path to the `.ts` file?
 * Are you using Webpack? By default, it outputs paths with a `webpack:///` prefix, which the debug adapter can't resolve. You can change this in your Webpack config with the `devtoolModuleFilenameTemplate` option, or try using `node2`, which provides some extra options for resolving these paths.
 * Have you opened the folder in VS Code with the incorrect case? It's possible to open folder `foo/` from the command line like `code FOO` in which case source maps may not be resolved correctly.
-* Try adding a `debugger` statement. If it breaks into the `.ts` file there, but breakpoints at that spot don't bind, that is useful information to include with a Github issue. If it breaks into a `.ts` file that says "read-only inline content from source map" next to it in Open Editors, then that's a sure sign that the source is inlined in the source map, and the path in the source map is incorrect.
+* Try searching for help with your particular setup on Stack Overflow or by filing an issue on Github.
+* Try adding a `debugger` statement. If it breaks into the `.ts` file there, but breakpoints at that spot don't bind, that is useful information to include with a Github issue.
 
 
 ### Attaching VS Code to Node.js
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -323,16 +323,16 @@ VS Code supports both the **inlined source maps** and the **inlined source**.
 
 A common issue when debugging with source maps is that you'll set a breakpoint, and it will turn gray. If you hover the cursor over it, you'll see the message, `"Breakpoint ignored because generated code not found (source map problem?)"`. What now? There are a range of issues that can lead to this. First, a quick explanation of how the Node debug adapter handles source maps.
 
-When you set a breakpoint in `app.ts`, the debug adapter has to figure out the path to `app.js`, the transpiled version of your TypeScript file, which is what is actually running in Node. But, there is not a straightforward way to figure this out starting from the `.ts` file. Instead, the debug adapter uses the `outFiles` attribute in the launch config to find all the transpiled `.js` files, and parses them for a source map, which contains the locations of its associated `.ts` files.
+When you set a breakpoint in `app.ts`, the debug adapter has to figure out the path to `app.js`, the transpiled version of your TypeScript file, which is what is actually running in Node. But, there is not a straightforward way to figure this out starting from the `.ts` file. Instead, the debug adapter uses the `outFiles` attribute in the `launch.json` to find all the transpiled `.js` files, and parses them for a source map, which contains the locations of its associated `.ts` files.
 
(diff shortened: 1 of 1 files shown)


docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
...
+
+When you build your `app.ts` file in TypeScript with source maps enabled, it either produces an `app.js.map` file, or a source map inlined as a base64-encoded string in a comment at the bottom of the `app.js` file. To find the `.ts` files associated with this map, the debug adapter looks at two properties in the source map, `sources`, and `sourceRoot`. `sourceRoot` is optional - if present, it is prepended to each path in `sources`, which is an array of paths. The result is an array of absolute or relative paths to `.ts` files. Relative paths are resolved relative to the source map.
+
+Finally, the debug adapter searches for the full path of `app.ts` in this resulting list of `.ts` files. If there's a match, it has found the source map file to use when mapping `app.ts` to `app.js`. If there is no match, then it can't bind the breakpoint, and it will turn gray.
+
+Here are some things to try when your breakpoints turn gray.
+* First and most important: do you have `"sourceMaps": true` in your launch config?
```
isidorn: No need to put first and most important here. I mean it is already first in the list :)

docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
 
 VS Code supports both the **inlined source maps** and the **inlined source**.
```
isidorn: I think using the term `launch.json` is clearer than launch config. Also I feel that launch.json is what we usually use in other places.

docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
...
+* Try the new experimental `node2` debug adapter. It can handle some more of the more complex source map cases.
+* Are the `sourceRoot` and `sources` properties in your source map correct? Can they be combined to get the correct path to the `.ts` file?
+* Are you using Webpack? By default, it outputs paths with a `webpack:///` prefix, which the debug adapter can't resolve. You can change this in your Webpack config with the `devtoolModuleFilenameTemplate` option, or try using `node2`, which provides some extra options for resolving these paths.
+* Have you opened the folder in VS Code with the incorrect case? It's possible to open folder `foo/` from the command line like `code FOO` in which case source maps may not be resolved correctly.
+* Try adding a `debugger` statement. If it breaks into the `.ts` file there, but breakpoints at that spot don't bind, that is useful information to include with a Github issue. If it breaks into a `.ts` file that says "read-only inline content from source map" next to it in Open Editors, then that's a sure sign that the source is inlined in the source map, and the path in the source map is incorrect.
+
```
isidorn: All of this bullet points should be copied into a fake stackOverflow question and answer. Since a lot of users will first seach there.
Example me asking and ben answering is the first result :D https://www.google.rs/?gws_rd=cr&ei=nQtRWJyWEoXPa9LGirgB#q=how%20to%20disable%20tabs%20vscode

docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
...
+Here are some things to try when your breakpoints turn gray.
+* First and most important: do you have `"sourceMaps": true` in your launch config?
+* Did you build with source maps enabled? Are there `.js.map` files, or inlined source maps in your `.js` files?
+* Did you set the `outFiles` property in your launch config? It should be a glob pattern for an absolute path that matches your `.js` files.
+* Try the new experimental `node2` debug adapter. It can handle some more of the more complex source map cases.
+* Are the `sourceRoot` and `sources` properties in your source map correct? Can they be combined to get the correct path to the `.ts` file?
+* Are you using Webpack? By default, it outputs paths with a `webpack:///` prefix, which the debug adapter can't resolve. You can change this in your Webpack config with the `devtoolModuleFilenameTemplate` option, or try using `node2`, which provides some extra options for resolving these paths.
```
isidorn: It is not 100% clear how to use the new `node2` adapter. Maybe mention to simple put it as a type in your launch.json config. Or add a link to the section where node2 is explained.

docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
...
+* Have you opened the folder in VS Code with the incorrect case? It's possible to open folder `foo/` from the command line like `code FOO` in which case source maps may not be resolved correctly.
+* Try adding a `debugger` statement. If it breaks into the `.ts` file there, but breakpoints at that spot don't bind, that is useful information to include with a Github issue. If it breaks into a `.ts` file that says "read-only inline content from source map" next to it in Open Editors, then that's a sure sign that the source is inlined in the source map, and the path in the source map is incorrect.
```
isidorn: The last point feels like too much detail that only a person who is filling a bug should know. I would remove it - unless we get a special page only for source maps.
Or I would rephrase it to the following: "Still not working, file an issue lalala and then mention this useful detail"

docs/editor/debugging.md
```diff
@@ -319,6 +319,27 @@ Source maps can be generated with two kinds of inlining:
 
 VS Code supports both the **inlined source maps** and the **inlined source**.
 
+#### JavaScript Source Map Tips
+
```
isidorn: I would like to see an example of how source maps need to be setup.
Maybe create a public git hub repositore as a sample and then just link it from here.

roblourens: We do have the sample projects, I can link them

docs/editor/debugging.md
roblourens: It does feel weird, it's also the only actually concrete piece of info. I'll make that change tho

docs/editor/debugging.md
roblourens: There are already a bunch of very not helpful SO posts. Not sure if I can cut through the noise :)

docs/editor/debugging.md
roblourens: Do you know what these sample projects are actually used for? I don't see where they are linked. And they might need a little work - https://github.com/Microsoft/vscode-samples/tree/master/node-express-typescript

isidorn: No clue. @chrisdias might know

chrisdias: these were very early samples of basic express apps that were fully configured to get the best out of VS Code. I don't think we reference them anywhere.

roblourens: I think that a really simple example of sourcemap debugging isn't that helpful because it's usually easy to set up. Problems come in with more complicated setups, and those are all different.

docs/editor/debugging.md
roblourens: Oh yeah I was going to do that but that section is only in vnext. I filed #753 to remind myself.

gregvanl:
Oops, I got merge-happy and didn't realize this was assigned to Isi and Andre.

PR #749, opened 2016-12-14
Update user and workspace settings and an outdated extension recommendations icon

sandy081:
Update user and workspace settings and an outdated extension recommendations icon


sandy081:
@gregvanl FYI

Issue #702, opened 2016-12-08
December Doc-a-thon


gregvanl:
The goal is to identify missing, stale or obsolete content in the main VS Code docs.

You can:

1. Enter a vscode-docs issue describing the necessary change or addition.
2. Enter an issue and point to relevant existing content to port (release notes, issue comment, protocol or sample repo content).
3. Create or update content and submit a PR against vscode-docs master.

Topics

-----
@gregvanl New topics that need a home:
- **deferred** Keyboard Layouts @alexandrudima
- [x] Extensions that are only available to insiders @joaomoreno
- [x] Extension Authoring>[Extension Packs](https://code.visualstudio.com/docs/extensionAPI/extension-manifest#_extension-packs) @sandy081
- [x] Recommended Extensions @sandy081
- [x] Customization>File icons @aeschli
- [x] Hot Exit @tyriar
- [x] Promote Go to the same level as other languages in the docs, PHP, Python etc. @ramya-rao-a
- [ ] (initial progress) Editor Architecture @rebornix
-----

Setup

- [x] Overview @chrmarti
- [x] Windows @ramya-rao-a
- [x] Mac @chrmarti
- [x] Linux @Tyriar
- [x] Additional Components @chrmarti

Editor

- [x] The Basics 
  - [x] @rebornix 
  - [x] @chrmarti
- [x] Extension Marketplace @joaomoreno
- [ ] IntelliSense 
  - [x] @mjbvz 
  - [ ] @waderyan [vscode-docs #742](https://github.com/Microsoft/vscode-docs/issues/742)
- [ ] Editing Evolved 
  - [ ] @alexandrudima 
  - [x] @ramya-rao-a
- [ ] Debugging 
  - [ ] @weinand 
  - [ ] @isidorn 
  - [x] @roblourens
- [x] Version Control @joaomoreno
- [x] Tasks @dbaeumer
- [x] Integrated Terminal@Tyriar
- [x] Tabs @bpasero
- [ ] Accessibility 
  - [ ] @alexandrudima 
  - [x] @rebornix
- [x] Why VS Code @mousetraps

Customization

- [x] Overview @sandy081
- [x] User and Workspace Settings @sandy081
- [ ] Key Bindings @alexandrudima
- [x] Snippets @jrieken
- [x] Colorizer @aeschli
- [x] Themes @aeschli
- [x] Display Language 
  - [x] @dbaeumer 
  - [x] @ramya-rao-a

Languages

[...]

PR #754, opened 2016-12-14
Hot exit documentation

Tyriar:
#702

This section be expanded upon when Microsoft/vscode#15467 is resolved


Tyriar:
@gregvanl not sure if you had a way of dealing with menu options that differ across OS's, I believe the text is Quit on OSX.

gregvanl:
File > Exit on Windows and Linux. Code > Quit on macOS. I'll update the text.

PR #762, opened 2016-12-15
Runtime>Node: updated to 1.8 (removed the references to typings)

egamma:
- removed the setup steps for the typings file
- removed the description of the jsconfig.json file, since it is not needed for this walk through.


egamma:
// CC Chris pls review

PR #768, opened 2016-12-15
doc-a-thon

chrisdias:


chrisdias:
this PR cleans up the current doc. i will update the doc further tonight to add in the additional functionality provided by the extension.

PR #774, opened 2016-12-17
refreshed "Why VS Code?" page

mousetraps:
#702 

- communicate value prop better
- added "pizazz" (e.g. "an editor that gets out of your way", "delightfully frictionless", "lightning fast", "make it your own", "built with love"... I was on a roll 😉)
- rewrote text to be more concise
- added headers and whitespace to make it more easily scannable
- updated content to reflect more recent state of affairs
- highlight customizability and extensibility more, as well as other key features that tend to be popular

There are enough changes it's probably easiest to read it directly from the branch:
https://github.com/mousetraps/vscode-docs/blob/125b0dedf71793f327058f8894d980be18d7ff34/docs/editor/whyvscode.md

### DIFF
--- a/docs/editor/whyvscode.md
+++ b/docs/editor/whyvscode.md
@@ -23,7 +23,7 @@ At its heart, Visual Studio Code features a lightning fast source code editor, p
 
 For serious coding, you'll often need to work with code as more than just text. Visual Studio Code includes built-in support for always-on IntelliSense code completion, rich semantic code understanding and navigation, and code refactoring.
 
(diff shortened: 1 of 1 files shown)


mousetraps:
cc @waderyan @seanmcbreen @chrisdias - worth a read-through. Also it seems like we are using "code-edit-debug", "code-compile-debug", and "code-build-debug" loop in different places... is there one we should be sticking to?

gregvanl:
I propose using "edit-build-debug". I prefer "edit" to "code" which is ambiguous and "build" rather than "compile" since there are lots of different pre-production stages (preprocess, transpile, deploy, etc).

PR #776, opened 2016-12-19
Fixed inconsistency in markdown

rogersachan:
The markdown example should not use two different kinds of headers.


gregvanl:
Thanks!

PR #781, opened 2016-12-20
Chrisdias/nodeblog

chrisdias:


chrisdias:
wade gave me thumbs up on merge

PR #792, opened 2017-01-03
Update codebasics.md documentation for opening code with URLs

KattMingMing:
@joaomoreno - Not sure if this is totally needed, but it seemed like it could be helpful.

##### Description
Adds documentation to open a project and to open a file at a specific line and column number


joaomoreno:
Thanks!

PR #801, opened 2017-01-13
Menu option just View -> Integrated Terminal

ebrandel:
The menu no longer says "Toggle Integrated Terminal". It's just "Integrated Terminal". Looking at v1.8.1


gregvanl:
Thanks!

PR #804, opened 2017-01-20
Updated example code

jawn:
The existing example code included the commands.registerCommand('extension.sayHello') which is code for a different example.

Fixed by replacing the sample code with https://github.com/Microsoft/vscode-wordcount/blob/master/extension.ts


gregvanl:
Hi @jawn, thank you for taking an interest in our docs and submitting a PR. The walkthrough is designed to be done step-by-step updating the default extension template as we add features to the Word Count extension. We don't want the reader to paste the final source code from the sample repo.

PR #806, opened 2017-01-23
Fixes #651

### LINKED ISSUE #651: Bad practice suggestion in JavaScript doc
TypeScript team recommends against the following suggestion in [JS doc](http://code.visualstudio.com/docs/languages/javascript). 

image

Recommendation is to use "include". 

image

waderyan:


docs/languages/javascript.md
```diff
@@ -95,24 +95,22 @@ The `exclude` attribute tells the language service which files are and are not p
...
-    "files": [
-        "src/"
+    "include": [
+        "src/**/*"
     ]
 }
 ```
 
-> **Tip:** Best practice, and least error prone route, is to use the `files` attribute with a single `src` folder. 
+The best practice, and least error prone route, is to use the `include` attribute with a single `src` folder. The `include` attribute cannot be used with the `exclude` attribute. If both are specified, the `files` attribute will take precedence.
```
gregvanl: Shouldn't this be either `include` or `exclude` not `files` in "If both are specified, the `files` attribute will take precedence."

waderyan: Yep. Thanks for catching.

PR #813, opened 2017-01-26
remove unused and large files

waderyan:


waderyan:
I wrote a script that removed all of these images. I did a spot check on a couple dozen. If you want to spot check that would be helpful.

PR #822, opened 2017-02-02
Fix an encoding issue with the blockComment example

brettcannon:


gregvanl:
@brettcannon Thank you for your PR. This topic is autogenerated from https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts so the fix will need to be made there.
I'll investigate if this is possible.

PR #826, opened 2017-02-03
dbaeumer/tasks january update

dbaeumer:


gregvanl:
Thanks. This was on my radar but thank you for doing it.

PR #831, opened 2017-02-06
Multiple tasks are supported since 1.9, removing previous workaround

borekb:


gregvanl:
Thanks for helping with the clean up!

PR #833, opened 2017-02-06
Added `textCompareEditorVisible` when-clause context

borekb:
Related: https://github.com/Microsoft/vscode/issues/9988


gregvanl:
Thanks. This context isn't used in the default keybinding.json but is likely useful for some binding scenarios.

PR #836, opened 2017-02-06
debug: update docs from 1.9 release notes

isidorn:
* Zen mode mention new settigns
* Debug: surface no launch.json debugging, automatic environment picking, add configuration from dropdown

I tried to find other places where we mention that debugging requires a folder but could not find that (since this is no longer needed).

fyi @weinand


gregvanl:
Thanks!

PR #837, opened 2017-02-07
change command from 'code' to 'code-insiders'

shinexia:


gregvanl:
Hi @shinexia. Thank you for your interest in our documentation and this PR. However "code ." is the correct command line invocation for the stable release of VS Code. "code-insiders" is the launch command for the insiders (nightly) build.

PR #840, opened 2017-02-08
Add Run command to keyboard shortcuts

adamburgess:
Run without debugging was added some time ago: https://github.com/Microsoft/vscode/issues/2780
but it's not on the keybindings page and also not in the online pdf/docx
I would edit the docx but since it's compressed/binary and thus won't turn up in the diff well, I'll leave that to a member.


gregvanl:
Thanks! There was a typo in your PR (both should be workbench.action.debug.run) which I'll fix.

adamburgess:
Ah, oops
thanks

PR #858, opened 2017-02-23
Vnext

horriblechild:


gregvanl:
@horriblechild Not sure why you submitted this PR. The vnext branch is for prerelease content and should not go to master yet.

PR #859, opened 2017-02-25
Update userdefinedsnippets.md

DuncanSungWKim:


gregvanl:
Thank you for noticing the issue. The issue is due to missing single quotes around kb(). The fix should be `kb(insertSnippet)`. I'll go ahead and make that fix.

PR #866, opened 2017-03-01
Add documentation for openSUSE/SLE

ismail:
Mostly the same as yum one, only the file path is different and we have to add repository type.


Tyriar:
Thanks @ismail this is exactly what I was after!

PR #869, opened 2017-03-02
Add parameterHints prop

ruskakimov:
Parameter hints is an intellisense feature, so it should be present. (+ It was hard to find the way to disable property info popup without knowledge of the exact keywords)


gregvanl:
Thanks!

PR #877, opened 2017-03-08
Describe the sourceFileMap section of launch.json

KUGA2:
I struggled to find this information. I found it by chance in the autocompletion of the json file. I dont think it is documented anywhere.

The example could be shorter, but i guess its better to have a real example.


pieandcakes:
@KUGA2 I approve. Just FYI, we also have documentation in our repo at https://github.com/Microsoft/vscode-cpptools

gregvanl:
Thanks!

PR #890, opened 2017-03-19
Update typescript.md

mduberstein:
Was jsconfig.json, proposed tsconfig.json.  Looks like there was a typo in section  ## Mixed TypeScript and JavaScript projects


gregvanl:
Hi @mduberstein,
Thank you for the PR and your interest in our documentation. The current documentation is correct. tsc will detect tsconfig.json files automatically but you need to give it a hint to use a jsconfig.json file for mixed TS/JS projects.

PR #898, opened 2017-03-21
Fix reference to wrong window title variable

DAXaholic:
Used window title variable in defaults is ${activeEditorShort} - not ${activeEditorName}.


gregvanl:
Thanks! Must have missed a late breaking change.

PR #916, opened 2017-03-31
Add dnf to Linux install docs

elektronik2k5:


Tyriar:
@elektronik2k5 the reason I didn't add `dnf` here is because `yum` seems to redirect to `dnf` (at least on Fedora), so until it becomes the standard this feels like it adds additional complexity to an already complicated page for little gain.

elektronik2k5:
@Tyriar, thank you for clarifying that - I like your intent and agree that documentation should be as concise and approachable as possible! :)
However, I'd also argue that for most GNU/Linux users, correctness is just as important - especially since we're dealing with users who are developers.
In fact, that's the exact reason I opened this PR - I assumed (incorrectly) that the docs aren't up to date :)

Tyriar:
Makes sense, I guess it doesn't add too much to the page. Thanks @elektronik2k5!  :smiley:

PR #929, opened 2017-04-06
Update javascript.md

htulipe:
Remind users to disable default validator

Eslint won't work unless you disable default JS default validator. You can see that by noticing that errors are not reported by `eslint` but by `js` in the errors panel.


gregvanl:
Hi @htulipe, I believe ESLint works fine with the default JavaScript validation enabled. Your time is validate as a suggestion to users if they want to limit their issues to ESLint

htulipe:
Hey @gregvanl, thanks for taking time to review this PR. I confirm that, for me at least, eslint was not working unless I updated my settings. Anyway, great job on this new editor!

PR #930, opened 2017-04-06
Sublime Text roundup

waderyan:
Would love for this to go out Monday. 

cc @seanmcbreen @egamma


egamma:
Looks good, some suggestions
- in the keymap section should mention that VS Code supports debugging and that there is a conflict for F5, CTRL-F5 etc.
- would mention that there is a minimap and that there is a setting to enable minimap
- would mention that there is a format on paste setting
- would mentioin the setting that snippets can be sorted to the top to make it more sublime like
- How about creating a Sublime Extension Pack with these extensions

waderyan:
@gregvanl this is good now. Tested locally.

PR #938, opened 2017-04-10
Small typo

Dzejms:
Minor change, but it had me concerned that I was doing something wrong when my experience didn't match the documentation.


gregvanl:
Thanks!

PR #940, opened 2017-04-11
Document workaround for keyboard dispatching

alexandrudima:
@gregvanl I have added a note documenting a workaround in case keyboard dispatching does not work with scan codes.


alexandrudima:
thank you!

PR #944, opened 2017-04-16
Update nodejs-tutorial.md

samkimhis:
The default variable generated by express-generator (version 4.15.0 in my case) is `var index = require('./routes/index')`, not `routes`. 

I think it is better to update relevant images as well: `images/nodejs_debugsession.png` and `images/nodejs_moduleintellisense.png`.


gregvanl:
Thanks!
We'll update the screenshots shortly.

PR #949, opened 2017-04-19
Update linux.md

Rick-Anderson:
You're not telling folks how to run it, you're telling them how to install it. That's the most important SEO word.


gregvanl:
Thank you for the suggestion. The platform specific topics also contain configuration and common questions when using VS Code on those platforms and so we like the "Running VS Code on..." title. So far our site traffic to those pages has been good. Not taking this PR now but will keep the suggestion in mind.

Rick-Anderson:
It has little to do with running VS Code on linux - that's what I was looking for. Consider at Installing and Running VS Code on Linux.

PR #950, opened 2017-04-20
Explain we have opened two VS Code instances (lang svr example)

vossad01:
I was not a very dutiful reader and somewhat skimmed the initial commands which lead me to pointing one instance of VS Code at the entire cloned repository.

Later when I read,
> Go to the VS Code instance containing the server code (see above) and press [...]

I was lost because I did not remember reading anything about opening dedicated instances.  A quick text search for _instance_ did not find anything earlier in the document, so I had to go back and re-read.

This adds a brief explanatory sentence that might be caught by a reader or at least provides something to search for when you encounter the text "see above".

### DIFF
--- a/docs/extensions/example-language-server.md
+++ b/docs/extensions/example-language-server.md
@@ -30,6 +30,7 @@ Clone the repository and then do:
 > npm install
 > code .
 ```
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks, nice clarification.

PR #951, opened 2017-04-20
Update the Quick Open keybinding for Win

justinwhite15:
Update from ctrl+p to ctrl+shift+p

### DIFF
--- a/scripts/keybindings/doc.keybindings.win.json
+++ b/scripts/keybindings/doc.keybindings.win.json
@@ -515,7 +515,7 @@
 { "key": "ctrl+shift+u",          "command": "workbench.action.output.toggleOutput" },
 { "key": "ctrl+pageup",           "command": "workbench.action.previousEditor" },
 { "key": "ctrl+e",                "command": "workbench.action.quickOpen" },
-{ "key": "ctrl+p",                "command": "workbench.action.quickOpen" },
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @justinwhite15, Ctrl+Shift+P opens the Command Palette on Windows.  Ctrl+P brings up Quick Open.
Also you are modifying one of our website keyboard shortcut mapping files under the scripts directory. What are you trying to do?

justinwhite15:
Ahhh...got it.  I think what I should be instead requesting/suggesting is that on the page:
https://code.visualstudio.com/docs/editor/versioncontrol#_branches-and-tags

That is should say you should use the Command Palette to run Git Branch which is Ctrl+Shift+P.  Does that make sense?

gregvanl:
That section is in flux as the new Git extension doesn't yet support the Git branch and checkout commands in Quick Open. If you switch back to the old version (bring up Command Palette, then run SCM: Disable Preview SCM), you'll see "git" in the Quick Open list.
Let me check with the developer working on this and see if git in Quick Open is coming back next release or if we need to update the documentation.
Thanks for catching this.

gregvanl:
Doesn't look like it is coming soon. https://github.com/Microsoft/vscode/issues/24408.

PR #952, opened 2017-04-21
Corrected description of Program Files (x86)

jesusperaltac:
Program Files (x86) is for a 32-bit machine


gregvanl:
Hi @jesusperaltac. Thank you for your interest in the VS Code documentation. The path is correct for 64 bit Windows OS machines. VS Code is a 32 bit application (you can see this in Task Manager) so it will be installed under C:\Program Files (x86)\Microsoft VS Code on a 64 bit machine. 
It will go under C:\Program Files\Microsoft VS Code on a 32 bit OS machine.
Most people are running 64 bit Windows for the larger address space so that is why we just call out that installation path.

jesusperaltac:
Hi @gregvanl, I had a misunderstanding. Thanks!

PR #957, opened 2017-04-25
Update "Unity Development with VS Code" page...

eppz:
...(removed redundant entries now supported by Unity 5.5 by default).

### DIFF
--- a/docs/other/unity.md
+++ b/docs/other/unity.md
@@ -74,24 +74,6 @@ Choose the `-CSharp` version of the solution file and VS Code will light up.
 
 ![Choose Solution](images/unity/selectsln.png)
 
-**Q: How can I manually configure Editor Settings?**
-
-**A:**  To set VS Code as the External Editor for Unity go to the **Edit** > **Preferences...** menu.  From there move to the **External Tools** tab where you should see a screen like the following:
-
-![External Editor settings](images/unity/editorsettings.png)
-
- Click the browse button and set the VS Code executable (under `Program Files (x86)\Microsoft VS Code\Code.exe`) as the External Script Editor.
-
- This will enable Unity to launch VS Code whenever you open a script.
-
-**Q: I want to set the command line arguments when I launch VS Code.**
-
-**A:**  The workflow between Unity and VS Code is much smoother if you reopen an existing editor window and pass in context so that the editor lands on the exact line number and column location (for example to navigate to a compile bug).
-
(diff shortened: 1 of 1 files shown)


docs/other/unity.md
```diff
@@ -74,61 +74,6 @@ Choose the `-CSharp` version of the solution file and VS Code will light up.
...
-
-**Q: I want to set the command line arguments when I launch VS Code.**
-
-**A:**  The workflow between Unity and VS Code is much smoother if you reopen an existing editor window and pass in context so that the editor lands on the exact line number and column location (for example to navigate to a compile bug).
-
->**Note:** This option is only available in the Windows version of Unity.  For Mac users, the Unity plug-in described above handles setting up the command line arguments.
-
-To configure Unity to pass this data to VS Code, set the Editor Args to be `. -g "$(File):$(Line)"`.  You can see this step completed in the image above.
-
-**Q: How can I change the file exclusions?**
```
gregvanl: This FAQ is probably still valid. It sounds like Unity will try to create an appropriate files.exclude list in settings.json if it doesn't already exist. It could be the user already has settings.json that doesn't include files.exclude or they'd like to modify the default setup by Unity. I think it's fine to leave this and it is still accurate.

gregvanl:
Looks good, I just had the comment about leaving in the files.exclude FAQ.

docs/other/unity.md
eppz: It feels more VS Code related than Unity related to me. Makes no difference. I just prefer removing stuff over adding in general. 😄

gregvanl:
Thanks for reviewing the FAQ. I've ported your cleanup with https://github.com/Microsoft/vscode-docs/commit/9adc136ef068d09f73a06061c575c3f2a88185c9. I'll publish the changes live shortly.

eppz:
It is merely technical, but shouldn't my account be showed up among contributors to the file [**unity.md**](https://github.com/Microsoft/vscode-docs/blob/master/docs/other/unity.md)?

eppz:
I mean I don't know how is this exacly supposed to happen on GitHub.

gregvanl:
There is something funny going on with your commit. I see your commit in unity.md history but the commit is by "eppz!" with an exclamation point and your GitHub account picture doesn't show up. Could the local account you are using for your commit be different from your GitHub sign in which generated the PR?

eppz:
Thanks for the clarification probably that was the case.
Do you see it correct in https://github.com/Microsoft/vscode-docs/pull/961?

PR #961, opened 2017-04-26
Update "Unity Development with VS Code" page...

eppz:
...(removed some obsolete image assets).


gregvanl:
Thanks for the image cleanup PR. I deleted the unused images myself with https://github.com/Microsoft/vscode-docs/commit/1429665e3864a7d506de7b7c9548255d200ed4fb since this PR also had the unity.md cleanup and I wanted to leave in the files.exclude section.
Really appreciate you taking time to update this topic.

eppz:
Aw, seems this workaround - https://twitter.com/_eppz/status/846859856787259392 - is still relevant (Unity does not setup any build support automatically).

Can we embed the tweet somehow? Should I use simple HTML, will the page render it?
I could put it into the common question section.

gregvanl:
@eppz Sorry for the delay getting back to you. Rather than embed the tweet, I think we could just add a Common Questions entry detailing the steps in words and show a fenced code block of the launch.json using xbuild. Something like "Can I build my solution with XBuild in VS Code?"
Do you want to write this up? I can also put something together but it would be next week.

PR #962, opened 2017-04-27
Add snap install instructions

flexiondotorg:
This pull request adds installation instructions for users of [snap enabled Linux distributions](https://snapcraft.io/docs/core/install). Snaps enable ISVs and projects to make their software available to millions of Linux users via the snap store and directly control delivery of software updates to their users. Snaps of stable desktop applications will also be available via the desktop Software Center, improving the discoverability of your software. Your software can be installed with a simple `snap install vscode --beta --classic` simplifying the installation instructions for your users.

Once the snap is released into the stable channel in the store, users no longer need to add the ```--beta``` and will appear in the desktop Software Center

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -16,6 +16,7 @@ MetaDescription: Get Visual Studio Code up and running on Linux.
 Install the snap on [snap supported](https://snapcraft.io/docs/core/install) systems from the store.
 
     snap install vscode --beta --classic
+    sudo snap alias vscode code
 
 ### Debian and Ubuntu based distributions
 
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -11,6 +11,12 @@ MetaDescription: Get Visual Studio Code up and running on Linux.
 
 ## Installation
 
+### Ubuntu and [snap supported distros](https://snapcraft.io/docs/core/install) 
+
+Install the snap on [snap supported](https://snapcraft.io/docs/core/install) systems from the store.
+
+    snap install vscode --beta --classic
+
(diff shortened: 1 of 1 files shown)


docs/setup/linux.md
```diff
@@ -11,6 +11,12 @@ MetaDescription: Get Visual Studio Code up and running on Linux.
 
 ## Installation
 
+### Ubuntu and [snap supported distros](https://snapcraft.io/docs/core/install) 
+
+Install the snap on [snap supported](https://snapcraft.io/docs/core/install) systems from the store.
+
+    snap install vscode --beta --classic
```
Tyriar: We would need to have the package named `code` as all our documentation references `code as the CLI, not `vscode`.

flexiondotorg: The name `code` is already taken in the snap store. I'll see if I can claim it for Microsoft Visual Studio Code but for the time being I've updated the snap to support the alias `code`.

Tyriar:
@flexiondotorg just following up on how the package would actually be distributed.

> I don't believe you explained how the packages are submitted? Not owning the upload ourselves would probably mean the release would lag behind the official release, no?

Tyriar:
Never mind, just saw https://github.com/Microsoft/vscode/issues/5458#issuecomment-300742444 and replied there

duttaditya18:
Getting the ownership of `code` in the snap store will take some time. In the mean time I think this can be merged.

gregvanl:
snap package now officially supported. https://snapcraft.io/code

Documentation updated at https://code.visualstudio.com/docs/setup/linux#_snap

PR #979, opened 2017-05-03
Theme setting documentation

aeschli:
For the workbench theming effort we now allow to configure a large set of colors.
These colors can be set in color themes as well as in the user settings (under `"workbench.colorCustomizations"`).

This is my attempt of documenting these colors.
The list is currently in https://github.com/Microsoft/vscode-docs/blob/vnext/docs/getstarted/themes.md#L70 but I guess this adds too much content to that doc. My PR creates a new page, but I would let @gregvanl decide where to include it.


aeschli:
replaced by #980

PR #983, opened 2017-05-05
Fix a typo.

TransformersWsz:


gregvanl:
@TransformersWsz Thank you for interest in our docs. Your change is correct but I think the current wording is better.

PR #995, opened 2017-05-10
Link to JSON schema changed to http://json-schema.org/

borekb:
I'm new to JSON schemas and after visiting both

- https://spacetelescope.github.io/understanding-json-schema/ (the original link)
- http://json-schema.org/ (the proposed change)

the later one felt much more useful and had these [helpful examples](http://json-schema.org/examples.html).

### DIFF
--- a/docs/languages/json.md
+++ b/docs/languages/json.md
@@ -46,7 +46,7 @@ You can format your JSON document using `kb(editor.action.formatDocument)` or **
 
 ## JSON Schemas & Settings
 
-To understand the structure of JSON files, we use [JSON schemas](https://spacetelescope.github.io/understanding-json-schema/). JSON schemas describe the shape of the JSON file, as well as value sets, default values, and descriptions.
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @borekb, Thank you for your PR and pointing out the json-schema.org site. Rather than replace the current link to the spacetelescope site (which is called out as an excellent guide from the json-schema.org landing page), I've added an extra sentence directing new users to jscon-schema.org. 
See https://github.com/Microsoft/vscode-docs/commit/00af100948c8079bade81c79890cb75b03c68e8f and let me know if you would like anything changed.

borekb:
json-schema.org is the canonical site about JSON Schema, it hosts the specification and contains useful examples so I think it should go first. The link to https://spacetelescope.github.io/understanding-json-schema/ should be second, IMO. (If present at all; as a newbie, I can say that the description there was not very helpful. It was lot of text but I couldn't get my first schema working.)

Of course all this is subjective but when linking to "Microsoft", I would expect the link to be https://www.microsoft.com/, not https://en.wikipedia.org/wiki/Microsoft :)

gregvanl:
You are right and I don't understand my earlier logic. If json-schema.org already has a link to the spacetelescope site, it is discoverable.

PR #996, opened 2017-05-11
Added warning about non-standard `$schema` mapping

borekb:
Related issue with an example and some discussion: https://github.com/Microsoft/vscode/issues/26407


aeschli:
Sounds good to me!

gregvanl:
Thanks!

PR #1003, opened 2017-05-17
Clarification for OpenSuse

kosmoit:
Added a line between initial two OpenSuse commands for install.    If system requests password for initial command, and the user had pasted both lines.  It advises incorrect password.   Even after putting the correct password, the second command is not executed. (repo is not added).    This change just tries to clarify that a little bit...

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -66,6 +66,7 @@ The yum repository above also works for openSUSE and SLE based systems, the foll
 
 ```bash
 sudo rpm --import https://packages.microsoft.com/keys/microsoft.asc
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks!

gregvanl:
Hi @kosmoit, thank you for your interest in our documentation. I reviewed your PR again and I am going to revert it. While your advice is correct, most Linux users would understand the requirements of using sudo to elevate privilege and so we don't need to mention it explicitly in our topic. There are also several other places where we show using two sudo commands together and we would need to add the advice there as well.

PR #1005, opened 2017-05-19
Fix sub-title for link

EbXpJ6bp:
It seems that a link(line:25 in this file) linked #advanced-customizations


gregvanl:
@EbXpJ6bp Thanks for pointing this out! I'll fix the link and leave the header title the same.

gregvanl:
Fixed with https://github.com/Microsoft/vscode-docs/commit/abd9f6dff4e180c9f714d01a7bcbd9a1a34957f1

PR #1011, opened 2017-05-24
Disable blue outline on scroll to top button

mjbvz:
**Bug**
Ugly blue outline on scroll to top button in release notes:

screen shot 2017-05-24 at 2 09 26 pm

**Fix**
Disable this outline


gregvanl:
Thanks!

PR #1016, opened 2017-05-30
Removed an extra call to updateWordCount

ytimenkov:
It looks unnecessary to call twice in the same function.
Besides, the second call has a comment.


gregvanl:
Thanks!

PR #1018, opened 2017-05-31
Reference the file with variable

mreyc:
Use variable ${fileBasenameNoExtension} instead of sample.html


gregvanl:
Hi @mreyc, Thank you for your interest in the VS Code documentation. This PR is a nice use of the variable but because this is an introductory example for Markdown users, I'd like to keep the example as simple as possible and use the filenames directly.

PR #1019, opened 2017-05-31
add info about how to use code-insider at cmd

samudiogo:
in propose with https://github.com/Microsoft/vscode/issues/27768


gregvanl:
Thanks!

PR #1020, opened 2017-06-01
updated with CSX info

filipw:
Omnisharp and VS Code support CSX files (C# scripts) for a few months now. Updated the docs accordingly.


gregvanl:
Thanks!

PR #1022, opened 2017-06-01
Edit pass on reactjs-tutorial

waderyan:
@gregvanl I did an edit pass. 

**Overall:** Fantastic! Easy to follow. Great for beginners to get going and experienced users to skim through to learn to make the most out of VS Code. Good use of images and code blocks. 

This PR is a few fixes I'd suggest. 

One other comment is on the title. "React Tutorial in VS Code" didn't feel right. Is it the preposition "in"? It is what we do for the Node.js Tutorial so might not be worth changing. It seems like we want something like "Using React in VS Code Tutorial" but that is too long and doesn't flow well. Worth discussing.

### DIFF
--- a/docs/nodejs/reactjs-tutorial.md
+++ b/docs/nodejs/reactjs-tutorial.md
@@ -16,7 +16,7 @@ MetaSocialImage: nodejs_javascript_vscode.png
 
 ## Welcome to React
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks!

gregvanl:
Agreed on the title. I changed it to "Using React in VS Code"

PR #1024, opened 2017-06-02
Adding recognition to Transifex localization contributors of May.

danyeh:
Adding recognition to Transifex localization contributors of May iteration release to Release Notes.


chrisdias:
@gregvanl general content looks great. can you work your editing magic?

PR #1033, opened 2017-06-09
Re-word overview ruler explanation

dsifford:
Small wording adjustment that I found slightly confusing on my first pass. 

This also takes care of a typo.


gregvanl:
Thanks!

PR #1049, opened 2017-06-19
Json-rpc specification not available on https

hakonrossebo:
Unreachable site for json-rpc - Changed back to http


gregvanl:
Thanks!

PR #1050, opened 2017-06-19
Missing video - Changed to related video

hakonrossebo:
Link to session from keynote was not available. Could not find the keynote so I found a breakout session from same conference with correct content.


gregvanl:
Thanks!

PR #1051, opened 2017-06-21
fixed url to open a file to line and column

otamir:
url to open a file to line and column uses ':', not '?'.


gregvanl:
Thanks!

PR #1058, opened 2017-06-29
Document contributes.problemMatchers

dlech:
Most of the text is copied from https://code.visualstudio.com/updates/v1_11#_tasks

I could not find any documentation on `problemPatterns`, so I am not sure what the usage is.


gregvanl:
Thanks!

PR #1060, opened 2017-06-29
Run docs images through imageoptim

mjbvz:
Passes all images in the docs through imageoptim to reduce file size. Average savings is about 10% without a change in quality


gregvanl:
Thanks!

PR #1069, opened 2017-07-03
Improve networking documentation

joaomoreno:
I've cleaned up and improved the networking/proxy documentation for VS Code.

@gregvanl This is supposed to be merged during July, it is **not** for June's release! The PR is created from the `proxy` branch to the `vnext` branch. I didn't know whether it should be merged to `master` or `vnext`.


gregvanl:
@joaomoreno If this describes new features or switches that won't be available until the July release, it should go into the vscode-docs vnext branch which will be merged and published at the time of the July release.

joaomoreno:
Ignore this.

joaomoreno:
Created #1073, which merges against vnext.

gregvanl:
Cool. I'll merge it to vnext after 1.14 content goes to master this week.

m451:
Reopen since workspace config options for proxy-bypass are not documented as the close of https://github.com/Microsoft/vscode/issues/25184 would suggest. In special it it not clear how proxy-bypass-list may be used as a configuration option in the workspace settings. 

_Also while testing I noticed that there seems to be an issue when the server that vscodes tries to contact is using a self-signed certificate. Note that the signing root has been added to the local machines trusted root CAs yet still vscode seems to ignore that. So, besides I'm missing something here: it would be great if "how to trust a self signed certificate on the proxy" or more general "hot to trust a self-signed certificate" would be documented as well._

joaomoreno:
`proxy-bypass-list` is not a configuration option, but a command line. It is documented here: https://code.visualstudio.com/docs/setup/network More information here: https://www.chromium.org/developers/design-documents/network-settings

SSL certificate trust is also addressed here: https://code.visualstudio.com/docs/setup/network#_ssl-certificates

m451:
Yes. that's exactly the issue. It should _also_ be an configuration option. Besides: the command line params don't work for me. They'll simply be ignored. e.g. 

`code --proxy-server="myproxy:8080" --proxy-bypass-list="*.someinternaldomain" --ignore-certificate-errors`

will do exactly nothing. No proxy will be set. No proxy-bypass will be set. And certificate errors (self-signed in my case) will not be ignored. 
Using the configuration parameters 

`"http.proxyStrictSSL": false,
  "http.proxy": "http://myproxy:8080"`

the proxy will work but now the issue is that we don't have a proxy-bypass configuration option there.
This was tested with vscode version

Version 1.20.1
Commit f88bbf9137d24d36d968ea6b2911786bfe103002
Date 2018-02-13T15:34:36.336Z
Shell 1.7.9
Renderer 58.0.3029.110
Node 7.9.0
Architecture x64

running on Windows Server 2012R2 with correct proxy settings configured within Windows aswell.

joaomoreno:
What exactly works and does not work? Which feature within VS Code?

PR #1075, opened 2017-07-05
Add manual instructions for adding VS Code to PATH

spence:
Adding `code` to my PATH didn't work using the Command Palette. Adding instructions to manually add this.


gregvanl:
Thanks! Your example was for Insiders, I'm going to change it to reflect Stable.

PR #1084, opened 2017-07-10
Update extension-manifest.md

Huachao:


gregvanl:
I think this is correct. You can set the qna value to a string to provide a website URL. The documentation could be clearer that it isn't the word "string".

gregvanl:
Updated the description with https://github.com/Microsoft/vscode-docs/commit/399f16e93d173e1b202916ec1f2e0eac8685f162

PR #1085, opened 2017-07-11
Update jsconfig.md

mudrz:
Add a paths example that utilizes pattern webpack aliases, because adding /* in the path key is non trivial and different than the webpack implementation


mudrz:
relevant to https://github.com/Microsoft/vscode/issues/30291

gregvanl:
Thanks!

PR #1091, opened 2017-07-14
Simplify Debian/Ubuntu installation instructions

LenzGr:
Instead of downloading, converting and copying the GPG key, simply
pipe the downloaded key into `apt-key add` to add the build key.

Signed-off-by: Lenz Grimmer <lenz@grimmer.com>


Tyriar:
We actually got advice from the apt team not to use apt-key as it's deprecated before adding the instructions https://github.com/Microsoft/vscode/issues/2973#issuecomment-266815997

PR #1093, opened 2017-07-16
add exitcrash argument to cleanly exit debugger

chesshacker:
The `--exitcrash` argument can save you from having to manually kill the nodemon process.

### DIFF
--- a/docs/nodejs/nodejs-debugging.md
+++ b/docs/nodejs/nodejs-debugging.md
@@ -260,14 +260,13 @@ Alternatively you can start your program `server.js` via **nodemon** directly wi
     "type": "node",
     "request": "launch",
     "runtimeExecutable": "nodemon",
+    "runtimeArgs": [ "--exitcrash" ],
     "program": "${workspaceRoot}/server.js",
     "restart": true,
     "console": "integratedTerminal",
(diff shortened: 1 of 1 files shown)


chesshacker:
Sorry. I thought this argument would prevent you from needing to kill the launched process, but it's only a partial fix. This fix only works as intended when no changes have been made to the watched files. If nodemon has restarted your application, then stopping the debugger fails to stop the process. I would love to find a solution to this that works all the time, but this isn't it.

I still like the `--exitcrash` argument because it saves me from having to kill the process some of the time... but the Tip I removed should probably be added back, and maybe it should be revised to explain that sometimes you must still kill the process manually.

Please let me know if you'd prefer I close this PR or adjust it. Thanks!

roblourens:
I didn't know about the `--exitcrash` argument, thanks for the tip. I can confirm that it saves me from having to kill nodemon but only when I haven't changed any files.

I would find it useful but I'd like to keep the recommended configuration simple, and don't want to explain that switch in the docs, so I'll close this for now. FYI @weinand, in case you'd like to keep this switch in the docs.

weinand:
Thanks for the tip. I didn't know either.
I'm planning a nodemon recipe, so I might mention it there.

PR #1097, opened 2017-07-19
Adds cmder specific details for integrated terminal

johnkors:
See https://github.com/Microsoft/vscode/issues/12006

### DIFF
--- a/docs/editor/integrated-terminal.md
+++ b/docs/editor/integrated-terminal.md
@@ -51,6 +51,21 @@ Correctly configuring your shell on Windows is a matter of locating the right ex
 "terminal.integrated.shell.windows": "C:\\Windows\\sysnative\\bash.exe"
 ```
 
+
+#### [Cmder](http://cmder.net/)
+Create a `vscode.bat` file in your cmder path with the following contents:
+```bat
+@echo off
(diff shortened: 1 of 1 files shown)


Tyriar:
@gregvanl the instructions for setting `cmd.exe` are going to be more complicated when 64-bit VS Code on Windows  is in stable. `C:\\WINDOWS\\Sysnative\\cmd.exe` is correct only on *32-bit processes* on *64-bit Windows*, everything else (32 on 32, 64 on 64) should use `C\\WINDOWS\\System32\\cmd.exe`.

Any ideas how to concisely indicate this here?

Also the Cmder section should probably go in common questions.

gregvanl:
@Tyriar We can explicitly call out and show an example for the most common case (64 on 64?). Users on other configurations should be able to adjust based on the example.

Tyriar:
@gregvanl well the most common case right now and likely for the next few months until people jump over to the 64-bit client is 32 on 64.

gregvanl:
Thanks! I move the tip to Common Questions. https://github.com/Microsoft/vscode-docs/commit/c4df2a2e57f320387a5ef396859da39433e85a36

johnkors:
:+1:

PR #1098, opened 2017-07-19
Add 1.14.2 update note

gregvanl:


gregvanl:
@waderyan @chrisdias Ready if we ship 1.14.2.

PR #1112, opened 2017-08-06
Simplify PGP key installation instructions

s0enke:
By using `apt-add-key` which is the de-facto standard on Debian/Ubuntu afaik.


gregvanl:
@s0enke Thank you for your PR and interest in the VS Code docs. 
We didn't take an earlier PR with this change as we were advised not to use apt-key.
See the comment in https://github.com/Microsoft/vscode-docs/pull/1091

PR #1113, opened 2017-08-08
Clarify what is waiting when --wait is used.

saschwarz:
When I read this I wasn't sure what a "launch command" meant. Assuming it's a shell command that starts VS Code, I added text to specifying why waiting for VS Code would be useful when using Git.


gregvanl:
Thanks!

PR #1114, opened 2017-08-09
Wrap settings in inline code blocks

luangong:


gregvanl:
Good suggestion. Thanks!

PR #1115, opened 2017-08-09
Wrap <img> tags in paragraphs

luangong:
This ensures that there are vertical margins between the images and their surrounding text.

For example, the following markdown:

```markdown
foo

<img src="http://www.example.com/some-image.png" />

bar
```

is rendered as the following HTML:

```html
<p>foo</p>
<img alt="some text" src="http://www.example.com/some-image.png" />
<p>bar</p>
```

while

```markdown
foo

![some text](http://www.example.com/some-image.png)

bar
```

is rendered as:

```html
<p>foo</p>
<p>
  <img alt="some text" src="" />
</p>
<p>bar</p>
```

We still use `` tags at some other places because we need finer control over the images to overcome the limitations of Markdown so we embed inline attributes (such as `style="..."`, `width="..."`, or `<center></center`) in those `` tags.

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -306,7 +306,7 @@ Instead of placing breakpoints directly in source code, a debugger can support c
 
 A 'function breakpoint' is created by pressing the **+** button in the **BREAKPOINTS** section header and entering the function name:
 
-<img alt="function breakpoint" src="https://az754404.vo.msecnd.net/public/function-breakpoint.gif" />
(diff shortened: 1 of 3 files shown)


gregvanl:
Thanks! Better to use straight Markdown than embed HTML

gregvanl:
Unfortunately I needed to rollback your PR as our Markdown -> HTML rendering can't handle URL image links.
The image ends up with a broken link:
restart frame

We could look into improving this but our .gif's are usually local.

PR #1121, opened 2017-08-10
Add Linux information to Linting section

Foxtrek64:
1. Added note to install full version of PHP if using XAMPP.
2. Divided settings.json additions into Windows and Linux.

I know that we cannot predict every configuration, so I'm only using the default values. If anyone thinks I should add a note about this, let me know.

### DIFF
--- a/docs/languages/php.md
+++ b/docs/languages/php.md
@@ -33,6 +33,8 @@ Visual Studio Code includes a set of common snippets for PHP.  To access these, 
 
 VS Code uses the official PHP linter (`php -l`) for PHP language diagnostics. This allows VS Code to stay current with PHP linter improvements.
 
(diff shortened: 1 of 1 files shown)


docs/languages/php.md
```diff
@@ -43,12 +45,22 @@ There are three [settings](/docs/getstarted/settings.md) to control the PHP lint
...
 
+### Windows
+
 ```json
 {
     "php.validate.executablePath": "c:/php/php.exe"
 }
 ```
 
+### Linux and OS X
```
gregvanl: We've switched to using "macOS" instead of "OS X" per Apple guidelines.

gregvanl:
Thanks! I'll update OS X to macOS.

PR #1127, opened 2017-08-14
Update nodejs-debugging.md

dominicwhyte:
Reflects deprecated --debug flag and using the inspector protocol on default port 9229

### DIFF
--- a/docs/nodejs/nodejs-debugging.md
+++ b/docs/nodejs/nodejs-debugging.md
@@ -173,12 +173,12 @@ lines="foo\nbar"
 If you want to attach the VS Code debugger to a Node.js program, launch Node.js as follows:
 
 ```
-node --debug program.js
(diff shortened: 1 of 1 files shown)


roblourens:
Thanks for pointing this out, but could you change the other usages in this doc too? I think all the places that reference `--debug` and `"port": 5858` on this page should be changed. Also, you shouldn't need to set `"protocol": "inspector"` for the attach config - it should be autodetected.

cc @weinand

gregvanl:
Thanks. I'll change the other --debug references as well.

gregvanl:
Finished update with https://github.com/Microsoft/vscode-docs/commit/54d3d82b98a3401fafa25aeefe6b1d0742f8fe9e

roblourens:
Also updated the node-debug defaults in package.json

PR #1138, opened 2017-08-19
Clarify wording about React's purpose and origin

clee:
The previous order of the first sentence made it a little ambiguous - is React only for web applications developed by Facebook? No.


gregvanl:
Fair enough. Thanks.

PR #1145, opened 2017-08-29
Wrap preferences in inline code blocks

luangong:


gregvanl:
Thanks. Good formatting suggestion.

PR #1147, opened 2017-08-29
Trim trailing whitespace

luangong:
This also ensures that trailing whitespace won’t be added in future commits by adding `.editorconfig` and changing `.vscode/settings.json`.


gregvanl:
Thanks.
One question: what does creating an .editorconfig do for vscode-docs? I didn't think VS Code uses .editorconfig.

luangong:
`.editorconfig` is for enforcing the same settings across different editors / IDEs. [vscode](https://github.com/microsoft/vscode) already uses EditorConfig and more and more other projects do. Everyone working on the same project should install the EditorConfig plugin for their editor (or maybe it's already built in). See http://editorconfig.org for more details.

gregvanl:
Cool. Nice noticing that vscode repo was already using these settings.

PR #1159, opened 2017-09-08
Adds a missing thank you

joaomoreno:
cc @eamodio


joaomoreno:
@gregvanl This is to get out still for August, there's a missing Thank You.

PR #1162, opened 2017-09-11
Update tasks.md

faabiopontes:
The default generated tasks.json is different than the one generated on the version 1.6 of VS Code.

Updated to be equal to the generated file.


gregvanl:
Thanks!

PR #1171, opened 2017-09-14
Fix outdated doc

chenxsan:
Looks like it's renamed according to this commit https://github.com/Microsoft/vscode-generator-code/commit/a88ba5be8581055123d35b5f7108643b94df0828


amadeann:
Here's another commit showing it was indeed changed:

https://github.com/Microsoft/vscode/commit/4a13944af6872446f83cf7e289696a8d6d859358

gregvanl:
Thanks!

PR #1179, opened 2017-09-23
Fix file names in both tasks templates

LWChris:
The file names in the tasks do not match the example


gregvanl:
Thanks! People have been saying the examples don't work but we'd never noticed the filenames.

PR #1182, opened 2017-09-25
remove extraneous comma from example

chaseadamsio:
The extraneous comma would cause using the example to fail in a JSON theme definition.


gregvanl:
Thanks!

PR #1183, opened 2017-09-25
java doc and blog

hexiaokai:


docs/languages/java.md
```diff
@@ -0,0 +1,105 @@
...
+---
+# Java in VS Code
+
+Java is fully supported in Visual Studio Code through [extensions](/docs/editor/extension-gallery.md). Popular extensions in the [Marketplace](https://marketplace.visualstudio.com) provide project support, code completion, linting, debugging, code formatting, snippets, and more.
+
+> [Download VS Code](/download) - If you haven't downloaded VS Code yet, quickly install for your platform (Windows, Mac, Linux).
+
+## Install Java Extensions
+
+VS Code is a fast editor and ships with only the basic features. Add Java language support to VS Code by installing the popular Python extensions.
```
gregvanl: popular Java extensions

docs/languages/java.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. Select an extension.
+2. Install the extension by typing `ext install` into the Command Palette `kb(workbench.action.showCommands)`.
+
+<div class="marketplace-extensions-java"></div>
```
gregvanl: This query doesn't exist. Do we want to have a curated list, run a query or leave out these tiles all together.

hexiaokai: Seems like those kind of query is commonly used in other language's pages, such as php and python. I think we don't need a curated list here, just leave out all Java related tiles.

PR #1186, opened 2017-09-26
Styling issue (minor fix proposal)

kant:


gregvanl:
Thanks for noticing and fixing!

PR #1191, opened 2017-10-02
Loc/v117: Adding September update release localization contributors.

danyeh:
Adding September update release localization contributors.


gregvanl:
Thanks!

gregvanl:
@danyeh We had trouble rendering the in-product release notes when the Markdown file included a BOM so I've switched the file encoding back to UTF 8 not UTF 8 with BOM.

PR #1193, opened 2017-10-03
Small typo fix for 'docs/extensions/themes-snippets-colorizers.md'

svekl:
Looks like it must be `more than that` instead of `more than than`


gregvanl:
Thanks!

PR #1205, opened 2017-10-15
Update color sublime url

imagentleman:
The domain of color sublime was recently changed (https://github.com/Colorsublime/Colorsublime-Themes/issues/364) in favor of a github.io (https://colorsublime.github.io) alternative.

And since they went fully static, the download links also changed to just point to the raw files on the repo.


gregvanl:
Thanks!

PR #1212, opened 2017-10-19
Update nodejs-debugging.md

dlindeque:
There are many languages transpiling to JavaScript. VSCode does not have support for all of them!


gregvanl:
Fair enough. Thanks.

PR #1215, opened 2017-10-21
Fix typo of word 'currently'

vnbhattarai:
Replace **current** with **currently** on line 415


gregvanl:
Hi @vnbhattarai, This text comes from the description field of the workbench.editor.openPositioning setting and is just copied directly into the settings.md document.  It should be fixed in the vscode product repo at https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/electron-browser/main.contribution.ts. Perhaps you'd like to create a PR for this?

PR #1219, opened 2017-10-24
Remove possible error in comment

machira:
Who is Ticino?


gregvanl:
Too funny, Ticino is the old product code name. Thanks!

PR #1224, opened 2017-10-26
Update reactjs-tutorial.md

jsephora:
just typing "code ." doesn't work for me. I am using Mac OS Sierra terminal. It worked after I found this solution on Stackoverflow https://stackoverflow.com/questions/29971053/how-to-open-visual-studio-code-from-the-command-line-on-osx
I fear I may have to do this EVERY time though, which would be a hassle.

### DIFF
--- a/docs/nodejs/reactjs-tutorial.md
+++ b/docs/nodejs/reactjs-tutorial.md
@@ -49,6 +49,11 @@ To open your React application in VS Code, open another terminal (or command pro
 cd my-app
 code .
 ```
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @jsephora, thank you for your interest in our documentation and your PR. We cover adding the code binary to various operating system paths in other topics ([macOS](https://code.visualstudio.com/docs/setup/mac#_launching-from-the-command-line)) so I don't want to update the React and other tutorials with this step.

PR #1230, opened 2017-10-31
fix(docs): unfold/fold

jpoon:
Add `direction` parameter. As an aside, is there a strategy in place to keep these documents up-to-date with the APIs?

### DIFF
--- a/docs/extensionAPI/vscode-api-commands.md
+++ b/docs/extensionAPI/vscode-api-commands.md
@@ -223,6 +223,7 @@ See [working with the HTML preview](/docs/extensionAPI/vscode-api-commands.md#wo
   Property-value pairs that can be passed through this argument:
 
   * 'level': Number of levels to unfold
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @jpoon, this content is autogenerated from the product. I don't know why this isn't being generated accurately. I'll investigate but the fix needs to come from the product vscode repo.

gregvanl:
Fixed with https://github.com/Microsoft/vscode-docs/commit/d12bdecffcccb5842f8077b245c5b3461056f386 generated from 1.18 sources. The topic will be updated on the website refresh for the 1.18 release. 
Thanks for calling out the discrepancy.

jpoon:
Thanks @gregvanl. Also noticed a slight typo here: https://github.com/Microsoft/vscode-docs/commit/d12bdecffcccb5842f8077b245c5b3461056f386

PR #1246, opened 2017-11-09
Multi-root is no longer Insiders-only

borekb:
Multi-root shipped in 1.18, removed note about Insiders only


gregvanl:
Thanks for being on top of this!

PR #1247, opened 2017-11-09
This feature is in the stable release since 1.18

JohnArcher:
The remark that this feature is just for insiders is not valid anymore as it was introduced in stable 1.18.


gregvanl:
Thank you for noticing. This is a duplicate of #1246. Thanks to our community for being on top of it!

PR #1251, opened 2017-11-09
A problem I faced and now found the solution

Hassuunna:
after installing github, vs code suggests changing the default terminal to bash instead of powershell.
I clicked OK thinking it's easy to change in terminal window by some button but I didn't find how to change it even through reading this document

so propose file change is about how to customize default terminal


gregvanl:
Thanks for the Common Question idea. I will do a little editing of your question and answer.

PR #1253, opened 2017-11-10
Fix for keybinding improperly rendering in VS Code view of Release notes

sbatten:
After looking into how `kbstyle` gets applied and other uses of it, I could only guess that the spaces were messing up the way the keybinding was rendered when viewing the release notes from within VS Code. 

On Windows this shows `Ctrl Unknown`
On Mac this shows `unassigned`

If this isn't the proper fix, please close.

### DIFF
--- a/release-notes/v1_18.md
+++ b/release-notes/v1_18.md
@@ -246,7 +246,7 @@ These actions are:
 
 ### Start without debugging shortcut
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks! I think you are right.

gregvanl:
Published and working correctly on Windows.

PR #1256, opened 2017-11-11
Changing link to Python extension

Rbeuque74:
Python extension is now under Microsoft namespace, changing link to avoid 404 Not Found


gregvanl:
Wow, you're quick! Thanks for noticing this.

PR #1259, opened 2017-11-13
Fix example using a variable in a snippet

ebeloded:
Single escape character is not accepted, whereas double works fine


gregvanl:
Thanks!

PR #1270, opened 2017-11-17
Fixing the broken hyperlink for "Java Extension Pack"

hakaneskici:
This PR fixes the URL for "Java Extension Pack" link which had a typo in the URL (**hhttps** instead of https). This results in having a broken link on the following page: https://code.visualstudio.com/docs/java/java-debugging (search for "Java Extension Pack" on that page). It doesn't repro when the markdown file is viewed on GitHub.


gregvanl:
Thanks!

gregvanl:
It may be a couple of days before we publish this change to our live site.

PR #1273, opened 2017-11-23
Change typo in autopep8Args

bellert:
Specify autopep8Args with "=" between argument and value instead of space


gregvanl:
Hi @bellert, thank you for your PR. While you're approach will work, I spoke with @DonJayamanne, the Python extension author, and he recommends using separate strings for each argument, whether they are a setting name or value.

gregvanl:
https://github.com/Microsoft/vscode-docs/commit/af48ff5532e4101048c61cfd4e065a11345eaffd

PR #1277, opened 2017-11-28
Path export would return an error in terminal

denisinla:
Spaces were not escaped correctly in path export causing an error of path not being found when refreshing your shell.


gregvanl:
Thank you for your interest in the VS Code documentation and submitting a PR. The existing example works for me and I didn't need to add the escape characters. I wonder what it is about your macOS environment that requires escaping the spaces in "Visual Studio Code".

Ben-Kincaid:
@gregvanl This is also happening to me when manually adding the path. Regardless, using spaces in the filenames doesn't seem like good practice.

PR #1279, opened 2017-11-28
Specify ptvsd 3.0.0 or later is required for remote debugging in Python

brettcannon:
Closes microsoft/vscode-python#299

### DIFF
--- a/docs/python/debugging.md
+++ b/docs/python/debugging.md
@@ -137,7 +137,7 @@ To debug an app that requires administrator privileges, use `"console": "externa
 
 Remote debugging allows you to step through a program locally within VS Code while it's executed on a remote computer. In this case, it's necessary to have the source code on both computers.
 
-1. On both development and remote computers, install the [ptvsd library](https://pypi.python.org/pypi/ptvsd).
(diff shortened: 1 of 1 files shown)


docs/python/debugging.md
```diff
@@ -137,7 +137,7 @@ To debug an app that requires administrator privileges, use `"console": "externa
 
 Remote debugging allows you to step through a program locally within VS Code while it's executed on a remote computer. In this case, it's necessary to have the source code on both computers.
 
-1. On both development and remote computers, install the [ptvsd library](https://pypi.python.org/pypi/ptvsd).
+1. On both development and remote computers, install the [ptvsd library](https://pypi.org/project/ptvsd/) (version `3.0.0` or later).
```
gregvanl: Are you fine with this link going to a pre-production deployment of the Warehouse? Is that only temporary and this will be the final URL as well?

brettcannon: Yes and yes. 😄 It's only "pre-production" due to some management features not being available, otherwise API traffic is already going there. Plus the project just got funding from Mozilla to finish the work that has led to that label being there.

PR #1282, opened 2017-11-30
Add fenced code block language identifier description

MikeRatcliffe:
The Markdown Syntax page does not include a list of supported languages that can be used as code block language identifiers. At the moment the only way to get that list is by reading the source code, which is a bit of a pain.

These definitions were gleaned from `vscode/extensions/markdown/syntaxes/gulpfile.js` on the latest master branch so they are current.


gregvanl:
@MikeRatcliffe Thank you for your interest in the VS Code documentation and your PR. VS Code is using the 'markdown-it' library to render not 'marked'. We'd also rather not publish and maintain the full list of code block supported language identifiers.

PR #1283, opened 2017-11-30
Replaced curl with wget

szaroubi:
Not sure that curl comes in by default on an ubuntu server.
The code snippet I changed doesn't work on my Ubuntu 17.10 server vanilla install.


Tyriar:
@szaroubi when I tried this locally wget spat out an error regarding my ~/.netrc which would cause issues with the key.

PR #1292, opened 2017-12-05
Link to active issue for Windows title bar theme

Templarian:
Updating the title bar section to include the discussion issue concerning title bar progress on the Windows platform.

Might clear up some confusion when reading this section without adding a lot explanation inline.


gregvanl:
Hi @Templarian, thank you for your interest in the VS Code documentation and submitting a PR. I'm not going to merge your PR because we don't usually include issue links in the main docs. You mention confusion with the title bar color section, what text might help it without linking to an issue?

Templarian:
That's okay, was pretty minor, plus I think people will be able to find it with a quick Google search (it's the first result). Thanks for the followup. 😸

PR #1302, opened 2017-12-14
improve the cmder terminal

iLuffy:
The cmder terminal always change back to the current user directory, so record the working directory and change back to the working directory after cmder init.


Tyriar:
Looks good, thanks!

PR #1303, opened 2017-12-14
themes-snippets-colorizers.md: yo code setup

waldyrious:
Provide the actual commands for running the Yeoman generator,
as is done in docs/extensions/example-hello-world.md


gregvanl:
Thanks!

PR #1307, opened 2017-12-18
Update git.autofetch documentation

joaomoreno:
@gregvanl I'd like this to go out asap. I will have a link to this paragraph in the November candidate release. #40282


gregvanl:
We use "VS Code" rather than "Code" to refer to the product. I'll fix this up next week when I'm back from vacation.

PR #1319, opened 2017-12-24
Fix link to AUR

Tyriar:
@dcelasun what was the reason for the change?


dcelasun:
@Tyriar Someone higher up insisted that a binary package must have a `-bin` suffix :/

Tyriar:
@dcelasun ah OK, no worries 😃

PR #1324, opened 2017-12-29
Tasks: document use of config settings

acdha:
I recently needed to run some tasks using a specific Python virtualenv
and hadn't known that this was possible before searching through the
issue tracker and finding a maintainer's comment:

https://github.com/Microsoft/vscode/issues/23597#issuecomment-290329622


gregvanl:
Thanks! Nice addition

Issue #85, opened 2015-12-17
Error in the SASS/LESS -> CSS guide


aeschli:
_From @malisbad on December 17, 2015 4:53_

There is an error in your SCSS/LESS - > CSS guide. The task shouldn't have the ">" in the task runner (anymore?). If you follow the guide, it writes the compiled .css to the ">". If you remove it, it writes to "styles.css" correctly.

_Copied from original issue: Microsoft/vscode#1401_

aeschli:
_From @dbaeumer on December 17, 2015 8:49_

@aeschli not sure if this is you

aeschli:
This is about the 'Transpiling Sass and Less into CSS' section in https://github.com/Microsoft/vscode-docs/blob/master/docs/languages/css.md
Not sure who authored that, @gregvanl ?

gregvanl:
Will be fixed in the December release documentation.

PR #1326, opened 2017-12-30
Minor fix on #85: jsbeautifiery 

kant:


gregvanl:
Hi @kant, Thanks for spotting the error. I think a better link would be to the js-beautify npm module documentation. [js-beautify](https://www.npmjs.com/package/js-beautify).
Can you change the text to say:
```md
The HTML formatter is based on [js-beautify](https://www.npmjs.com/package/js-beautify).
```

gregvanl:
@kant Thank you again for pointing out the typo. I've manually made the fix.

Issue #514, opened 2016-08-12
JavaScript - Consistent theme


waderyan:
We should have a consistent theme for screenshots on this page. I recommend we use the dark theme.

waderyan:
Fixed with PR #542

Issue #133, opened 2016-01-28
Extensions Intro Video


waderyan:
There is a very helpful video on [channel 9](https://channel9.msdn.com/Events/Visual-Studio/Connect-event-2015/Visual-Studio-Code-Extensions) for building an extension. It would be great to link to this somewhere in the docs.

gregvanl:
There is resistance to linking out from our docs to Channel 9 for several reasons:

1. We want traffic to come to our site, not leave.
2. Channel 9 content can become stale and we don't control it.
3. Channel 9 content quality varies.

PR #1329, opened 2018-01-03
Only PTVSD version 3.0.0 is supported at present.

dmk255:
See issues [#514](https://github.com/Microsoft/vscode-python/issues/514) and [#133](https://github.com/Microsoft/vscode-python/issues/133)


gregvanl:
Thanks!

Issue #1280, opened 2017-11-29
Create Refactoring topic with examples of TypeScript and JavaScript refactorings


mjbvz:
Add a new section in the docs overviewing refactoring support in TS and JS

gregvanl:
New Refactoring topic added https://github.com/Microsoft/vscode-docs/commit/dcf6703edcb33e76f61008c1a18fee6d7b69fd53

PR #1334, opened 2018-01-04
Better document included code lenses for js/ts

mjbvz:
- Documents the implementations code lens for TS
- Documents the references code lens for JS


mjbvz:
@gregvanl Is there a better place to document features that are shared between JavaScript and TypeScript? This PR adds a little duplication but changes #1280 may add a lot more

PR #1338, opened 2018-01-08
Add note to differentiate btw TextDoc & TextEditor

seanpoulter:
These are some of the things I wish I had known before using the `TextDocument` event listeners.

cc @jrieken re: Microsoft/vscode#40653


jrieken:
@seanpoulter Thanks but can you please make these changes in the vscode-repro in the `vscode.d.ts` file. The file you have changed here is generated then. Thanks

gregvanl:
I'll also take this PR so that the API topic can be updated now as we won't update the latest vscode.d.ts until the next release at the end of the month.

seanpoulter:
Will do @jrieken.

gregvanl:
My mistake, it looks like any direct changes to vscode-api.md will be overwritten on the next doc build, so this change won't be published until the next release.

PR #1343, opened 2018-01-10
Fix snippet example prefix inconsistency

deini:
Fixed inconsistency with text below:

> `prefix` defines how this snippet is selected from IntelliSense and tab completion. In this case `for`.


gregvanl:
Thanks for noticing and submitting a PR.  PR https://github.com/Microsoft/vscode-docs/pull/1342 beat you to it by only a few hours.

deini:
Nice timing 😄 , thanks for letting me know.

PR #1345, opened 2018-01-11
Change the path of extension.ts

cuspymd:
The path of the `extention.ts` file in the generated code is different from the document.


gregvanl:
Thanks for noticing and fixing!

PR #1346, opened 2018-01-11
Improve Cmder common question

xavero:
VsCode use the integrated terminal to run Tasks Builds and expects the terminal to exit after executing.
This requirement was not accounted in Cmder bat script.

### DIFF
--- a/docs/editor/integrated-terminal.md
+++ b/docs/editor/integrated-terminal.md
@@ -164,6 +164,10 @@ The basics of the terminal have been covered in this document, read on to find o
 
 ```bat
 @echo off
+IF "%*" NEQ "" ( 
+    CMD %*
+    EXIT %ERRORLEVEL% 
+)
 SET CurrentWorkingDirectory=%CD%
 SET CMDER_ROOT=C:\cmder (your path to cmder)
 CALL "%CMDER_ROOT%\vendor\init.bat"
--- a/docs/editor/integrated-terminal.md
+++ b/docs/editor/integrated-terminal.md
@@ -164,8 +164,10 @@ The basics of the terminal have been covered in this document, read on to find o
 
 ```bat
 @echo off
-IF "%*" NEQ "" ( 
-    CMD %*
+SET ARGS=%*
+IF "%ARGS:~0,5%" EQU "/d /c" ( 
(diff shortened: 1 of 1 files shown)


docs/editor/integrated-terminal.md
```diff
@@ -164,6 +164,10 @@ The basics of the terminal have been covered in this document, read on to find o
 
 ```bat
 @echo off
+IF "%*" NEQ "" (
```
Tyriar: Is this asking whether there are arguments with the command? Does CMD then run is under cmd.exe?

xavero: Yes for both. Code calls the terminal with the arguments: "/d /c command_name command_args"

The /c flag indicates that cmd.exe should run and terminate

xavero: Maybe would be better to check first if the arguments passed were in fact "/d /c" ? I will redo the PR.

Tyriar: But doesn't cmder run some type of bash, not cmd? Is this meant to be a way to work around https://github.com/Microsoft/vscode/issues/36216?

xavero: Yes, it felt strange to run CMD, so in the new commit I changed to just run the command, ignoring the _"/d /c"_ part of string. I don't think Cmder run a special bash or anything. 
I wasn't aware of the issue, but it seems to be caused by the same flags.

docs/editor/integrated-terminal.md
```diff
@@ -164,6 +164,12 @@ The basics of the terminal have been covered in this document, read on to find o
 
 ```bat
 @echo off
+SET ARGS=%*
+IF "%ARGS:~0,5%" EQU "/d /c" ( 
+    REM Tasks call the terminal expecting it runs and exits
+    %ARGS:~6%
```
xavero: In this new commit, I'm calling the desired Task command directly in the terminal, removing the _"/d /c"_ flags that are not applicable in here.

Tyriar:
I think I'm going to close this off as it's related to a _tasks_ edge case, not just the terminal.

@dbaeumer FYI since this is related to running tasks through a terminal that runs the cmder bat init script. You may want to incorporate into the tasks docs?

PR #1359, opened 2018-01-19
Corrected page title from VS to VS Code

amosharper:
Erroneous page title 'Visual Studio Tips and Tricks' is confusing to users finding this page via a search engine. Corrected to 'Visual Studio Code Tips and Tricks'.


gregvanl:
Thanks! Yes, that would be confusing.

PR #1360, opened 2018-01-20
Updated required .NET Framework version

steffen-s:
When trying to run VS Code on a PC with Windows 7 (64 bit) that only had .NET Framework 4.5.2 installed (as per this guide), an error popped up when opening VS Code stating that the required .NET Framework version is 4.6. After updating .NET Framework on the PC, VS Code ran without issues. Hence the higher version requirement for the .NET Framework should be reflected in the documentation.

### DIFF
--- a/docs/setup/windows.md
+++ b/docs/setup/windows.md
@@ -17,7 +17,7 @@ MetaDescription: Get Visual Studio Code up and running on Windows
 
 You can also find a Zip archive [here](/docs/?dv=winzip).
 
->**Note:** .NET Framework 4.5.2 or higher is required for VS Code.  If you are using Windows 7, make sure you have at least [.NET Framework 4.5.2](https://www.microsoft.com/en-us/download/details.aspx?id=42643) installed.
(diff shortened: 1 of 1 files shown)


steffen-s:
Related to: [There is no error to prompt the user when no .NET Framework is installed](https://github.com/Microsoft/vscode/issues/41915)

gregvanl:
Hi @steffen-s Thank you for your PR. Interestingly, the 4.6 requirement is coming from the C# extension see the changelog.md entry https://github.com/OmniSharp/omnisharp-vscode/blob/ee50939cdb35b169f00b6739f5e8b7d25daa364f/CHANGELOG.md#152-november-15-2016 and not from VS Code itself. I believe VS Code will work correctly with just 4.5.2. installed.

steffen-s:
Thanks for looking into this @gregvanl, it is much appreciated. That is interesting indeed - and it will also teach me to test VS Code without extensions next time. I do apologize for making you spend your time on doing the research I should have done. I will close this pull request now.

gregvanl:
Not a problem. We really appreciate people taking an interest in our documentation. Requirements change so it's great to revisit the docs to make sure they are accurate. Thanks.

PR #1361, opened 2018-01-21
Clarify prerequisites for Windows users

fredsa:
Update prerequisites to address OSX and Windows users, with O/S specific steps.


gregvanl:
Thanks!

PR #1364, opened 2018-01-24
Added two hostnames to network.md

RenatoPassos:
Added `raw.githubusercontent.com` to the section "Common hostnames" of network.md, since access to that host is needed so you can see the screenshots and pictures in the extensions pages at both VS Marketplace website and at the VS Code "Extensions" view.

Also added `vsmarketplacebadge.apphb.com` so the Marketplace Status Badges can be seen inside the VS Code "Extensions" view.


gregvanl:
Thanks!

RenatoPassos:
You're welcome!

PR #1379, opened 2018-02-02
Update "App Service Tools" usage.

fiveisprime:
Removed the occurrences of "tools" as it's redundant and will be removed from
the official extension name.


gregvanl:
@fiveisprime I also updated the name in the Docker tutorial https://github.com/Microsoft/vscode-docs/commit/06a6072bed73757df6ffd3f90457536a5a08feb7

fiveisprime:
Looks like the Docker extension fell off the list somehow

image

PR #1385, opened 2018-02-07
Obsolete allowJs in compilerOptions

rjmreis:
Docs still reference `allowJs` which has been replaced by `checkJs`.

### DIFF
--- a/docs/languages/jsconfig.md
+++ b/docs/languages/jsconfig.md
@@ -15,7 +15,7 @@ The presence of `jsconfig.json` file in a directory indicates that the directory
 
 > **Tip:** If you are not using JavaScript, you do not need to worry about `jsconfig.json`.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @rjmreis, thank you for your interest in the VS Code documentation and submitting a PR. 
I believe the documentation is correct as allowJs is a valid attribute for tsconfig.json. 
Do you think the Note should more clearly point out that it is talking about tsconfig.json, not jsconfig.json?

rjmreis:
@gregvanl I believe you are right, on a more detailed reading. Perhaps if the props had the same name it would be more obvious.

gregvanl:
@rjmreis tsconfig.json has both allowJS and checkJS, jsconfig.json just has checkJS.

PR #1388, opened 2018-02-09
[Live Share] Adding FAQ question about tooling requirements

lostintangent:
Since starting the private preview, we've had the question about tooling requirements come up frequently enough that it makes sense to address it in our FAQ.


lostintangent:
Merging due to getting a sign-off from the team "offline".

PR #1390, opened 2018-02-12
more clearly hint to use login, package before publish

paulvi:
In current Usage section https://code.visualstudio.com/docs/extensions/publish-extension#_usage
it is too much about publishing, and it skipped 2 essential steps before publishing


gregvanl:
Thanks!

PR #1404, opened 2018-02-16
Remove outdated screenshot, fixes #1403

### LINKED ISSUE #1403: Outdated screenshot in angular javascript tutorial
The documentation says to click an information icon to show a flyout. In the current version of code, there's no information icon, the flyout for the selected item is automatically shown.
screen shot 2018-02-16 at 12 59 24 pm

RMacfarlane:


jens1o:
Instead of removing it, you could replace it by an up2date screenshot in my opinion.

RMacfarlane:
@jens1o I agree we should update screenshots whenever possible, but in this case the screenshot shows a button to expand something in the UI that is now automatically expanded. The next step in the tutorial shows the automatically expanded UI, so there is no need for the first screenshot

gregvanl:
The flow is correct on the first run case when the flyout hasn't been triggered before. The flyout state is sticky across sessions and you may have already expanded it once.

PR #1416, opened 2018-02-18
Make "repofile create command" nicer and more readable

ruzickap:
Make "repofile create command" nicer and more readable


Tyriar:
Hi @ruzickap I think this pushes the size of the page out unnecessarily since generally people won't be reading this script line.

ruzickap:
Ok... Actually I prefer to see what is going go be written and where - especially if I run it as root. 
But I understand your philosophy to make it easy copy&paste for the beginners who do not care much what are "executing"...

PR #1457, opened 2018-02-26
Precise the file to update for changing options

ibensalah:
This precision could avoid confusion for new VS Code or TypeScript users between the options that has to be changed in the User Settings of VS Code and the one that has to be changed in the tsconfig.json file.


gregvanl:
Good clarification. Thanks!

ibensalah:
You're welcome :)

PR #1460, opened 2018-02-27
Fix typo nupad_separator -> numpad_separator

bantic:
As named in vscode: https://github.com/Microsoft/vscode/blob/728f71d1bf7f7f99ce58fee8206ac1848268892d/src/vs/base/common/keyCodes.ts#L356


gregvanl:
Thanks!

PR #1472, opened 2018-03-05
Adding February update release localization contributors.

danyeh:
Adding February update release localization contributors.


robert1ishere:
With all due respect how do I quit following or get off of this thing I
thought I could learn something but hell I can't even be in the
administrator of my I was kind of hoping to learn but you guys are way too
advanced,...  I look for a place to unsubscribe but I can't seems like it
doesn't exist

gregvanl:
Hi @robert1ishere, I may not understand your situation but perhaps you need to click the Unwatch button on GitHub and select the Ignoring, Never be notified option.
If you do that you shouldn't get any more email notifications from the vscode-docs repo.

PR #1477, opened 2018-03-07
Change the git repository to upstream repository

kulikala:
The forked repository (https://github.com/microsoft/gs-spring-boot) used in the Java tutorial does not work well on current Spring Boot.
The upstream one (https://github.com/spring-guides/gs-spring-boot) worked well in my environment.


gregvanl:
Thanks! Yes, we should be pointing to the spring-guides repo as mentioned earlier in the paragraph.
Adding @hexiaokai for visibility.

PR #1482, opened 2018-03-08
Fix Electron version numbers in Node.js debugging guide

zeke:
The current doc shows `7.4` as the version where Electron started supporting the new debugger, but it was actually `1.7.4`. See https://electronjs.org/releases#1.7.4

image


gregvanl:
Thanks!

PR #1488, opened 2018-03-12
renamed username from maxibanki to mxschmitt

mxschmitt:
Hi,
I've renamed my account from maxibanki to mxschmitt, so it would be cool, when we could fix those dead URLs.
Best regards
Max


gregvanl:
You bet, happy to have you update your handle.

PR #1494, opened 2018-03-14
Update linux.md

szunaj13pl:
set permissions for key


Tyriar:
I don't really think this step is necessary and just adds another step in the script. People who want to do this are free to do so on their own.

Issue #1489, opened 2018-03-12
Document that debugging pytest tests with coverage does not work


brettcannon:
Basically if you have pytest-cov installed you will need pass `--no-cov` in order for things to work (see Microsoft/vscode-python#693 for details). We might want to have an aside mention of this fact due to pytest's popularity and the fact PyCharm documents the same issue.

Issue #1483, opened 2018-03-08
Document how .env files work with the Python extension


brettcannon:
Microsoft/vscode-python#544

The `.env` file support isn't documented extensively. E.g. what's the file format? Is it shell-specific or somehow agnostic (answer: specific)? And mentioning how it is commonly used to set `PYTHONPATH` to work around having your Python code in a subdirectory of your workspace would probably help people out who ask this regularly.

PR #1495, opened 2018-03-14
Document .env files and pytest-cov caveat

kraigb:
https://github.com/Microsoft/vscode-docs/issues/1483
https://github.com/Microsoft/vscode-docs/issues/1489


kraigb:
@brettcannon @DonJayamanne Can you review, especially the section added on .env files. I tried to come up with a usage scenario, but perhaps there's something better in your experience.

PR #1506, opened 2018-03-18
Add details on checking TS version

user135711:
Simply following the guide from the beginning, this command is not available at the moment that it's suggested, which will be confusing and likely result in wasted time trying to figure out why to people setting up their IDE.


gregvanl:
Thanks! Good clarification.

PR #1510, opened 2018-03-20
Corrected the typo in 'cwd'

Vest:
The previous word was "directly", but the meaning is "directory".

PR #1511, opened 2018-03-20
Wording change

kraigb:
#1510

### DIFF
--- a/docs/python/debugging.md
+++ b/docs/python/debugging.md
@@ -95,7 +95,7 @@ Specifies how program output is displayed.
 
 ### `cwd`
 
-Specifies the current working directory for the program being debugged, defaulting to the current folder. The standard configuration uses `${workspaceFolder}` to use the project folder.
+Specifies the current working directory for the program being debugged. The standard configuration uses `${workspaceFolder}` to use the project folder. If is not specified, `cwd` defaults to the current folder open in VS Code.
 
 ### `debugOptions`
 


docs/python/debugging.md
```diff
@@ -95,7 +95,7 @@ Specifies how program output is displayed.
 
 ### `cwd`
 
-Specifies the current working directory for the program being debugged, defaulting to the current folder. The standard configuration uses `${workspaceFolder}` to use the project folder.
+Specifies the current working directory for the program being debugged. The standard configuration uses `${workspaceFolder}` to use the project folder. If is not specified, `cwd` defaults to the current folder open in VS Code.
```
gregvanl: Would "If not specified, .." read better? Or "If it is not specified, ..."

kraigb: Doh. Hate it when that happens.

Vest: Does "_current folder_" mean the folder that is opened via **File -> Open Folder...** ? Or this could be the folder-child of the project, where the py file is?

Thank you, if you can clarify this question as well.

kraigb: It's the folder containing the Python file that you started in the debugger. I'll clarify that in the next set of changes, along with an example. Thanks for bringing that up, and I appreciate the push for clarity. :)

kraigb: Correction, the default for cwd if omitted is always ${workspaceFolder}, which is what we had there originally. The next set of updates clarifies this and adds examples (see PR #1532).

PR #1518, opened 2018-03-22
** swapped for * in search.

user135711:
The directory wildcard was used for the character wildcard.


user135711:
The original was correct, My bad:<

PR #1525, opened 2018-03-23
Fix the Windows settings path

dtzar:
The current path listed is not correct - at least on my Windows 10 machine.


gregvanl:
Hi @dtzar Thank you for your interest in the VS Code documentation. The Windows environment variable APPDATA should map to C:\users\{username}\AppData\Roaming. You can check this by running:

```
set APPDATA
```

from the Windows Command Prompt.

dtzar:
Doh!  Would you be open to a PR to add a minor comment to the docs to explain how to retrieve this value (as you have listed here)?

PR #1533, opened 2018-03-29
Add static website deploy tutorial

bowdenk7:
Still need to add a section that includes Travis, but I think this is good enough to start reviewing.

[Hosted here](http://staticsite-tutorial.azurewebsites.net/tutorials/static-website/getting-started ) for easier reviewing.


gregvanl:
Added 'don't merge' label while this is under review.

chrisdias:
Can't you just remove that file from the PR? it probably got built while running the site...

gregvanl:
The API reference topic is auto-generated and it isn't a big deal if we merge your changes in since we can do a baby build afterwards to regenerate it before we publish live.

bowdenk7:
@gregvanl, @fiveisprime, @kennethkau, @DanielRosenwasser - anyone get a chance to review this? Would love to get these published.

gregvanl:
@bowdenk7 Let me know when the technical content has been reviewed by others and then I can do an edit/conventions pass.

PR #1535, opened 2018-03-30
Clarify that shellArgs can also be used on Windows

thernstig:
terminal.integrated.shellArgs.* only specifically referred to
Linux and macOS, when it in fact also exists for Windows


gregvanl:
Thanks!

PR #1537, opened 2018-03-31
Fix typo: replace squigglies with squiggles

gregoryfm:


gregvanl:
Thanks. We've been use the diminutive squigglies but you are correct, it isn't really a word.

PR #1545, opened 2018-04-05
Tasks example problemMatcher regexp fix

budavariam:
The handcrafted tsc problemMatcher example does not match errors with **location**s that consists of **2 numbers**.

Example:  `src/subfolder/Example.ts(52,42): error TS2300: Duplicate identifier 'DuplicatedIdentifier'.`


gregvanl:
Thanks

PR #1553, opened 2018-04-07
Update javascript.md

itai-stratoscale:
Isn't Object Rest/Spread already supported?


gregvanl:
Thanks for pointing this out. I'll follow up since we can remove the entire paragraph.

PR #1556, opened 2018-04-09
change markdown syntax from json to javascript

mehagar:
The comment was showing up as highlighted red before because comments are not allowed in JSON.


gregvanl:
Hi @mehagar, the code block is json so I don't want to label it as javascript just to silence the errors in the GitHub rendering. The code block renders better on code.visualstudio.com and in VS Code itself when correctly marked as json.

PR #1558, opened 2018-04-09
Docs: change Launch.json console attributes

JPeer264:
Seems like `internalConsole` is not available as option, as shown in the attached image.

screen shot 2018-04-09 at 18 51 17


gregvanl:
Hi @JPeer264 Thank you for your PR. What VS Code build and debugger as you using? "internalconsole" is still a valid option for "console" with the Node.js debugger.

image

Different debuggers support different options.

JPeer264:
Ah alright got it. I was using it with python.

PR #1560, opened 2018-04-12
Update the tutorials to match on recent release.

fiveisprime:
We released some updates to the extensions that change some behavior in
the creation steps. This cleans that up so everything matches.


gregvanl:
@fiveisprime This will go out around lunch time with our 1.22.2 update.

gregvanl:
@fiveisprime Published.

PR #1562, opened 2018-04-14
[reactjs] Update debugger configuration with new defaults

patridge:
It appears the default `launch.json` debugger configuration for the **Chrome** environment no longer creates the `attach` configuration. If this is indeed the case, this will update the React example to reflect that and avoid any confusion. (If not, I need to figure out what's wrong with my vscode setup.)

If later tutorials use the `attach` configuration, I could put in the missing steps to add it afterwards, but I wasn't seeing it used anywhere in this tutorial.

Additionally, I clarified the next port setting to make sure folks are looking in the `url` setting rather than a discrete `port` value. This is less necessary without the `attach` configuration, which has a dedicated `port` setting, but still points people int he right direction.

The last change fixes the filename casing on an image. This doesn't affect the live site, but [GitHub's preview](ESLint will now analyze open files and shows) appears to be case-sensitive (search for `ESLint will now analyze open files and shows` to see the broken image in that link).


patridge:
Since I'm still working through this tutorial, I just added some typo fixes to add the missing `.` at the beginning of mentions of `.eslintrc.json`.

gregvanl:
Thanks!

PR #1567, opened 2018-04-18
Change `-p` to the more explicit `--load`

PeterJCLaw:
Using the clearer version of this option helps readers understand what they're being asked to actually do when running commands from the internet.

Note: I'm admittedly not sure whether `--load` is available on all platforms and versions, though it seems likely that it is present.


Tyriar:
> Note: I'm admittedly not sure whether --load is available on all platforms and versions, though it seems likely that it is present.

I'm a little hesitant to change this for this exactly reason, changing this could cause headaches for people on some distros. If people want to find out more they can always do `sysctl --help` so going to close this off.

PR #1580, opened 2018-04-25
Capture the case sensitive nature of search

sri-prasanna:
The search for folder and file names is case sensitive by default. I have added a question to capture this.


gregvanl:
Thanks!

gregvanl:
Hmm, playing around with global search, I don't think you are correct. Search doesn't appear to be case sensitive unless the Match Case toggle is set.

image

Match case on:

image

I'll take out your change until I understand why you were seeing case sensitive by default.

sri-prasanna:
Hi @gregvanl. The file names (to include/exclude) are case sensitive. I meant that.  I will update the pull request now.

PR #1585, opened 2018-04-30
Fix error in `docs/supporting/faq.md`

yeskunall:
As of VSCode v1.22.2, "Preferences" is under "Code" and not "File" menu item.

As seen on MacOS. I can try and confirm the same on Windows, if needed.

### DIFF
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -103,7 +103,7 @@ VS Code collects usage data and sends it to Microsoft to help improve our produc
 
 If you don’t wish to send usage data to Microsoft, you can set the `telemetry.enableTelemetry` setting to `false`.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @yeskunall, Thank you for your PR and interest in the VS Code documentation. File > Preferences is correct for Windows and Linux, only macOS has Preferences under Code. I'll make sure this is mentioned in the FAQ, we have usually put the macOS location in paratheses. Thanks for pointing out this confusion.

gregvanl:
@yeskunall I've updated the FAQ to mention the macOS menu with https://github.com/Microsoft/vscode-docs/commit/51fcbd9e072b33ec4ce7ca857ebdaf6387a63529. Thanks for pointing out the confusion for macOS users.

PR #1600, opened 2018-05-04
 docs(manifest): update list of allowed categories

kumarharsh:
update list of allowed categories

resolves mistake in earlier PR #1559


gregvanl:
Hi @kumarharsh I appreciate your refreshing the document but we'd like to keep the order as is since it put s the most common categories for community extenders first. We also don't want to highlight Azure since that is also not a category we expect community extenders to use.

kumarharsh:
OK. got it.

PR #1601, opened 2018-05-05
Clarified some tsconfig.json relations

phistuck:
- Added a link to the tsconfig.json reference.
- Added the expected string syntax to "include" and "exclude".

A follow up to https://github.com/Microsoft/vscode-docs/pull/1595#issuecomment-386771168.


gregvanl:
Thanks! Just the right level of detail.

PR #1605, opened 2018-05-06
Removed two exclaimation marks to solve inconsistency in a React code snippet.

bitsol:
Fixed inconsistency with number of exclamation marks in the React code snippet.
The image shows one, and the code example had three.


gregvanl:
Thanks, three is too loud anyway.

PR #1608, opened 2018-05-07
New default of "console" for Python launch options is "integratedTerminal"

fireattack:
See https://github.com/Microsoft/vscode-python/issues/1320#issuecomment-379162146


gregvanl:
Thanks!

PR #1609, opened 2018-05-08
Update CPP docs

bobbrow:


docs/languages/cpp.md
```diff
@@ -33,21 +31,19 @@ If you run into any issues or have suggestions for the Microsoft C/C++ extension
...
+**Note**: The C/C++ extension does not include a C++ compiler or debugger. You will need to install these tools or use those already installed on your computer. Popular C++ compilers are [mingw-w64](http://www.mingw-w64.org/) for Windows, Clang for [XCode](https://developer.apple.com/xcode/) for macOS, and [GCC](https://gcc.gnu.org/) on Linux. Also make sure your compiler executable is on your platform path for VS Code to find.
 
-### IntelliSense
+### Configuring IntelliSense
 
-**To enable code completion and navigation, you will need to generate a `c_cpp_properties.json` file:**
+The extension will attempt to determine your folder's basic configuration info based on compilers it finds on your system. If for any reason, that configuration is incomplete, you can generate a `c_cpp_properties.json` file by running the **C/Cpp: Edit configurations...** command from the **Command Palette** (`kb(workbench.action.showCommands)` and add the missing information.
```
ronglums: ...it finds on your system, including macOS, Linux, Windows (MSVC), Windows subsystem for Linux (WSL), MinGW. and Cygwin.

docs/languages/cpp.md
```diff
@@ -33,21 +31,19 @@ If you run into any issues or have suggestions for the Microsoft C/C++ extension
...
 
-* Find any green squiggle in a source file (for example, an #include statement) and set the cursor on the line.
-* Click the lightbulb that appears in the left gutter.
-* Click **Update "browse.path" setting**.
+If a `#include` file or one of its dependencies cannot be found, you can also click on the squiggle to view suggestions for how to update your configuration.
```
ronglums: ...green squiggles under include statements to view....

docs/languages/cpp.md
```diff
@@ -59,20 +55,18 @@ Below you can see that the MinGW C++ include path has been added to `browse.path
         "_DEBUG",
         "UNICODE"
     ],
+    "compilerPath": "C:\\mingw-w64\\bin\\gcc.exe",
     "intelliSenseMode": "msvc-x64",
```
ronglums: I thought it would be "clang-x64" mode for MinGW?

docs/languages/cpp.md
```diff
@@ -59,20 +55,18 @@ Below you can see that the MinGW C++ include path has been added to `browse.path
         "_DEBUG",
         "UNICODE"
     ],
+    "compilerPath": "C:\\mingw-w64\\bin\\gcc.exe",
```
ronglums: I thought MinGW would use "clang-x64" mode?

docs/languages/cpp.md
```diff
@@ -182,10 +176,10 @@ The C/C++ extension for Visual Studio Code supports source code formatting using
 
 You can format an entire file with **Format Document** (`kb(editor.action.formatDocument)`) or just the current selection with **Format Selection** (`kb(editor.action.formatSelection)`) in right-click context menu. You can also configure auto-formatting with the following [settings](/docs/getstarted/settings.md):
```
ronglums: This section seems to be talking about code formatting, but the title says building?

docs/languages/cpp.md
bobbrow: I don't think this additional verbosity adds to the topic.

docs/languages/cpp.md
bobbrow: I'm not seeing what you're seeing.  This is directly under the "Editing Code" -> "Code Formatting" section.

bobbrow: Oh, I see it now.  GitHub's diff view cuts out a lots of unchanged lines.  It's really confusing, but you can see the jump in line numbers in the left column.

docs/languages/cpp.md
ronglums: I'd like to call out the platforms supported, otherwise users would have never thought of using VS Code for WSL, for example. And it makes clear which ones we don't support :)

PR #1611, opened 2018-05-09
Document icon when contributing a new command

TomasHubelbauer:
I am basing this on:

https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/services/actions/electron-browser/menusExtensionPoint.ts#L245

### DIFF
--- a/docs/extensionAPI/extension-points.md
+++ b/docs/extensionAPI/extension-points.md
@@ -80,7 +80,11 @@ The following example contributes default editor configurations for the `markdow
 
 ## contributes.commands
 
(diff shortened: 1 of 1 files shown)


gregvanl:
@TomasHubelbauer Thank you for your interest in the VS Code documentation and you have pointed out a documentation hole. I'm not going to merge this PR since it may lead developers to think that simple Command Palette commands can support icons, which they don't. Adding icons to commands is available for only certain menus and needs to be documented in https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributesmenus.

PR #1615, opened 2018-05-13
Fixed numbering of Prerequisites.

DarqueWarrior:
The live site shows 1, 2, 1 instead of 1, 2, 3.
I also cleared two linting errors.


kraigb:
Thanks for picking those up!

PR #1620, opened 2018-05-17
add the commad for increase the amount of inotify watchers

inyee786:
Add command to increase the inotify watchers inside ```/proc/sys/fs/inotify/max_user_watches```


Tyriar:
That command is only temporary according to https://github.com/kubernetes/kubernetes/issues/46230#issue-230451350, so it would need to be run every time you restart.

PR #1622, opened 2018-05-19
Create custom snippets

AlisaLiso:
On MacOS this opening from Code > Preferences > User Snippets


gregvanl:
Hi @AlisaLiso File > Preferences > User Snippets is correct for Windows and Linux. 
What we often do in the VS Code documentation is add (Code > Preferences > * for macOS). See for example https://code.visualstudio.com/docs/getstarted/themes#_selecting-the-color-theme. 
Would you like to update your PR to add that?

gregvanl:
Thanks!

PR #1628, opened 2018-05-23
Updated Welcome to app image link

dkthedon:
Name of image file was not correct in link. image name has capital 'W'.


gregvanl:
Thanks!

PR #1640, opened 2018-05-28
Python linting lintOnSave 

umyuu:
The default values ​​of python.linting.lintOnSave do not match in the following two documents.

docs/python/settings-reference.md#linting-settings
docs/python/linting.md


gregvanl:
Thanks!

PR #1648, opened 2018-05-30
document relativePath predefined variable

bitjson:
This adds documentation for the relativePath feature: https://github.com/Microsoft/vscode/pull/50758


bitjson:
@gregvanl ah, didn't realize – thanks! I'll do that now.

PR #1649, opened 2018-05-30
document relativePath predefined variable

bitjson:
This adds documentation for the relativePath feature: https://github.com/Microsoft/vscode/pull/50758

(Replaces #1648)


gregvanl:
Added "don't merge" label as dependent on vscode PR which hasn't been accepted yet (not for 1.24)

gregvanl:
Closing since the fix is now proposed as an extension rather than a change to the VS Code product.

PR #1656, opened 2018-06-04
Python May updates

kraigb:
To be published after the extension update happens.


kraigb:
@gregvanl I just merged this PR with the updates for the most recent Python extension, which was just released. Please publish to live when you can.

gregvanl:
@kraigb Published

PR #1659, opened 2018-06-05
Fix a relative link

brettcannon:


kraigb:
Thanks for fixing these little details. They can slip through. :)

PR #1667, opened 2018-06-07
Update link to "Create a Docker group"

5hay:
Updated the link to point to the right file in the Docker Docs.

The current link doesn't point to the right file anymore.


gregvanl:
Thanks!

PR #1670, opened 2018-06-07
Fix flake8 config file docs

jamy015:
- Fix user config file path for Windows
- Remove sentence stating that only the first file is considered (I tried using two files simultaneously on my local install, and flake8 picked up configuration from both)
- Add `.flake8` file to allowable project config files (fixes https://github.com/Microsoft/vscode-python/issues/1884)


kraigb:
Thanks for the contribution. I appreciate it.

PR #1683, opened 2018-06-17
Add a section for pydocstyle to Python linting

pzelnip:
Adds a section detailing pydocstyle to the Python Linting documentation, to mirror what's in place for Pylint, mypy, etc


kraigb:
Great contribution! Thanks for adding those details (I did a small edit pass).

PR #1685, opened 2018-06-19
Update docs on cmder

amorijs:
The comment in the vscode.bat snippet causes the batch file to not work. This can cause confusion for beginners who may think it is okay to leave it in there, and may not understand what they are doing wrong. My update takes the comment out of the batch file.


gregvanl:
Thanks! Your change should help avoid confusion.

PR #1687, opened 2018-06-20
Fixes #1686 add black to list of possible formatters

### LINKED ISSUE #1686: Black not indicated as possible option for python.formatting.provider
On https://code.visualstudio.com/docs/python/editing#_general-formatting-settings in the table it indicates that the only options for ```python.formatting.provider``` are "autopep8" or "yapf", but "black" is now a supported option.

pzelnip:


gregvanl:
Merged per @brettcannon comment in https://github.com/Microsoft/vscode-docs/issues/1686

Issue #1680, opened 2018-06-14
Update Python code execution keyboard shortcut to be Shift+Enter


brettcannon:
Microsoft/vscode-python#1875

This will be in the next release of the extension.

/cc @Microsoft/pvsc-team

kraigb:
@brettcannon Did this change already ship?

DonJayamanne:
Yes it did. However it might change yet again

kraigb:
Fixed by https://github.com/Microsoft/vscode-docs/pull/1702. Please open another issue when there's a change.

PR #1690, opened 2018-06-25
Fix the hotkey to run selected code

uctptep:
Correct version is referenced in https://code.visualstudio.com/docs/python/editing#_run-selectionline-in-terminal-repl

### DIFF
--- a/docs/python/python-tutorial.md
+++ b/docs/python/python-tutorial.md
@@ -97,7 +97,7 @@ The command opens a terminal panel in which your Python interpreter is automatic
 
 There are two other ways you can run Python within VS Code:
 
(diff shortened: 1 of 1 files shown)


gregvanl:
The keyboard shortcut may also be changing shortly. https://github.com/Microsoft/vscode-docs/issues/1680

kraigb:
Yes, it's changing to Shift+Enter in the next rev of the extension, but you're right, that it's Ctrl+Enter for now (although it doesn't really work, see https://github.com/Microsoft/vscode-python/issues/1875. So I'll merge this for now and set it to Shift+Enter with the next update.

PR #1694, opened 2018-06-26
Update Go extension screenshot

matt-block:
Since the Go extension changed publisher from `lukehoban` to `Microsoft`, it could be confusing for users to see a publisher in the docs and a different one in the Extensions tab. Some might also think that the two are different extensions.

This PR updates the screenshot `go-extension.png` with an up-to-date one.


gregvanl:
Thanks!

PR #1701, opened 2018-07-01
Remove extra newline in formatting table

rupert:
The extra newline was causing the last two rows in the table to be rendered incorrectly.


gregvanl:
Thanks!

PR #1702, opened 2018-07-01
Update run in python terminal keyboard shortcut to shift+enter

rupert:
Updated to ⇧⏎ in this commit: https://github.com/Microsoft/vscode-python/commit/5eb48c276c9bffda5ab3b0f8335681b37187cacf


kraigb:
Thanks; it may be changing again, but Shift+Enter is there for now.

PR #1708, opened 2018-07-06
Update tutorial-flask.md

signifox:
fix __init__.py


kraigb:
Thanks. I thought I'd scrubbed all those mistaken instances, but I'm glad you caught it!

PR #1709, opened 2018-07-06
Updated tutorial-flask, minor changes for programs

brendonotto:
In the Go to Definition and Peed Definition Commands the directions referenced view.py which has not been created yet during the tutorial, updated to app.py instead.

The naming of __init.py__ is incorrect, changed to __init__.py


kraigb:
Thanks for the fixes! I appreciate you taking the time.

PR #1718, opened 2018-07-11
Merging existing preferences and extensions to portable

JochenHu:
Getting your preferences and extensions from an existing installation of code to portable folder.


gregvanl:
Hi @JochenHu, thank you for noticing a gap in the VS Code docs and your PR. We're going to be updating the portable topic shortly (see https://github.com/Microsoft/vscode/issues/53653). Once the topic is updated, have a look and let us know if you have feedback.

PR #1724, opened 2018-07-12
Updated the link for express application link

dkthedon:
images folder was one directory up but was referenced from the same directory.

PR #1725, opened 2018-07-12
Corrected the links for images.

dkthedon:
images folder was one directory but referenced from same directory.


gregvanl:
See explanation in https://github.com/Microsoft/vscode-docs/pull/1724

PR #1726, opened 2018-07-12
Updated the link for app service image.

dkthedon:
image folder is one directory up but was referenced from same directory.


gregvanl:
See explanation in https://github.com/Microsoft/vscode-docs/pull/1724

PR #1728, opened 2018-07-12
Fix minor error in theme-color-reference.md

kankaristo:
Custom title bar colors also work on Linux (not only on macOS), if `window.titleBarStyle` is set to "custom".

Tested on Fedora 28 with Gnome 3.28.2.


gregvanl:
Hi @kankaristo, Thanks for your interest in the VS Code documentation and taking the time to submit a PR. This feature is still in Preview https://vscode-westeu.azurewebsites.net/updates/v1_25#_custom-title-bar-and-menus-for-windowslinux for Linux and Windows and we'll update this topic once it is on by default.
P.S. I've been to Turku. I traveled there to play hockey in high school. Also Uusikaupunki. :-)

PR #1729, opened 2018-07-12
fix sub-word link in v1_25

mattfwood:
Corrected markdown anchor link for sub-word navigation section


gregvanl:
Thanks!

gregvanl:
Hi @mattfwood Unfortunately this change broke the in-product release notes navigation. It has trouble when there are hyphens in the header name. I'm going to change to link to the h2 Editor section. Thanks again for noticing the wrong link.

PR #1731, opened 2018-07-13
Variable transform example - change part 3 pointer

amitlevy21:
In the "Creating your own snippets" tutorial, the section for Variable transform shows a example. Each part that is required for making a transform is pointed, but i think that part 3 - which is pointed as "no options" should point to the forward slash and not $1 which belongs to part 2 of the transform syntax.


gregvanl:
Thanks! Great attention to detail.

amitlevy21:
Thank you!

PR #1732, opened 2018-07-15
Add stuff I didn't know about authoring (language) extensions

silvenon:
- The recommended icon size should be 256x256. We'll always want icons to be sharp on every screen, so I added this note in parenthesis.

- I added the documentation for `contributes.grammars.embeddedLanguages`. This field is very useful because VS Code seems to use it to inherit language _features_, not just syntax highlighting.

- It took me a long time to figure out that the grammar file doesn't have to be a Plist file. generator-code does a good job of making that clear, but I didn't use it at first because I didn't know how to answer all its questions. I added the `.json` extension to all `.tmLanguage` files in code examples and explained that using JSON is more convenient to use.


gregvanl:
Thanks!

PR #1733, opened 2018-07-16
Language change for example-language-server.md

Pyrobolser:
Hello,

English is not my native language and I had some trouble understanding this part.
* If the start and end positions are the same, VS Code will **_squiggle_** the word at that position.
* If you want to **_squiggle_** until the end of the line, then set the character of the end position to Number.MAX_VALUE.

I know it might seems unnecessary now that I understand. But I think putting some emphasis on the fact that VS Code is just underlying the word and not really _"doing"_ anything with it can make the documentation more easily understandable.

Let me know what you think.
Thanks a lot for reading,

### DIFF
--- a/docs/extensions/example-language-server.md
+++ b/docs/extensions/example-language-server.md
@@ -506,8 +506,8 @@ documents.onDidChangeContent(async (change) => {
 
 ### Diagnostics Tips and Tricks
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks! A nice clarification.

PR #1748, opened 2018-07-25
Add mentions of logpoints to Python debugging articles

kraigb:


kraigb:
@gregvanl I requested your review because I made a small edit to the main debugging article.

PR #1755, opened 2018-07-31
Fix typo in Python unit test config example

sedders123:
Change the Nose configuration example to enable Nose as opposed to pyTest


kraigb:
Doh! Copy-paste error in changes I just made...thanks @sedders123 for fixing it!

PR #1761, opened 2018-08-02
fixing of broken video link

johnlow95:


gregvanl:
Hi @johnlow95 Thank you for your interest in the VS Code docs. The href links without the trailing '.md' file extension are correct for the live site https://code.visualstudio.com/docs/getstarted/introvideos. It is confusing that they don't work when viewing the topic on GitHub.

PR #1768, opened 2018-08-06
Improved installation instructions for Debian/Ubuntu

AVeenstra:


Tyriar:
I got advice from a maintainer of apt that apt-key is deprecated and the current way is the best way to do it for now.

AVeenstra:
Sorry, it seems I skipped over that line in the man page.

Could I still suggest an alternative?
```bash
sudo curl -o /etc/apt/trusted.gpg.d/microsoft.asc  https://packages.microsoft.com/keys/microsoft.asc
```
Asc files seem to be supported directly according to the man page:
```
...
           Note: Instead of using this command a keyring should be placed directly in the /etc/apt/trusted.gpg.d/ directory with a descriptive name and
           either "gpg" or "asc" as file extension.
...
```

Tyriar:
I'd prefer not to touch this as it works right now, touching it could potentially break it under some distros.

PR #1781, opened 2018-08-10
Remove escape from manual path example

fprimex:
The "manually add VS code to your path" code is not copy-pasteable because there is an escape (`\`) before the `$`.


fprimex:
Hi. In this context you do want `$PATH` to be expanded, otherwise the only path that will end up being valid in the user’s environment is the one to the `code` command. The user would lose `ls`, `find`, `which`, `awk`, etc, etc.

fprimex:
Hi, I reread the page and forgot the code was in the middle of a `cat`. I copy / pasted because I did not want it at the end of my `.bash_profile`. Closing this!

PR #1785, opened 2018-08-13
Update linux.md

Kurgol:
* Replaced manual file move with ```apt-key add``` command in the Ubuntu and Debian instructions.
* Installed ```apt-transport-https``` before ```apt-get update``` as it's required.


Tyriar:
@Kurgol `apt-key` was intentionally left out, see https://github.com/Microsoft/vscode-docs/pull/1768

The `apt-transport-https` change is a good idea though, can you include only that in a new commit?

Kurgol:
yep no problem.

PR #1787, opened 2018-08-13
Updates to extensions re 1.26

### LINKED ISSUE #1786: Improvements to the Extensions page
The below are a few suggestions to improve https://github.com/Microsoft/vscode-docs/blob/298332ba2e8ad312f673389f462ffb800d403d66/docs/editor/extension-gallery.md

- The extensions viewlet is said to show the most popular extensions. But for users with 1 or more installed extensions, the viewlet shows Enabled, Recommended and Disabled extensions
- `you'll see an Reload button which will prompt you to restart VS Code`. This implies that clicking on the button will result in a prompt which is not the case.
- **Extension Details** Here we refer to it as `extension details window` but previously we say `You can click on the extension item to display the extension's VS Code Marketplace page where you can learn more.` The previous reference should use the term `Extension Details` page/window
- **Extension Pack**: This is changing with the current version of 1.26
- **Extensions view commands**: "More" is now "More Actions..."
- **Search for an extension**: Note that if you are using the id of the extension, you will have to use `@id:`. People often copy paste the extension id from other docs and not see any results
- **Search for an extension**: A link to filters from here can be helpful
- **List installed extensions**: `By default, the Extensions view will show your installed extensions`. Nope, we show Enabled by default
- **Uninstall an extension**: The image needs an update. It should show the Enabled extensions not Installed
- **Disable an extension**: How a about combining Uninstall and Disable? The same image can be re-used between the 2? 
- **Extension auto-update**: Needs an update to include the new setting `extensions.autoCheckUpdate`
- **Update an extension**: Rename this to `Update an extension manually` and include the point that if you have disabled both `autoUpdate` and `autoCheckUpdate`, then you can run `Check for Updates` to first see which extensions need update, and then update them
- **Command line extension management**: `show-versions` and `extensions-dir` is missing. Instead of the code sample in the example, why not show an image of the Extensions Management section of `code -h`? This will include docs for all the switches.
- **Extensions view filters**: Needs update as per 1.26

JacksonKearl:
Closes #1786


docs/editor/extension-gallery.md
```diff
@@ -23,31 +23,32 @@ This will show you a list of the most popular VS Code extensions on the [VS Code
...
+On the extension details page, you can read the extension's README as well as review the extension's:
 
 * **Contributions** - The extension's additions to VS Code such as settings, commands and keyboard shortcuts, language grammars, debugger, etc.
 * **Changelog** - The extension repository CHANGELOG if available.
 * **Dependencies** - Lists if the extension depends on any other extensions.
 
 ![extension contributions](images/extension-gallery/extension-contributions.png)
 
-If an extension is an Extension Pack, the **Dependencies** section will display which extensions will be installed. [Extension Packs](/docs/extensionAPI/extension-manifest.md#extension-packs) bundle separate extensions together so they can be easily installed at one time.
+[Extension Packs](/docs/extensionAPI/extension-manifest.md#extension-packs) bundle separate extensions together so they can be easily installed at one time. If an extension is an Extension Pack, the **Dependencies** and **Extension Pack** sections will display which extensions will be installed when you install the pack.
```
JacksonKearl: @gregvanl here I'm not sure of a good way to say that using `Dependencies` to specify a pack should be considered deprecated, as it causes unwanted behavior when disabling a component extension. `Extension Pack` is a new section this release so everything out there is going to be using `Dependencies` for some time now.

docs/editor/extension-gallery.md
```diff
@@ -23,31 +23,32 @@ This will show you a list of the most popular VS Code extensions on the [VS Code
 
 ![popular extensions](images/extension-gallery/extensions-popular.png)
 
-Each extension in the list includes a brief description, the publisher, the download count and a five star rating. You can click on the extension item to display the extension's [VS Code Marketplace](https://marketplace.visualstudio.com/VSCode) page where you can learn more.
+Each extension in the list includes a brief description, the publisher, the download count and a five star rating. You can click on the extension item to display the extension's details page where you can learn more.
```
JacksonKearl: not sure if we want to be sure to say that this comes from the market place page here.

docs/editor/extension-gallery.md
```diff
@@ -129,27 +132,37 @@ To make it easier to automate and configure VS Code, it is possible to list, ins
 Example:
 
 ```
+code --extensions-dir <dir>
```
JacksonKearl: @ramya-rao-a here I didn't want to add just a picture because all of the information would be in the picture, rather than the picture just supporting the text, so we'd have to have the same exact content again in the alt-text

docs/editor/extension-gallery.md
gregvanl: I think this is fine. We provide plenty of links to the Marketplace website so users will likely notice that the content is the same.

docs/editor/extension-gallery.md
gregvanl: I didn't think the Dependencies list is deprecated. Some extensions really do have dependencies on other extensions (for example most Azure extensions depend on the Azure Account extension). Might be good to describe the difference here.

ramya-rao-a: Let's not touch this section for now. Once 1.26 is out in the wild for a month and all extension packs have updated to the new model, we can come back and update this.

docs/editor/extension-gallery.md
```diff
@@ -71,13 +72,15 @@ For example, typing 'python' will bring up a list of Python language extensions:
...
+
 ## Manage extensions
 
 VS Code makes it very easy to manage your extensions. You can install, disable, update, and uninstall extensions through the Extensions view, the **Command Palette** (commands have the **Extensions:** prefix) or command line switches.
 
 ### List installed extensions
 
-By default, the Extensions view will show your installed extensions. You can also use the **Show Installed Extensions** command, available in the **Command Palette** (`kb(workbench.action.showCommands)`) or the **More** (`...`) dropdown, to clear any text in the search box and show the list of installed extensions.
+By default, the Extensions view will show the extensions you currently have enabled. You can also use the **Show Installed Extensions** command, available in the **Command Palette** (`kb(workbench.action.showCommands)`) or the **More** (`...`) dropdown, to clear any text in the search box and show the list of installed extensions, including those that have been disabled.
```
ramya-rao-a: lets talk about how we show the split of enabled, recommended and disabled. And then follow it up with something like 'if you want to see a complete list of your installed extensions..`

docs/editor/extension-gallery.md
```diff
@@ -87,27 +90,27 @@ To uninstall an extension, click the gear button at the right of an extension en
 
 ### Disable an extension
 
-Disable an extension by clicking the gear button at the right of an extension entry. You can disable an extension globally or just for your current Workspace. You will be prompted to reload VS Code after you disable an extension.
+If you don't want to permanently remove an extension, you can instead temporarily disable an extension by clicking the gear button at the right of an extension entry. You can disable an extension globally or just for your current Workspace. You will be prompted to reload VS Code after you disable an extension.
```
ramya-rao-a: temporarily disable **the** ?

docs/editor/extension-gallery.md
```diff
@@ -87,27 +90,27 @@ To uninstall an extension, click the gear button at the right of an extension en
...
 
-### Update an extension
+### Update an extension manually
 
-If you have extensions auto-update disabled, you can quickly look for extension updates by using the **Show Outdated Extensions** command which uses the `@outdated` filter.  This will display any available updates for your currently installed extensions. Click the **Update** button for the outdated extension and the update will be installed and you'll be prompted to reload VS Code. You can also update all your outdated extensions at one time with the **Update All Extensions** command.
+If you have extensions auto-update disabled, you can quickly look for extension updates by using the **Show Outdated Extensions** command which uses the `@outdated` filter. This will display any available updates for your currently installed extensions. Click the **Update** button for the outdated extension and the update will be installed and you'll be prompted to reload VS Code. You can also update all your outdated extensions at one time with the **Update All Extensions** command. If you also have automatic checking for updates disabled, you can use the `**Check for Updates** command to check which of your extensions can be updated.
```
ramya-rao-a: Missing ` in the end

docs/editor/extension-gallery.md
ramya-rao-a: agreed

PR #1791, opened 2018-08-14
Add apt-transport-https

Kurgol:
Installed ```apt-transport-https``` before ```apt-get update``` as it's a dependency for external repositories accessed via https.


gregvanl:
Thanks!

PR #1792, opened 2018-08-14
Update Docker definition in java-container

IAmHopp:
Docker is not a Linux toolkit and VS Code is not a Linux editor, so I don't see a reason to single out Linux there.


gregvanl:
Thanks! That was an odd description.

Issue #1794, opened 2018-08-14
Python debugging: Remote debugging no longer supports the 'secret' field


d3r3kk:
Our current Python debugging documentation states that the user should set the `"secret": "my_secret"` field in the `launch.json` of the remote debugging configuration. However, from the code of PTVSD here, you can see that we no longer support any use of `secret` at all.

The code snippet we ask the user to add to their Python scripts need to be updated in the documentation as well, as our current code sample looks like so:

```python
  ptvsd.enable_attach("my_secret", address = ('1.2.3.4', 3000))
```
...this should also be updated to show the current usage:

```python
  ptvsd.enable_attach(address = ('1.2.3.4', 3000), redirect_output=True)
```

Note that the documentation from the code file might be helpful here:

```python

# address : (str, int), optional
#        Specifies the interface and port on which the debugging server should
#        listen for TCP connections. It is in the same format as used for
#        regular sockets of the `socket.AF_INET` family, i.e. a tuple of
#        ``(hostname, port)``. On client side, the server is identified by the
#        Qualifier string in the usual ``'hostname:port'`` format, e.g.:
#        ``'myhost.cloudapp.net:5678'``. Default is ``('0.0.0.0', 5678)``.
#
# redirect_output : bool, optional
#        Specifies whether any output (on both `stdout` and `stderr`) produced
#        by this program should be sent to the debugger. Default is ``True``.
```

For our upcoming August release we should remove references to the `secret` field.

cc: @Microsoft/pvsc-team 

I know @brettcannon is prepping an update in time for our August release...

Issue #1793, opened 2018-08-14
Python extension: August release debugger has launch.json issues


d3r3kk:
[In the documentation for Python debugging](vscode-docs/docs/python/debugging.md) `launch.json` doesn't support the key names `localRoot`, `remoteRoot`, or `secret`, when `type`: `pythonExperimental`. When `pythonExperimental` goes away for the August release, these fields will actually still be supported.

The squiggles are expected, but there is a new format, we should probably update our docs from this:
```js
{
    "name": "Python Attach (Remote Debug 1.2.3.4)",
    "type": "python",
    "request": "attach",
    "localRoot": "${workspaceFolder}",
    "remoteRoot": "c:\\py\\hello",  // Set to the program location on the remote computer.
    //"remoteRoot": "~/hello",      // Linux example
    "port": 3000,                   // Set to the remote port
    "secret": "my_secret",          // Set to your specific secret
    "host": "1.2.3.4"               // Set to your remote host's public IP address
}
```
to this:
```js
{
    "name": "Python Attach (Remote Debug 1.2.3.4)",
    "type": "python",
    "request": "attach",
    "pathMappings": [
        {
            "localRoot": "${workspaceFolder}",
            "remoteRoot": "C:\\py\\hello",     // set to the program location on the remote computer
            // "remoteRoot": "/project/folder" // Linux example
        }
    ],
    "port": 3000,                   // Set to the remote port
    "secret": "my_secret",          // Set to your specific secret
    "host": "1.2.3.4"               // Set to your remote host's public IP address
}
```
It would also be prudent to state that the old path specifiers still work today, but this might change in the future and it would be best to update.

DonJayamanne:
That's odd, `secret~ too should have squiggles, as that has been removed in the new config!

kraigb:
Changes merged.

PR #1764, opened 2018-08-02
Python updates for August release

kraigb:
To be merged when August release is made

PR #1795, opened 2018-08-15
Updates for Python debugging

### LINKED ISSUE #1780: some python debugging instructions are not valid anymore
First Part:

As written here:
https://code.visualstudio.com/docs/python/environments#_choose-a-debugging-environment

"Although selecting a different environment changes the python.pythonPath value in your Workspace settings.json file, it doesn't affect the User settings.json file, which is used by default for debugging. You can also specify an interpreter for debugging using the pythonPath in a debugger configuration in the launch.json file. See Debugging."

is not valid anymore. The debugger uses the selected interpreter in the workspace settings also for debugging, even if in the user settings a different interpreter is selected.

I created a venv, in this installed the package matplotlib with pip3, then created a file test.py which inherits:
'
import matplotlib

print('test')
'
My workspace settings are:

'
{
    "python.pythonPath": "${workspaceFolder}/env/bin/python"
}
'
My user settings are:
'
{
    "python.pythonPath": "/usr/bin/python3",
    "python.autoComplete.addBrackets": true
}
'
And with "Run Python File in Terminal" and "Start Debugging" I get no errors, even if matplotlib is just installed in the venv and not in the global python3 environment.
So the statement in the documentation is not valid anymore.

Second part:

Also in https://code.visualstudio.com/docs/python/debugging
"
Note: The debugger settings don't support relative paths, including when relying on the main python.pythonPath setting. To work around this, use an environment variable, or create a variable such as ${workspaceFolder} that resolves to your project folder, then use that variable in the path, as in "python.pythonPath": "${workspaceFolder}/venv/bin/python".
"
is not valid anymore. Because the same setup like above also works with just relative paths without producing any error.

With this workspace settings:

'
{
    "python.pythonPath": "./env/bin/python"
}
'
I am on Debian Stretch with VSCode installed from the MS repository and used the instruction in the offical documentation.

brettcannon:
Fixes #1780 , #1793 , #1794 .

Not to be merged until #1764 lands which will be once the August release of the Python extension goes out.


docs/python/debugging.md
```diff
@@ -191,58 +189,66 @@ Remote debugging allows you to step through a program locally within VS Code whi
...
+1. Both computers: install [ptvsd](https://pypi.org/project/ptvsd/) using `python -m pip install --upgrade ptvsd` into your environment (while using a form of virtual environment is not required, it is strongly recommended).
 
 1. Remote computer: open the port you wish to use for debugging in the appropriate firewall or other networking configuration.
 
-1. Remote computer: in the source code, add the following lines, replacing *my_secret* with a passphrase that you'll use to authenticate remote debugging, and replacing *address* with the remote computer's IP address and port number (IP address 1.2.3.4 is shown here for illustration only).
+1. Remote computer: there are two ways to specify how to attach to the remote process. Due note that you may need to specify the remote computer's private IP address, if applicable (Linux VMs on Azure, for example, have both a public and private address). If you use the public IP address, you might see the error "Cannot assign requested address".
```
kraigb: Typo "Due" should be "Do".

docs/python/debugging.md
```diff
@@ -191,58 +189,66 @@ Remote debugging allows you to step through a program locally within VS Code whi
...
 
-    ```python
-    import ptvsd
+   1. in the source code, add the following lines, replacing *address* with the remote computer's IP address and port number (IP address 1.2.3.4 is shown here for illustration only).
```
kraigb: Typo change "in" at beginning to "In".

docs/python/debugging.md
```diff
@@ -191,58 +189,66 @@ Remote debugging allows you to step through a program locally within VS Code whi
...
+        ptvsd.enable_attach(address=('1.2.3.4', 3000), redirect_output=True)
+
+        # Pause the program until a remote debugger is attached
+        ptvsd.wait_for_attach()
+        ```
+
+        The IP address used in `enable_attach` should be the remote computer's private IP address. You can then launch the program normally, causing it to pause until the debugger attaches.
 
-    The IP address used in `enable_attach` should be the remote computer's private IP address, if applicable (Linux VMs on Azure, for example, have both a public and private address). If you use the public IP address, you might see the error "Cannot assign requested address".
+    1. Launch the remote process _through_ ptvsd, e.g.
```
kraigb: Change "e.g." to "for example:"

docs/python/debugging.md
```diff
@@ -191,58 +189,66 @@ Remote debugging allows you to step through a program locally within VS Code whi
...
 
-1. Local computer: in the source code, add a commented-out copy of the same code added on the remote computer. Adding these lines makes sure that the source code on both computers matches line by line.
+       ```
+       python3 -m ptvsd --host 1.2.3.4 --port 3000 -m myproject
+       ```
+
+       This will start the package `myproject` using `python3`, with the remote computer's private IP address of `1.2.3.4` and listening on port `3000` (you can also start the remote Python process specifying a file path instead of using `-m`).
```
kraigb: Change "This will start" to "This starts" (minor stylistic thing).

docs/python/debugging.md
```diff
@@ -191,58 +189,66 @@ Remote debugging allows you to step through a program locally within VS Code whi
...
+
+1. Local computer: **Only If you modified the source code on the remote computer as outlined above**, then in the source code, add a commented-out copy of the same code added on the remote computer. Adding these lines makes sure that the source code on both computers matches line by line.
```
kraigb: Typo, change "Only If" to "Only if" (no caps on if)

docs/python/environments.md
```diff
@@ -81,7 +81,13 @@ To avoid activating virtual and conda environments when using these terminal com
 
 ### Choose a debugging environment
 
-Although selecting a different environment changes the `python.pythonPath` value in your Workspace `settings.json` file, it doesn't affect the *User* `settings.json` file, which is used by default for debugging. You can also specify an interpreter for debugging using the `pythonPath` in a debugger configuration in the `launch.json` file. See [Debugging](/docs/python/debugging.md).
+Debugging will use the the value from `python.pythonPath`in selecting what Python interpreter to use in the following order of precedence:
```
kraigb: Change "will use" to "uses".

kraigb:
Thanks for addressing these bits while I was away!

PR #1809, opened 2018-08-23
Update launch.json path in python-tutorial.md

timmhayes:
Added the ${workspaceFolder} variable to provide python with the full path to the "hello.py" script - without which this code will not work.


kraigb:
Thanks! (Belatedly...I was away on vacation. :))

PR #1811, opened 2018-08-24
Fix grammar by using correct possessive pronoun

rarrarrarr:
s/it's/its


gregvanl:
Hi @rarrarrarr Thank you for your interest in the VS Code documentation and your PR. The settings text comes from the VS Code product itself (you can see the same grammar issue in the setttings editor descriptions). The grammar should be corrected in the VS Code product repo https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/electron-browser/main.contribution.ts. Feel free to submit a PR there.

PR #1812, opened 2018-08-25
explicitly show which package.json file is referred to

amadeann:
With the current wording, it's not obvious to see if `package.json` file refers to the one in root directory, or client directory.


gregvanl:
Thanks

PR #1814, opened 2018-08-26
Fix typos in extensions.autoCheckUpdates setting name

DeeUnderscore:
The setting's name is `extensions.autoCheckUpdates` ([vscode source](https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/parts/extensions/electron-browser/extensions.contribution.ts#L211)).


gregvanl:
Thanks!

PR #1815, opened 2018-08-27
Fix description of the debugging process

DaveSprague:
In the original text it refers to the debug "continue" command as the "run" command but those are different commands.  Once you start debugging, F5 is assigned to "continue" so I fixed the description  to change "run" to "continue" in that section and to remove the reference to pressing the green arrow (which generates an error message to the user saying that a debug session is already underway).


kraigb:
Thanks for the detailed corrections! Much appreciated.

PR #1817, opened 2018-08-28
Update introvideos.md with correct video durations

domdomegg:
All the other videos are done using the number of whole minutes of the video - i.e. rounding down to the next minute. (3:54 -> 3 minutes)

Alternatively the durations could be to the nearest minute, but it would involve changing more of them (3:54 -> 4 minutes). Either way, the duration for "Debugging" is way out as it is now.


gregvanl:
Thanks!

PR #1820, opened 2018-08-29
Update linux.md

bdejean:
Debian installation: fix microsoft.gpg permissions which has to be readable by _apt user.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -26,7 +26,8 @@ The repository and key can also be installed manually with the following script:
 
 ```bash
 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-sudo install -o root -g root -m 644 microsoft.gpg /etc/apt/trusted.gpg.d/
+sudo apt-key --keyring /etc/apt/trusted.gpg.d/microsoft.gpg add microsoft.gpg
+rm -f microsoft.gpg
 sudo sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/vscode stable main" > /etc/apt/sources.list.d/vscode.list'
 ```
 
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -26,8 +26,7 @@ The repository and key can also be installed manually with the following script:
 
 ```bash
 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-sudo apt-key --keyring /etc/apt/trusted.gpg.d/microsoft.gpg add microsoft.gpg
-rm -f microsoft.gpg
+sudo install -o root -g root -m 644 microsoft.gpg /etc/apt/trusted.gpg.d/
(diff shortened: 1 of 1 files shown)


Tyriar:
@bdejean I don't recall having issues with install on Debian, don't you just enter `su` and run as normal?

bdejean:
@Tyriar sorry, I thought I had included a trace:

```
W: http://debug.mirrors.debian.org/debian-debug/dists/experimental-debug/InRelease: The key(s) in the keyring
/etc/apt/trusted.gpg.d/microsoft.gpg are ignored as the file is not readable by user '_apt' executing apt-key.
```

I think the problem is that the key download is done as a user (which is ok) but the assumption is that the downloaded file as at least 444 for apt to work. But if the downloading user has narrow umask (like i do), this is no longer true true. Moreover, moving the file to `/etc` keep it owned by the installing user.
Doing the download as root would have the same issue if umask is restrictive.
So I think the right thing to do is to fix the rights and change the ownership to root.
This could also be achieved by using `install(1)`.

bdejean:
Or we can even simply `sudo apt-key --keyring /etc/apt/trusted.gpg.d/microsoft.gpg add microsoft.gpg`.
I'm a bit new to github interface for pull-request, i'm not sure if you have the option to squash all of this.

docs/setup/linux.md
```diff
@@ -26,7 +26,8 @@ The repository and key can also be installed manually with the following script:
 
 ```bash
 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
-sudo mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg
+sudo apt-key --keyring /etc/apt/trusted.gpg.d/microsoft.gpg add microsoft.gpg
```
Tyriar: We were advised by someone from the apt team that apt-key is deprecated and the way we're doing it is the best way right now

Tyriar:
Something I've wanted to avoid was making unnecessary changes to these instructions and also keeping them as small as possible. I'm confused as to why this is the first time I've heard of the issue you're describing after all this time of the package being out there - if this is some customization that you're doing to restrict the file system then I think it should be on you to know how to work around the error. This seems preferable to me than making everyone run more commands that some may not know what they do.

bdejean:
I'm not doing anything unusual, I'm only using a umask != 0022. And what I'm also saying is the newly installed key should be owned by root and world readable, just like the other keys. So using install seems the best solution.

bdejean:
Btw, there's also a permission problem with vscode.list which has too be world readable as well:

```
$ LANG=C apt-cache search foobar
lxmusic - LXDE music player
man2html - browse man pages in your web browser
node-camelcase - Convert a string to camelCase
node-strict-uri-encode - stricter URI encode adhering to RFC 3986
python-os-apply-config - Creates config files out of cloud metadata
W: Unable to read /etc/apt/sources.list.d/vscode.list - open (13: Permission denied)
```

PR #1831, opened 2018-09-05
Issue launching code when using a remote desktop on Ubuntu.

dynamicwebpaige:


Tyriar:
I'm a little hesitant to officially recommend this as while it may work it's a pretty dodgy workaround and likely will break other things, I don't know the extent to which it will break what though. The best course of action might be to leave the workaround in [the issue](https://github.com/Microsoft/vscode/issues/3451) which will probably be found when users search for it.

dynamicwebpaige:
Sounds good to me! ☺ Just wanted to make sure that the issue was surfaced, since many of our Linux end users will be connecting remotely to VMs. Feel free to close the pull request.

dynamicwebpaige:
This solution is one of the top 6 on StackOverflow - but only surfaces during searches for Ubuntu Server + XRDP. The fail is a bit annoying (you download VS Code; type `code` into an xterm; and the app just silently fails to launch). 😥

Tyriar:
After thinking about this some more I think it's probably safest for us not to recommend doing this as we don't know what the impact is on the user's system. Thanks for the PR though 😃

PR #1833, opened 2018-09-06
🐛 FIX: UI Editor Name

ahmadawais:
The new editor is not `Preview` anymore so replaced the instructions with `UI` instead.


gregvanl:
Thanks!

ahmadawais:
Always happy to help. 🙌

PR #1835, opened 2018-09-07
Update identifiers.md

jprichardson:
I was tearing my hair out trying to figure out why my custom snippets weren't applying in my `*.tsx` files. I realized the scope was set to `typescript` and the the list of language identifiers does not contain the correct `typescriptreact` identifier.


gregvanl:
Thanks!

PR #1836, opened 2018-09-08
Updating Testing extensions page

lannonbr:
I added a section to include how to disable other extensions in the testing instance of VS Code while developing your own extension.


Almenon:
@yozlet I reccomended this to lannonbr, thanks for the suggestion!

gregvanl:
Thanks! Great suggestion

PR #1839, opened 2018-09-10
Add note about using "markdownDescription"

roblourens:
Fix Microsoft/vscode#58353

@octref is this the right place for this?


octref:
Yep, looking good.

PR #1847, opened 2018-09-13
docs/setup/linux.md: Update (open)SUSE instructions

mook-as:
No need to manually edit configuration files; make zypper do it for us, so that it probes the repository and makes sure we don't make any typos.
Also turn on auto-refresh, and we won't need to refresh it manually.


Tyriar:
Previously `gpgkey=https://packages.microsoft.com/keys/microsoft.asc` was being added to `/etc/zypp/repos.d/vscode.repo`, is that endpoint included as part of `--gpgcheck` or is it no longer needed?

ismail:
> Previously `gpgkey=https://packages.microsoft.com/keys/microsoft.asc` was being added to `/etc/zypp/repos.d/vscode.repo`, is that endpoint included as part of `--gpgcheck` or is it no longer needed?

That's a good question, I need to check zypper docs to be sure, I'll update here next week.

mook-as:
(Sorry for the delay, I seem to have lost the GitHub notification mail)
As far as I can tell, `--gpgcheck` is the default configuration (but the PR has been updated to include it anyway).  Also, I don't think it actually takes an argument; it just looks at the RPM database for installed keys, as far as I know.  And I don't think it even read the `gpgkey=` line in the repo config (see [bug 1088037](https://bugzilla.opensuse.org/show_bug.cgi?id=1088037)).

ismail:
@mook-as This is fixed now: https://github.com/openSUSE/libzypp/pull/133

So please update to documentation to manually add the following line to the repo file:

`gpgkey=https://packages.microsoft.com/keys/microsoft.asc`

PR #1861, opened 2018-09-19
Mark the preview html command as deprecated

mjbvz:
Adds a note to the `previewHtml` command that it is now deprecated. Delete the supporting material around this command, most of which is now covered by the webview documentation


docs/extensionAPI/vscode-api-commands.md
```diff
@@ -11,13 +11,6 @@ MetaDescription: Visual Studio Code extensions (plug-ins) complex commands Refer
 
 This document lists the set of Visual Studio Code complex commands. They are called complex commands because they require parameters and often return a value. You can use the commands in conjunction with the `executeCommand` API.
 
-The following is a sample of how to preview a HTML document:
```
gregvanl: Is there another simple example of using the executeCommand with a parameter?

PR #1875, opened 2018-09-23
I've added a paragraph related with Node Version Manager.

dimitardanailov:
I'm using Nodejs since 2012 and I prefer to install Node Version Manager vs pure Nodejs version installation. 
It's much easier for support and adding a new version.


gregvanl:
Thanks for this PR. I'm going to merge your suggestion to use nvm into the previous paragraph.

PR #1876, opened 2018-09-23
Consistent spacing in typescript.md

truckingsim:
Fix only example that doesn't put a space before the closing curly brace.


gregvanl:
Thanks!

PR #1879, opened 2018-09-25
Update editingevolved.md

kevinvn1709:
Change typing : to @ in Go to Symbol section


gregvanl:
Hi @kevinvn1709 Thank you for submitting a pull request. However I don't think your suggested change is correct. If you run the Go to Symbol command (Ctrl + Shift + 0), you will see the @ character:

image

If you then press the ':' character, the symbols will be grouped by category as described:

image

I can see the confusion. Perhaps we need to move the ':' description below the image and add a second image showing the categories.

PR #1889, opened 2018-10-01
Change json code blocks language identifier to 'json5' to support comments

plibither8:
GitHub's code blocks support `json5` syntax highlighting that allows for comments to be written in json code. Since most of the json code blocks in the docs have comments, instead of them showing an ugly red highlight, they should be rendered as proper comments.

With language identifier set as `json`:
2018-10-01-16-59-github com

...and with language identifier set as `json5`:
2018-10-01-16-57-github com

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -153,7 +153,7 @@ If you are listing out multiple key bindings, you can use a table.
 
 For source code we use the fenced code block notation ```` ``` ````.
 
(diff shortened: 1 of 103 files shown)


gregvanl:
Hi @plibither8 Thank you for your interest in the VS Code documentation. While GitHub Markdown rendering supports jscon5, unfortunately the VS Code website and in-product Markdown preview do not yet support json5.

PR #1898, opened 2018-10-01
misc: Bring the default settings.json up-to-date with latest VSCode

KingDGrizzle:
This PR brings the default settings.json code block up to date with the latest VSCode (at the time of PRing, VSCode 1.27.2). I've ran VSCode with `--disable-extensions` to get only the VSCode settings.


gregvanl:
@KingDGrizzle Thank you for your pull request. Updating settings.md is a little tricky as we combine several platform specific settings, remove experimental settings and also filter out values which would make the sections very long. The VS Code team handles this update each release in the vnext branch.

PR #1899, opened 2018-10-01
Update information on Portable Mode for Windows

conradj:
Provide further info that Portable Mode can help in corporate environments or if you want to change the extension installation folder. 

Make it clear that this mode prevents auto updates.

### DIFF
--- a/docs/editor/portable.md
+++ b/docs/editor/portable.md
@@ -9,11 +9,13 @@ MetaDescription: Visual Studio Code supports a Portable Mode.
 ---
 # Portable Mode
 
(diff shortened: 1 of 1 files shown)


joaomoreno:
While it's true that portable mode on Windows does not yet support auto update, the correct order of statements is:

1. Windows ZIP doesn't support auto update.
2. Portable mode on Windows is only supported on the ZIP.

You can currently have portable mode and auto update on macOS.

gregvanl:
I'll merge this PR and then make the updates.

PR #1910, opened 2018-10-02
tips and tricks - select all occurrences

taranvohra:
VSCode Tip - how to select all occurrences of current selection to change them at once


gregvanl:
Thanks

PR #1919, opened 2018-10-03
Trivial fixes in Python topic

eallrich:
I think Django is pretty consistent about using the single word "superuser" and the use of the two words "super user" in the section title was the only use I found which didn't match.

The sub-list in the Flask page wasn't being formatted as a list (including [in the online VS Code docs](https://code.visualstudio.com/docs/python/tutorial-flask#_refactor-the-project-to-support-further-development)) so it was hard to read.


kraigb:
@eallrich Nice catches on those details--thanks for fixing them.

PR #1920, opened 2018-10-03
doc: added Textmate snippet syntax exception for `\u` (#1882)

### LINKED ISSUE #1882: Snippet syntax not totally following TextMate
According to the VSCode documentation, the snippet syntax follows the one of TextMate snippet syntax.
However some TextMate syntax like the \u command in transform format cannot be used. This has been reported in the [vs-code issue #59554](https://github.com/Microsoft/vscode/issues/59554).
I would recommend clarifying the documentation since it is confusing. The confusion comes from the usage of the word "follows" and then the mention of a single exception. The \u syntax is not part of the stated exception.

Thanks.

mcgrathg:
## Description
Added another case where snippet syntax does not follow the TextMate snippet syntax, that `\u` is not supported (resolves #1882)


gregvanl:
Thanks!

PR #1925, opened 2018-10-03
Added links to python formatters

Hoenn:
VSCode docs are wonderful for being link-rich, so I added some links to the different python formatting styles available in the official python extension


kraigb:
@Hoenn  Thanks for adding those! I think I'd intended to do that originally but didn't circle back.

Hoenn:
@kraigb sure thing!

PR #1928, opened 2018-10-04
Fixed image link issues in tutorials

muradkhan101:


muradkhan101:
The fix was to make the links based off the root instead of the relative links present currently. However I noticed the 'docs' section has an images folder for each folder in docs, which is different than the 'tutorials' folder. To bring the two inline with each other, we can make an images folder in each subfolder of 'tutorials'. I didn't do that however, because I wasn't sure if that would break anything else in the repo.

gregvanl:
@muradkhan101 Thank you for your interest in the VS Code documentations. These images render correctly on our website and we need to keep the image links as they are. I know it is confusing that they don't render correctly on GitHub.

drecali:
@gregvanl thanks for your explanation. It's a very odd issue. Are the relative links kept to save time and not have to update the image links in case the folder structure changes in the future?

gregvanl:
@drecali The link paths are an artifact of our website build process. In general, the vscode-docs repo is tuned to support the live website but we do make some effort to support GitHub rendering (such as adding .md for topic links) if possible.

PR #1929, opened 2018-10-04
Minor improvements/corrections to Python docs

pzelnip:
Makes a few minor grammatical corrections to some of the Python docs, adds links to the DBMS homepages in the Django tutorial.

### DIFF
--- a/docs/python/python-tutorial.md
+++ b/docs/python/python-tutorial.md
@@ -12,7 +12,7 @@ MetaSocialImage: images/tutorial/social.png
 
 In this tutorial you use Python 3 to create the simplest Python "Hello World" application in Visual Studio Code. By using the Python extension, you make VS Code into a great lightweight Python IDE (which you may find a productive alternative to PyCharm).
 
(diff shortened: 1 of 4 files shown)


docs/python/python-tutorial.md
```diff
@@ -12,7 +12,7 @@ MetaSocialImage: images/tutorial/social.png
 
 In this tutorial you use Python 3 to create the simplest Python "Hello World" application in Visual Studio Code. By using the Python extension, you make VS Code into a great lightweight Python IDE (which you may find a productive alternative to PyCharm).
 
-> **Note**: You can use VS Code with Python 2 with this tutorial, but you need to make appropriate changes to the code.
+> **Note**: You can use VS Code with Python 2 with this tutorial, but you will need to make appropriate changes to the code.
```
kraigb: Stylistically, we try to avoid using future tense like "you will", even though it's common in vernacular English. One reason is that it doesn't translate well to other languages. So in this case I'm reverting the edit.

PR #1934, opened 2018-10-04
Add ui guide for node deployment

dougajmcdonald:
This PR adds a sibling guide to the `nodejs-deployment` (called `nodejs-deployment-ui`) guide but uses the Azure web portal rather than the CLI.

Some important points:

- Opted not to rename the existing folder `nodejs-deployment` to 'nodejs-deployment-cli' as was my first instinct to avoid external links breaking
- Opted to duplicate pages and make minor changes rather than try to link back to the originals. Aiming for having a single reason to change.
- Have used the original images where possible

Happy to discuss any issues with the PR but wanted to help and not having the UI guide caught me out a few times.

Also corrected what I thought were a few typos through the blog posts.


gregvanl:
Hi @dougajmcdonald Thank you for your interest in the VS Code documentation. Unfortunately, your tutorial using the Azure portal doesn't highlight the use of VS Code or a VS Code Azure extension so does not fit into our doc set. A better home for an Azure portal tutorial might be with the Azure content on docs.microsoft.com https://docs.microsoft.com/en-us/azure.

dougajmcdonald:
Hi @gregvanl, thanks for the feedback. Afraid I'm a little confused by your comment though. The tutorial in the PR shows you how to use VS Code to publish changes after creating your initial app, and search extentions, as per the pure `cli` node deployment tutorial here: https://code.visualstudio.com/tutorials/nodejs-deployment/getting-started

I have amended my tutorial to also commit the files initially through VS Code so it's a bit more `VS Codey` in case that helps.

Overall, I was trying to write a tutorial to help those who cannot or do not want to use the Azure CLI deploy a site they've written in VSCode. 

I'm unable to re-open this PR, so have submitted another at #1954

gregvanl:
Hi @dougajmcdonald Thanks for submitting your PR again but we don't want to be documenting the Azure portal on the VS Code website. In addition, the Azure portal approach for website creating and deployment goes counter to recommended [DevOps ](https://itrevolution.com/book/accelerate/) practices where configuration should be repeatable and treated like source code.

Issue #1151, opened 2017-08-31
Weird extra details for enumeration values


bpasero:
I am not sure what the purpose of this is:

![image](https://user-images.githubusercontent.com/900690/29932505-c4e0c990-8e74-11e7-89b4-840199b82b43.png)

It looks like the ordinal value of the enumeration, but is that useful?

/cc @jrieken

jrieken:
Yeah, kind of upstream as the library add the 'value' of enums into them

gregvanl:
Would be fixed if we adopted the suggestion in #665. Close as duplicate

PR #1938, opened 2018-10-05
Changes enumerated values collapsed items to markdown table

### LINKED ISSUE #665: Use table instead of collapsed items while documenting enumerated values
https://code.visualstudio.com/Docs/extensionAPI/vscode-api#_a-namecompletionitemkindaspan-classcodeitem-id690completionitemkindspan talks about the different Completion Item Kinds.

When you expand any of the below, all you see is the number associated with the enumeration.
If its just the number we want to document, then we should use a table. Why all the clicks just to see a number?

image

victorcmoura:
In the file https://github.com/Microsoft/vscode-docs/blob/master/docs/extensionAPI/vscode-api.md there were lots of enumerated values listed with collapsed items.

According to the discussions in https://github.com/Microsoft/vscode-docs/issues/665 and https://github.com/Microsoft/vscode-docs/issues/1151, it would be better to display the values in a table.

This pull request changes all collapsed items to markdown tables.

closes #665

### DIFF
--- a/docs/extensionAPI/vscode-api.md
+++ b/docs/extensionAPI/vscode-api.md
@@ -2943,180 +2943,33 @@ line completions were <a href="#CompletionItemProvider.provideCompletionItems">r
 
 #### Enumeration members
 
-
-
-<a name="CompletionItemKind.Class"></a><span class="ts" id=827 data-target="#details-827" data-toggle="collapse"><span class="ident">Class</span></span>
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @victorcmoura Thank you for your interest in the VS Code documentation. The vscode-api.md file is auto-generated by our documentation build from the vscode.d.ts file and the API generation tooling is not open source.

PR #1942, opened 2018-10-06
Update create eslintrc command

koole:
The name of the command for creating an `.eslintrc` has changed. I've updated the text and the screenshot.


gregvanl:
Thanks!

PR #1944, opened 2018-10-06
Fix link to "manually specify an interpreter"

danielthepope:
I found a broken internal jump link, so I fixed it.


gregvanl:
Thanks!

PR #1950, opened 2018-10-08
User Interface/ Adds zen mode image

NanduKMG:
Adds zen mode preview to the document under gettingstarted/userinterface.


gregvanl:
Hi @NanduKMG Thank you for your interest in the VS Code documentation and submitting a PR. An image of Zen mode isn't really required to understand the text. We hope most users will just try it. :-)

PR #1954, opened 2018-10-09
Add ui guide for node deployment

dougajmcdonald:
This PR adds a tutorial to demonstrate how to deploy a website written in VSCode to azure using the azure web portal rather than the CLI in the existing docs (https://code.visualstudio.com/tutorials/nodejs-deployment/getting-started).

This was closed at #1934 but I didn't have permission to re-open so have re-submitted with some additional changes to use VS Code further than the original docs (such as during the initial commit)


gregvanl:
Hi @dougajmcdonald Thanks for submitting your PR again but we don't want to be documenting the Azure portal on the VS Code website. In addition, the Azure portal approach for website creating and deployment goes counter to recommended [DevOps](https://itrevolution.com/book/accelerate/) practices where configuration should be repeatable and treated like source code.

dougajmcdonald:
Hi @gregvanl thanks for the feedback.

In terms of justification, I feel the article is very similar to the other article relating to deployment to Azure which gave me the impression that the subject matter was appropriate.

The reason I feel it's important to document the UI is that not everyone is able to run the CLI, in our commercial environment for example, running CLI tools with internet access is prohibited. The CLI approach also requires installing tools, which again, in certain environments is not permitted, using a web based UI is a helpful tool in these environments. I have a number of devs in my team asking specifically for a walkthrough of the portal UI.

We have not found any MS documentation which focuses on a UI walkthrough which is why I felt it useful to add. 

I'm a bit disappointed that during something like hacktoberfest, a first time contribution is handled with such a "we don't want this in our repo as we've made a decision not to document UI in docs" type rejection, you should probably make this clear to people trying to improve the range of topics documentation covers.

As a final point, whilst I can understand not feeling there is such value in documenting other parts of the MS toolset, I am certainly of the opinion that it can't hurt. What specifically do you think the harm of publishing a guide based on the portal UI would be?

Would appreciate any feedback so I can understand the decision.

Cheers,

Doug

gregvanl:
Hi @dougajmcdonald, The VS Code doc set is pretty small (about 130 topics) with clear owners who maintain the topics as the product changes. One worry about including a topic about the Azure portal is that the VS Code team won't be aware of portal changes and UI changes are not usually considered "breaking changes" so are not widely announced. Any topic on the VS Code website needs to be maintained and we are not ready to commit to owning a new tutorial.
Have you considered publishing the tutorial yourself, perhaps on a site such as [medium.com](https://medium.com)?

PR #1962, opened 2018-10-10
 Setup: add Arch Linux installation instructions

DepthSec:
Added detailed instructions for vscode installation on Arch Linux with git & makepkg.


DepthSec:
@Tyriar Yes, maybe this would be better, but I thought I'd write it in more detail because most other distros have more detailed descriptions, including the installation of the vscode repo.

DepthSec:
@Tyriar I made a new pull request with the suggested modifications.

Tyriar:
@DepthSec it's intentionally a little brief because we don't officially support this option.

PR #1964, opened 2018-10-10
doc: add robot framework document

tumit:


gregvanl:
Hi @tumit Thank you for your interest in the VS Code documentation. We don't list test frameworks in our Languages section and your PR appears to repeat the content already found in the Robot extension README. It is also better to keep extension documentation in the README so that it is kept up to date by the extension author.

PR #1969, opened 2018-10-11
Fix typo in tutorial-django.md in python docs

shivansh007:


gregvanl:
Hi @shivansh007 Thank you for your PR. The existing casing for the Path environment variable is correct.

image

PR #1972, opened 2018-10-11
Include oxford comma in list of OSes supported

ialford:
Updated the README to include an additional comma, representing that each of the three listed operating systems are supported.


gregvanl:
Thanks

PR #1975, opened 2018-10-11
Update link to report issue

OliverMcB:
Issue link is better pointing at the choose screen to increase production.


gregvanl:
@OliverMcB Thank you for your PR. We'd like to keep the link going to the issues overview page in the hope that people will search for an existing issue first, before creating a new issue.

PR #1977, opened 2018-10-12
Updated integrated terminal tip

Streppel:
Menu option is "Terminal" instead of "Integrated Terminal" at this moment (version 1.28)


gregvanl:
Thanks!

PR #1980, opened 2018-10-12
Fix image paths in tutorials/static-website

shivansh007:


gregvanl:
Hi @shivansh007 Thank you for your interest in the VS Code documentations. These images render correctly on our website and we need to keep the image links as they are. I know it is confusing that they don't render correctly on GitHub.

PR #1991, opened 2018-10-13
Remove outdated information about OS X El Capitan

josh-:
The OS X El Capitan Public Beta was released in 2015, so I believe the "Common Questions" section with this item can be removed – more than happy to close this PR if you believe it should stay in however 😄


gregvanl:
Thanks. Nice clean up.

PR #1995, opened 2018-10-14
Add Scope Field

### LINKED ISSUE #1957: Docs: Explain `scope` Field within `configuration` Contribution Point...
the docs for the `configuration` contribution point make no mention of the `scope` field - yet we are warned about setting it / not setting it under certain circumstances. what does `scope` do?

docs:
https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributesconfiguration

example:
https://github.com/Microsoft/vscode-extension-samples/blob/master/configuration-sample/package.json#L42

warning:
https://github.com/Microsoft/vscode/issues/36623
https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs#run-time-diagnostics

thanks 🙏

HrushikeshBodas:
Resolve #1957

### DIFF
--- a/docs/extensionAPI/extension-points.md
+++ b/docs/extensionAPI/extension-points.md
@@ -50,12 +50,14 @@ You can read these values from your extension using `vscode.workspace.getConfigu
             "typescript.useCodeSnippetsOnMethodSuggest": {
                 "type": "boolean",
                 "default": false,
-                "description": "Complete functions with their parameter signature."
+                "description": "Complete functions with their parameter signature,
+                "scope": " Scope in Which the configuration is applicable."
             },
             "typescript.tsdk": {
                 "type": ["string", "null"],
                 "default": null,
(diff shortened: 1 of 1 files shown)


busticated:
@HrushikeshBodas it's a start - thanks! i would ask:

* what options are currently supported?
* is there a default? (seems like the implementation defaults to `window`)
* what is the behavior for each option?
* when would you chose one option over another?
* what if you have a setting which spans "scopes"? (iow, a setting that doesn't fit neatly into one of the available options)
* what negative consequences are there for mismatched scopes?

thanks again for working on this :pray::+1:

gregvanl:
Closing this PR as the contributor hasn't responded and values in the example need to be either 'resource' or 'window', repeating the IntelliSense comment is invalid.

gregvanl:
@HrushikeshBodas Great feedback and we'll make sure it gets into the documentation. The extensibility docs are going through a big refactoring at present.

PR #1997, opened 2018-10-14
fixed broken image path in tutorials section

vjs3:
Inside *tutorials/static-website/create-app.md* the image path was broken and looked like this first image.

With this commit, I have fixed the path and now the image is visible in the document as shown in the second image.

**Current state**
screen shot 2018-10-14 at 9 05 08 pm

**After the fix**

screen shot 2018-10-14 at 9 04 41 pm


vjs3:
Combined with #1999

PR #2003, opened 2018-10-15
Add missing white space in tips-and-tricks.md

jarifibrahim:
This PR
 - Fixes minor typo in the word `everyday`. It should be `every day`.

`Everyday` is an adjective that means **commonplace** or **happening every day**.
`Every Day` is an adverbial phrase that means **each day** or **daily**


gregvanl:
Thanks

PR #2017, opened 2018-10-17
Correct paths for template files

kraigb:


kraigb:
@gregvanl Please publish these changes when you can...fixes two errors in the Django tutorial that a verbatim pointed out.

PR #2019, opened 2018-10-17
Adding more about unity debugging

zosman1:
linking anyone looking for unity debugging back to the debugging doc so they can learn how that works


gregvanl:
Thanks

PR #2027, opened 2018-10-19
Update tutorial-django.md

eric-urban:
Fix typos.

Nitpick: Actual output is in the terminal itself, not the separate Output window. Feel free to ignore this suggestion.


kraigb:
Thanks for those, @eric-urban !

PR #2028, opened 2018-10-20
Update all tutorial .mds

JW33:
Fix paths for images in tutorials.

Images are now showing instead of alternative text


gregvanl:

PR #2030, opened 2018-10-20
Missing commas after introductory prepositional phrases

MeredithGodar:
In general, it seems to be in the convention in the VS Code docs to use commas after introductory prepositional phrases.  These sentences were missing them.


gregvanl:
Thanks

PR #2040, opened 2018-10-21
Adds a tip about editor.action.addSelectionToNextFindMatch to Get Started

janhohner:
I just realized that a feature I use a lot and which I taught a co-worker a couple of days ago is not mentioned in the Tips and Tricks area of the Get Started documentation so I wanted to change that :)


gregvanl:
Thanks!

PR #2041, opened 2018-10-21
Fix some broken links in docs

rkaranam:


gregvanl:
Hi @rkaranam Thank you for your PR but I don't think the last link update in command-line.md is correct.

PR #2046, opened 2018-10-23
Update angular-tutorial.md

fabioloreggian:
Aligning the Angular Getting Started guide with that of the Angular Getting started guide and added a link to the guide


gregvanl:
@fabioloreggian Thank you for your PR. We'd like to keep the user on the current topic and there is already a link earlier to root of the angular.io site.  I don't think we need to launch a browser with --open as some folks find this annoying.

PR #2047, opened 2018-10-23
Image Path Fix in "markdown.md"

hananezlitni:
I fixed a typo in an image path in "markdown.md" ("Outline View" section), which was preventing the image from being displayed.


gregvanl:
Thanks! Subtle as it doesn't affect our website but does mess up GitHub.

PR #2050, opened 2018-10-24
fix a mistake

591097213:
It seems to be red squiggles under the include statements if a `#include` file or one of its dependencies cannot be found.


gregvanl:
Thanks

PR #2053, opened 2018-10-24
Increase readablilty

flash76:


gregvanl:
Hi @flash76, thank you for submitting a PR. We have found that inline code blocks don't render well in headers so we try not to use them. The folder name is 'build' not 'build/'.

PR #2064, opened 2018-10-25
added instructions of where to find extension build help

jluo117:


gregvanl:
Hi @jluo117 Thank you for your interest in the VS Code documentation. Building VS Code extensions is a more advanced topic and not appropriate as a Next Step for someone setting up VS Code.

PR #2065, opened 2018-10-25
Fix broken image links in tutorials

spncrhrstn:
Looks like the images were all moved to a common folder, rather than in the specific tutorials. I am viewing these tutorials on Github directly, I don't know how this would effect to the publishing site. 

There was also no line break on the links at the bottom of each page. This may be due to a [VS Code bug](https://github.com/Microsoft/vscode/issues/39928) as I was experiencing it too while editing the files.


gregvanl:

PR #2077, opened 2018-10-26
Added missing question marks

abbas10m:
makes it consistent with the rest of the questions


gregvanl:
Hi @abbas10m Thank you for submitting a PR. These headers are linked to from many places, including the VS Code product, so we don't want to change the text, even the punctuation.

PR #2080, opened 2018-10-26
Update Python test discovery docs

### LINKED ISSUE #2068: Clarify that "python.unitTest.autoTestDiscoverOnSaveEnabled" applies to all test frameworks
Microsoft/vscode-python#3004

The phrasing at https://code.visualstudio.com/docs/python/unit-testing#_test-discovery and https://code.visualstudio.com/docs/python/unit-testing#_unittest-configuration-settings suggest that `"python.unitTest.autoTestDiscoverOnSaveEnabled"` only applies to `unittest`, when in fact it works with all the test runners.

jgierer12:
Closes #2068 

I decided to completely remove the paragraphs detailing the individual test frameworks, since these are explained more in-depth in the [linked section](https://code.visualstudio.com/docs/python/unit-testing#_test-framework-configurations) anyways.

### DIFF
--- a/docs/python/unit-testing.md
+++ b/docs/python/unit-testing.md
@@ -119,15 +119,9 @@ To enable Nose:
 
 VS Code uses the currently enabled unit testing framework to discover tests. You can trigger test discovery at any time using the **Python: Discover Unit Tests** command.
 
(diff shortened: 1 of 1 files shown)


kraigb:
Good call--thanks for the edits and for handling the open issue!

PR #2086, opened 2018-10-27
Add powershell/conda note on environments docs

d1vanloon:
Add a note mentioning that conda environments cannot be auto-activated if the shell is PowerShell (see https://github.com/Microsoft/vscode-python/issues/2898)


kraigb:
Thanks for adding that detail; I don't typically use or test with PowerShell, so I don't encounter such matters.

PR #2088, opened 2018-10-27
Added Links to Python Docs

liamrosenfeld:
It matches the precedent set up in the rest of the file and should save most readers a few clicks.


kraigb:
Thanks for the contribution!

PR #2095, opened 2018-10-27
Fixed image URLs

drecali:
Image links were broken, so I updated with correct URLs from the repo.


gregvanl:
Please review the VS Code live site at code.visualstudio.com before submitting link issues.

PR #2099, opened 2018-10-28
Fixed typo errors in tips and tricks

rkaranam:
Fixed typos in Customization and Editing hacks sections


gregvanl:
Thanks! Good catch.

PR #2102, opened 2018-10-28
Add --add and --install-extension to the tips and tricks

jorgelbg:
Add examples for the `--add` and `--install-extension` flags to the tips and tricks Command line section.


gregvanl:
Hi @jorgelbg Thank you for your interest in the VS Code documentation. Both --add (for creating multi-root workspaces) and --install-extension (we'd prefer people use the Extensions view workflow) are more advanced switches that we want to leave out of the Get Started content.

PR #2103, opened 2018-10-28
Update linux.md

dee-see:
Add mention of VSCode package in official Arch Linux repository


gregvanl:
Hi @dee-see Thank you for your interest in the VS Code documentation and taking the time to submit a PR. To keep things simple for our users, we're just documenting one AUR package.

PR #2104, opened 2018-10-28
Remove unused import

BoAnd:
Under the section **Create A Django app**. We `import render` which is unnecessary. And under section **Use a template to render a page** we ask the user to add `import render` which is incorrect with the code state in section **Use a template to render a page** if we don't remove the `import render` in section **Create A Django app**. 

I hope it makes sense.


kraigb:
Good catch, thanks for the fix!

PR #2105, opened 2018-10-28
Practical examples of common arguments passed to node

dandv:


gregvanl:
Hi @dandv Thank you for your interest in the VS Code documentation. You have provided good examples but we'd like to keep the attributes section as simple as possible without examples.
If you think one or both of those examples are something people might commonly be looking for, you could add an entry in the Common Questions section at the end of the topic.

PR #2108, opened 2018-10-29
Use apt-key for gpg key management

angelabad:
Please use apt-key instead of install gpg repo key manually.

Cheers


gregvanl:
See https://github.com/Microsoft/vscode-docs/pull/1768

PR #2116, opened 2018-10-30
Fix Jinja links

bracar:
Added link to Jinja on first mention, and changed Jinja links to point to the latest version automatically


kraigb:
Thanks for those!

PR #2123, opened 2018-10-30
Change description of "python.venvPath" setting

pythonsway:
Proposed change should help with right selecting of this path, because many users think that it should be pointing to virtual environment folder itself, but it should be the parent one instead.


kraigb:
Thanks for the edit! I admit I struggled with that description, so I appreciate you taking a pass at it. :)

pythonsway:
I thank you for VSCode! For me it's the best product in the market.

PR #2126, opened 2018-10-31
Update settings.md

anacodes:
Add information related to new settings GUI


gregvanl:
Hi @anacodes Thank you for submitting a PR. Mentioning that something is new is appropriate for release notes but since the GUI is now the default we don't need to mention it in the main document.

PR #2128, opened 2018-10-31
Update flask strftime examples

hrjakobsen:
Different strftime formats are shown, but because the month in the example is May, then two  (%b and %B) of the examples shows the same output. If a month with a longer name such as October is used, then the different format strings has different outputs.


kraigb:
That's a good point, something I didn't notice when I wrote the tutorial in May.

PR #2130, opened 2018-10-31
Add cursorColumnSelect tip

lnicola:


gregvanl:
Hi @lnicola Thank you for your PR. There are existing keyboard shortcuts for Windows and macOS, just not Linux. I'm not sure we need to more to this section since keyboard navigation is covered in the editor topic https://code.visualstudio.com/docs/editor/codebasics#_column-box-selection.

PR #2132, opened 2018-10-31
Add a mention about vertical selection using keyboard

lnicola:
Similar to #2130. Sorry, I didn't realize it was already documented -- I've been using this feature since it was implemented. Feel free to close if you think it's too inconsequential, I'm not sure it's worth it.

@gregvanl


gregvanl:
Thanks! Cross linking is a good thing.

PR #2137, opened 2018-11-01
Update and tidy Linux setup documentation a bit

Gawdl3y:
This PR just adds/tweaks some punctuation and grammar where I feel the flow of sentences is improved by doing so, as well as adds formatting where I believe it may have been overlooked. Additionally, I updated the information on fixing Chinese character display on Ubuntu to make use of the new settings interface.


gregvanl:
Thanks

PR #2138, opened 2018-11-01
Made a sentence more clear.

512jay:


gregvanl:
@512jay Thank you for your interest in the VS Code documentation.  The current text is preferred since "source is edited" matches better with "while ... is running".

512jay:
Ok, thanks for the input, the change in tense threw me before, and it took me a little while to understand the sentence. However, a friend of mine read your sentence and had no problem with it, so I guess it is fine.

PR #2139, opened 2018-11-01
Replace apt-get with apt

Gawdl3y:
This is perhaps a controversial change, but I believe it to be a positive one. Debian and Ubuntu have both moved to using `apt` rather than `apt-get` as the primary interface for their package manager as of a few years ago, as it contains numerous improvements to usability and consolidates a few other commands into one. This should only affect fairly old versions of the distributions - Ubuntu switched over to it in 16.04 (two LTS releases ago), for example, and 14.04 is only a few months away from EOL.


Tyriar:
I don't think we should do this until EOL Ubuntu 14.04 which is April 2019, Debian 7 might also not have apt which is EOL in 31 May 2019.

PR #2140, opened 2018-11-01
Minor grammar fixes and clarification

kjowong:
- Added a comma as it didn't sound quite right
- One tutorial had "LTS" as "Long Term Support" and another as "Long Term Stable"
- One page had both "redeploy" and "re-deploy"
- Changed exclamation mark to a colon.


gregvanl:
Thanks!

PR #2153, opened 2018-11-02
Use zypper to add openSUSE repo

milliams:
Rather than relying on creating the repo file by hand, use the tool to do it for us.


Tyriar:
This was also opened in https://github.com/Microsoft/vscode-docs/pull/1847#issuecomment-422678106, that comment indicates that we should have a `gpgkey=` arg as well?

PR #2158, opened 2018-11-02
linux.md: update apt deb install instructions

mvo5:
The more recent versions of apt support installing deb packages directly without the dpkg -i && apt-get install -f trick. This should work on Ubuntu 14.04 and later and on the current Debian stable.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -17,6 +17,10 @@ The easiest way to install Visual Studio Code for Debian/Ubuntu based distributi
 
 ```bash
 sudo apt install ./<file>.deb
+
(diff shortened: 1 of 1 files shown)


mvo5:
Sorry, my memory was faulty - the feature to install debs directly from `apt` was added in 1.1~exp1 which was in 14.10 - not in 14.04 (which is slightly embarrassing since I added that feature :blush:) . 

So you *may* consider keeping the current instructions. However given that 14.04 will EOL soon(ish) you may also consider putting the nicer and more modern instructions here - the benefit is that apt will do automatic dependency resolution immediately and if for some reason some deps are missing it will warn in a clean way about this.

PR #2170, opened 2018-11-07
Azure Repos extension name

liguori:
Visual Studio Team Services extension renamed in Azure Repos


gregvanl:
Thanks for tracking this. We didn't know when the extension name was going to be updated or what it would be.

PR #2181, opened 2018-11-13
Fix Framework name

pinguin999:
In Changes for Django apps was Flask


kraigb:
Thanks so much for catching that one! It certainly escaped my eyes the many times I went over the text.

PR #2182, opened 2018-11-13
📦 NEW: Debug Stack Frame Highlight Colors

ahmadawais:
In the new VSCode release v1.9 there are two new colors which are not yet present in the theme reference. This PR adds them.


gregvanl:
Thanks!

PR #2185, opened 2018-11-14
Cleanup default keyboard shortcuts

chyipin:
Remove shortcuts that are no longer defaults


gregvanl:
@chyipin Thank you for submitting a PR. Including unassigned commands is intentional so that users can find the command id for common or like actions and create their own keybindings.

chyipin:
I see, I understand the intention now.

Maybe the section header should be renamed appropriately instead of `Default Keyboard Shortcuts`?
When it is unassigned to a keybinding, it isn't a shortcut anymore. 😛 

https://code.visualstudio.com/docs/getstarted/keybindings#_default-keyboard-shortcuts

gregvanl:
We can certainly make that clearer in the content above.

PR #2193, opened 2018-11-15
explain the setting required for convert to async

johnpapa:
> Note that this feature requires that the setting `javascript.validate.enable` is set to true.


mjbvz:
Thanks!

johnpapa:
thanks for accepting it :)

PR #2201, opened 2018-11-19
Add files.defaultLanguage description to FAQ

inarius:
This is a common question:
https://stackoverflow.com/questions/35904221/change-vscode-default-language-for-new-files
https://stackoverflow.com/questions/48456682/how-to-set-default-file-type-for-new-files-in-vs-code

### DIFF
--- a/docs/languages/overview.md
+++ b/docs/languages/overview.md
@@ -114,3 +114,14 @@ You can also configure full file paths to languages if needed. The following exa
 ```
 
 Note that the pattern is a [glob pattern](https://en.wikipedia.org/wiki/Glob_%28programming%29) that will match on the full path of the file if it contains a `/` and will match on the file name otherwise.
+
+### How do I set the default language for new files?
+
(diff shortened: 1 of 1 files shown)


docs/languages/overview.md
```diff
@@ -114,3 +114,14 @@ You can also configure full file paths to languages if needed. The following exa
 ```
 
 Note that the pattern is a [glob pattern](https://en.wikipedia.org/wiki/Glob_%28programming%29) that will match on the full path of the file if it contains a `/` and will match on the file name otherwise.
+
+### How do I set the default language for new files?
+
+Using the `files.defaultLangauge` [setting](/docs/getstarted/settings.md), you can map all new files to a default language. Whenever a new blank file is opened, the editor will be configured for that langauge mode.
```
gregvanl: Typo: setting is `files.defaultLanguage`, you have the 'u' and 'a' transposed.

inarius: Darn, thanks!

gregvanl:
Thanks!

PR #2218, opened 2018-11-22
Update emmet.md

afroguy16:
There is no way to search for emmet in Vscode


gregvanl:
Hi @afroguy16 Emmet abbreviations are included in VS Code by default and don't require installing an extension.

PR #2250, opened 2018-12-01
Improving readability of the page

OwainWilliams:
It's just a small update but I feel it makes the sentence read better.


gregvanl:
Hi @OwainWilliams Thank you for your interest in the VS Code documentation. I prefer the way the current text reads and think having the word "method" used twice in different contexts (method as technique or process versus method as a programming entitiy) could be confusing.

PR #2253, opened 2018-12-02
Cmd -> Alt

lisfer:
Looks like this is old behavior: Cmd + Click => open in new panel.
For now All + Click is used instead.

Cmd + Click allows to select several files in the file explore


gregvanl:
Thanks!

PR #2257, opened 2018-12-05
Fix linter error on strftime

### LINKED ISSUE #2252: self.log_date.strftime in section "Python/Django Tutorial" chapter "Define models"
In [section "Python/Django Tutorial" chapter "Define models"](https://code.visualstudio.com/docs/python/tutorial-django#_define-models) there is pylint [E1101]:
[pylint] Instance of 'DateTimeField' has no 'strftime'...

Refering to [stackoverflow](https://stackoverflow.com/questions/11909071/using-strftime-on-a-django-datetime-produces-a-utc-time-in-the-string)
"timezone.localtime" function has fixed the problem.

kraigb:
Close #2252

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -678,6 +678,7 @@ For example, add the following class in `models.py` to define a data model that 
 
 ```python
 from django.db import models
+from django.utils import timezone
 
 class LogMessage(models.Model):
     message = models.CharField(max_length=300)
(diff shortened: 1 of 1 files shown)


docs/python/tutorial-django.md
```diff
@@ -678,14 +678,16 @@ For example, add the following class in `models.py` to define a data model that
...
 
 class LogMessage(models.Model):
     message = models.CharField(max_length=300)
     log_date = models.DateTimeField("date logged")
 
     def __str__(self):
         """Returns a string representation of a message."""
-        return f"'{self.message}' logged on {self.log_date.strftime('%A, %d %B, %Y at %X')}"
+        date = timezone.localtime(self.log_date)
+        return f"'{self.message}' logged on {date.strftime('%A, %d %B, %Y at %X')}
```
gregvanl: @kraigb Are you missing a trailing double quote at the end of the return statement

gregvanl: I checked the example repo and added the trailing double quote to match.

kraigb: Thanks. I hadn't noticed that the quote disappeared...wonder why that happened...but thanks for fixing it and the sample!

PR #2271, opened 2018-12-12
Fix sample snippet in .bash_profile.

Elderry:


gregvanl:
Hi @Elderry This PR came up once before and was closed. See https://github.com/Microsoft/vscode-docs/pull/1781.

PR #2277, opened 2018-12-14
Add the Lilypond language

Marr11317:
Added [here](https://github.com/TrudyFirestone/VSLilyPond/blob/master/lilypond/package.json)


gregvanl:
Hi @Marr11317 Thank you for your interest in the VS Code documentation. The Lilypond language looks very interesting but it isn't one of our core languages so we don't want to add it to our language identifiers list. Users can find the language id from the Lilypond extension contribution section which will register the language id.

From truefire.lilypond:

image

PR #2282, opened 2018-12-17
Specify `shell` and `shellArgs` settings allowed scopes.

nuno-andre:
`terminal.integrated.shell.*` and `terminal.integrated.shellArgs.*` can only be set at user scope.


gregvanl:
Thanks! Nice to include that clarification in the topic.

PR #2287, opened 2018-12-20
Add more font size tips

joech4n:
Added more examples per:
https://github.com/Microsoft/vscode/issues/21489#issuecomment-353609526

I also think it should be added [here](https://code.visualstudio.com/docs/getstarted/settings) in the default settings page, but I'm not sure if it actually has a default value...

### DIFF
--- a/docs/getstarted/tips-and-tricks.md
+++ b/docs/getstarted/tips-and-tricks.md
@@ -178,10 +178,17 @@ Format on paste
 "editor.formatOnPaste": true
 ```
 
-Change the font size
+Change the font size of various UI elements
 
 ```json
-"editor.fontSize": 18
+// Main editor
+"editor.fontSize": 18,
+// Terminal panel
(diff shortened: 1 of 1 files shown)


docs/getstarted/tips-and-tricks.md
```diff
@@ -178,10 +178,17 @@ Format on paste
...
+Change the font size of various UI elements
 
 ```json
-"editor.fontSize": 18
+// Main editor
+"editor.fontSize": 18,
+// Terminal panel
+"terminal.integrated.fontSize": 14,
+// Output panel
+"[log]": {
```
gregvanl: I think this needs to be `"[Log]"` with a capital L.

gregvanl: I'll take care of this.

gregvanl:
Thanks!

PR #2290, opened 2018-12-22
Grammar and usage

BlackHoundNate:
This pull request contains proposed changes to the Python debugging document. Most of the changes fix (what I take to be) grammatical errors, but a few are stylistic / for clarity.


kraigb:
Thanks for the careful editing! Much appreciated.

BlackHoundNate:
Thanks for the prompt attention and the kind note. Although I code every day / work in the field, this is my first merged PR on a public project. I'm very grateful to have had a good experience.

kraigb:
Glad to have you on board with PRs! We really do appreciate when people take the time to catch some of the small details that often slip by in the volumes of documentation we handle. It's open-source at its best. :)

PR #2297, opened 2018-12-27
Fix Indentation Rules

namelessGonbai:
In this document "increaseIndentPattern" and "decreaseIndentPattern" were reversed.
Also, this document did not include the case of ")" when "decreaseIndentPattern".


rebornix:
Thanks for your help!

PR #2298, opened 2018-12-27
Modify an incorrect word

zwdsjl:
This word should be from the above registerReferenceProvider copied.


octref:
Thanks! This file is auto-generated and I have fixed it in the source, but I'll take the contribution.

PR #2309, opened 2019-01-03
Make `PYTHONPATH` a code literal

brettcannon:
Also fix some links to not be version-specific.


kraigb:
Yes, I'd purposely not used a code format for PYTHONPATH because it occurs often enough that the formatting becomes cumbersome rather than helpful. But we need to keep the changes the links. Brett, can you remove the ``'s on PYTHONPATH?

brettcannon:
Easier to just open a new PR to fix the URLs.

PR #2319, opened 2019-01-07
update line or block comments, honoring the current language v:1.30.1

datamweb:
snippet variables that insert line or block comments, honoring the current language. Use BLOCK_COMMENT_START and BLOCK_COMMENT_END for block comments and LINE_COMMENT otherwise.
v : 1.30.1


gregvanl:
Thanks!

PR #2320, opened 2019-01-07
Fix link in syntax highlight guide

jayjun:
The link to Matt Neuberg’s blog only works if all characters are lowercased.


gregvanl:
Thanks!

PR #2321, opened 2019-01-07
Update nodejs-tutorial.md

JosephAllen:
Updated express command to use pug so that users to not get the vulnerability messages related to jade


gregvanl:
Hi @JosephAllen Thank you for your interest in the VS Code documentation and taking the time to submit a PR. It looks like they are going to update the express generator to no longer use jade so I'd prefer to leave the tutorial with the simple `express myExpressApp` command line.

```
C:\Users\gregvanl\deleteMe>express myExpressApp

  warning: the default view engine will not be jade in future releases
  warning: use `--view=jade' or `--help' for additional options
```

PR #2331, opened 2019-01-10
fix typo for grammar scopes example

clarkio:
There is a typo that should show `source.abc` which is used as the example throughout this documentation. This PR fixes those instances where `souce.abc` was found to avoid any confusion.


octref:
Very nice, thank you!

PR #2332, opened 2019-01-11
Syntax highlighting uses Oniguruma regular expressions

jayjun:
Ruby uses Onigma, a fork of Oniguruma, for regular expressions since version 2.0. So it’s no longer accurate to say TextMate uses _Ruby_ regular expressions.

VS Code uses Oniguruma to [implement](https://github.com/Microsoft/vscode-textmate/blob/bd860d3d87b37343cc61239e5e90d813c1b3595d/package.json#L30) grammars.


octref:
Thanks for the clarification!

PR #2334, opened 2019-01-11
Update "Tune your settings" part

rlax:
Add info about new default editor for settings (visual one) and guidance for switching to JSON settings, that are listed below in tips.

Reference: https://github.com/Microsoft/vscode-tips-and-tricks/pull/142


gregvanl:
Thanks!

PR #2336, opened 2019-01-12
Update file path in tutorial-djanjo.md

ArthurEzenwanne:
Updated an invalid file path: template/hello_there.html --> template/hello/hello_there.html


kraigb:
@ArthurEzenwanne Thanks for catching that!

PR #2340, opened 2019-01-15
Fix illegal regex in language configuration guide

arxanas:
This regex was missing a closing parenthesis at the very end. Discovered by pasting it into https://debuggex.com. Confirmed by testing the pattern `if (true) {` against this regex.


rebornix:
@arxanas it seems `"^((?!\\/\\/).)*(\\{[^}\"'`]*|\\([^)\"'`]*|\\[[^\\]\"'`]*))$"` can not be parsed correctly ?

image

arxanas:
Hm, I must have made a mistake when I was using this regex, probably due to the `\\(` in some way. Thanks for following up!

PR #2350, opened 2019-01-18

JeffreyRoosendaal:
`code ./helloworld` doesn't make any sense here, as it just opens a new empty file which has nothing to do with the tutorial. Left me puzzled for a while actually. It should probably be `cd ./helloworld`, since creating a new extension doesn't take you to the new folder automatically.

### DIFF
--- a/api/get-started/your-first-extension.md
+++ b/api/get-started/your-first-extension.md
@@ -31,7 +31,7 @@ yo code
 # ? Initialize a git repository? Yes
 # ? Which package manager to use? npm
 
-code ./helloworld
+cd ./helloworld
 ```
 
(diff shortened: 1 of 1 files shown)


octref:
> It should probably be cd ./helloworld, 

Are you following the instruction right above that line, which is to use `yo code` to generate a `helloworld` folder?

JeffreyRoosendaal:
After playing around a little I see what caused the confusion here. I used `yo code` directly from the VS Code terminal. Since the terminal is still in the folder below, `cd ./helloworld` would actually be the first thing to do after `yo code`.

_However_, using `yo code`, `code ./helloworld` from the command prompt, it makes perfect sense, since it opens VS Code in the correct folder 🙂

My bad.

PR #2352, opened 2019-01-19
Fixed typo "by instead of but"

saadsaifse:
Fixed a typo error in the line "You can test that your gulp installation was successful by typing `gulp -v`."


gregvanl:
Thanks!

Issue #2344, opened 2019-01-17
runOptions is not give in schema for task.json


vvavrychuk:
Seems there is runOptions property https://code.visualstudio.com/docs/editor/tasks#_run-behavior. But somehow it has not managed to get into https://github.com/Microsoft/vscode-docs/blob/master/docs/editor/tasks-appendix.md

gregvanl:
Fixed with https://github.com/Microsoft/vscode-docs/pull/2353

PR #2353, opened 2019-01-20
Added runOptions property in the schema for task.json

waseemahmad31:
Issue #2344 

Issue: runOptions property https://code.visualstudio.com/docs/editor/tasks#_run-behavior. But it is missing in https://github.com/Microsoft/vscode-docs/blob/master/docs/editor/tasks-appendix.md

Fix: Added runOptions property in the schema for task.json


gregvanl:
Thanks!

PR #2357, opened 2019-01-22
Updated gulp task creation code to v4.0

saadsaifse:
Gulp v4.0 no longer supports synchronous tasks. The last code was probably written for v3, that did not work when i installed gulp v4.0. So I updated the code as per official guidelines [here](https://gulpjs.com/docs/en/getting-started/creating-tasks#compose-tasks)


gregvanl:
Thanks!
The funny thing is that this was on my TODO list today.

saadsaifse:
Glad that I got the chance of PR on time :100:

PR #2360, opened 2019-01-23
Move cmd+p to top of docs

jotto:
I'd argue this is more commonly used than the other features yet here only exists as a minor subpoint. Why not put it at the top?


gregvanl:
Good suggestion

PR #2361, opened 2019-01-23
Update functions tutorial.

fiveisprime:
This update emphasizes the local development experience by moving the
Azure sign up/in to the end. This also updates the context to more of a
"Try Azure Functions" experience.

This is an experiment to determine if showing the development experience
first increases the conversion rate and decreases the funnel drop off. This
also allows customers to try Functions for free in a different way than with
app service.

I intentionally didn't change the TOC because I didn't want to break the
consistency of the links. I'm open to changing that if it makes sense.


gregvanl:
I tuned a couple of tutorial page titles with https://github.com/Microsoft/vscode-docs/commit/a8c20392c5eee4009941d02df84974a4577c6a5d

fiveisprime:
Thanks, Greg!

PR #2382, opened 2019-02-06
Updating February update release localization contributors.

janbaltus:


shawndon:
Looks good to me as we [Jan and I] discussed.

PR #2399, opened 2019-02-10
[api/extension-guides/debugger-extension] Fix typos in guide

gkalpak:


gregvanl:
Thanks, especially for finding the right commit reference.

PR #2406, opened 2019-02-13
Update mac.md

NestorNarvaezAFP:
I tested on a Macbook this issue


gregvanl:
Hi @NestorNarvaezAFP Thank you for your interest in the VS Code documentation. There are many ways to right-click on a Mac to bring up the context menu (two finger tap, ctrl+tap, assign a corner of the trackpad) as well as your recommendation to tap and hold. We would like to keep the documentation as general as possible so we just say "right-click" and Mac users can bring up the context menu as they prefer. I will add a comment to say that the Option is on the context menu. Thank you for pointing out a possible source of confusion.

PR #2434, opened 2019-02-24
Fix typo

mellervin:
Change 'it' to 'there'


gregvanl:
@mellervin Thank you for your interest in the VS Code documentation. I think the current text reads correctly. "It is a foo that does bar".

PR #2441, opened 2019-03-01
Fix typo in FoldingRange section

laginha87:
I think the initial idea was for it to read: "line must be zero or larger", switched it to "line must be bigger than zero"


gregvanl:
Hi @laginha87 This content is auto-generated from https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts in the vscode product repo. Feel free to create a PR to update the comment in that repo.

PR #2444, opened 2019-03-04
Fix typo 43 -> 3333

KexyBiscuit:
See https://github.com/Microsoft/TypeScript/releases/tag/v3.3.3333.


gregvanl:
@KexyBiscuit Thanks for noticing. We'll just show 3.3.3 to keep it simple.

PR #2454, opened 2019-03-06
Click is not accessible language

mscharlock:
Came up in another doc I am reviewing (that references this one) that this doc uses "click." Click is not an accessible term (because it implies that you are using a mouse and can click it). Select is the preferred MSFT voice term because it's neutral about how a person goes about doing the action but implies the same end action. See Microsoft Voice/Style guidelines for more details.

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -18,13 +18,13 @@ One of the key features of Visual Studio Code is its great debugging support. VS
 
 VS Code has built-in debugging support for the [Node.js](https://nodejs.org/) runtime and can debug JavaScript, TypeScript, and any other language that gets transpiled to JavaScript.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @mscharlock Thank you for your interest in the VS Code documentation. I'm reluctant to merge your PR since "select" usually means selecting a region of text in an editor and VS Code has may "Select" commands.

image

gregvanl:
@mscharlock Your PR has merged conflicts so I'll apply your changes manually. The last update related to the editor margin should remain click as only a mouse action can toggle a breakpoint there. I can add a mention of using the F9 to also toggle a breakpoint.

gregvanl:
https://github.com/Microsoft/vscode-docs/commit/b464a1ae705b41bb987d801653c3253f7a05b923

PR #2463, opened 2019-03-07
Remove inline HTML from code example

### LINKED ISSUE #2447: Remove the HTML concatenation examples
The write-ups for getting started with [Flask](https://code.visualstudio.com/docs/python/tutorial-flask) and [Django](https://code.visualstudio.com/docs/python/tutorial-django) are **fantastic**! However, I'd love to see the example of HTML concatenation, which is called out as a bad practice, removed from the example. If it's something developers shouldn't be doing, it's probably best to leave it out of the docs.

kraigb:
Close #2447

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -334,9 +334,11 @@ During your work with Django or any other library, you may want to examine the c
 
 ## Use a template to render a page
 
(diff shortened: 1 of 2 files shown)


GeekTrainer:
It looks fantastic! And I love the tightening of the description by putting the code inline. 🚢

kraigb:
Excellent. Merging now.

PR #2464, opened 2019-03-08
add explanation for version mismatch error

kejxu:


octref:
There's already a warning notification saying version mismatch, so I think we shouldn't repeat ourselves again.

PR #2468, opened 2019-03-10
Correct "dashes" to "hyphens" under Bookmarks

kyerlasswell:
While most people wouldn't confuse the two, there is technically a difference so I figure it's best to have the docs be as technically accurate as possible.


gregvanl:
Thanks

PR #2486, opened 2019-03-15
Remove reload after extension install description

glitchperfect:
Reloading vscode after installing the chrome debugger isn't needed anymore


gregvanl:
Thanks!

PR #2510, opened 2019-03-22
Remove duplicate breadcrumb.background in theme color reference

emroussel:
The `breadcrumb.background` line was in the theme color reference docs twice, so this removes one of the entries.


gregvanl:
Thanks!

PR #2515, opened 2019-03-26
Update reactjs-tutorial.md

xiaohuoni:


gregvanl:
Hi @xiaohuoni Thanks for pointing out the inconsistency. We like to use quotes to text that the user will type rather than an inline code block so I won't be taking your PR but we really appreciate you taking the time to submit a PR.
Thank you as well for noticing the unnecessary leading spaces.

gregvanl:
https://github.com/Microsoft/vscode-docs/commit/8d4892c1e0aed878dd1fbb1de72027a4944e6de4

xiaohuoni:
👍

PR #2517, opened 2019-03-26
Parallel execution of dependsOn tasks.

engelchrisi:
dbaeumer commented on Sep 18, 2018
If you list more than one task in the dependsOn property they are executed in parallel.


gregvanl:
Thanks!

PR #2520, opened 2019-03-28
Fix small grammatical error in LSP sample

radu-matei:
Minor change: _This way, Language Servers can be implemented in any language and run in [~~its~~ their] own process_


gregvanl:
Thanks!

PR #2527, opened 2019-03-29
deprecate the word workspace when folder is meant

jon-freed:
It's clear from questions like [this one](https://stackoverflow.com/questions/44629890/what-is-a-workspace-in-vs-code/) that both the documents and the UI could be clearer about what is a workspace versus an open folder.  I'm proposing that if you don't have an open .code-settings file, then you don't have an open workspace.

### DIFF
--- a/docs/getstarted/settings.md
+++ b/docs/getstarted/settings.md
@@ -7,35 +7,44 @@ PageTitle: Visual Studio Code User and Workspace Settings
 DateApproved: 3/7/2019
 MetaDescription: How to modify Visual Studio Code User and Workspace Settings.
 ---
-# User and Workspace Settings
+# User, Workspace, and Folder Settings
 
 It is easy to configure Visual Studio Code to your liking through its various settings. Nearly every part of VS Code's editor, user interface, and functional behavior has options you can modify.
 
-![VS Code settings](images/settings/hero.png)
+[Default settings](/docs/getstarted/settings.md#default-settings) are listed towards the bottom of this page.
 
-VS Code provides two different scopes for settings:
+VS Code provides three different scopes for settings:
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @jon-freed Thank you for your interest in the VS Code documentation. I'm reluctant to take this PR since the term "workspace" to refer to a simple folder or root project is not deprecated and common in the VS Code UI (commands, Settings, extension recommendations). I agree the naming is confusing when talking about multi-root workspaces.
Perhaps it helps to think of a multi-root workspace with a .code-workspace file as a special case of a simple folder workspace. The user can add folders to an existing workspace with the Add Folder to Workspace command. We have also seen that the use of multi-root workspaces is pretty low, around 1%, so continuing to use Workspace for the single root case matches most user scenarios.

jon-freed:
@gregvanl  - Thank you for the consideration.  I think I understand your reasons for closing this PR.

To clarify, I know that "the term 'workspace' to refer to a simple folder or root project is [indeed] not deprecated" at this time.  I was trying to suggest through this PR that if the PR was accepted, then at that point that colloquial definition would be deprecated.  That's why I used "deprecated" (past tense) in my proposed revisions to the documentation.

Because the definition is in such common use, I knew this proposal would be a stretch.  :)  Through my edits, I tried to strike a balance between the colloquial definition of "workspace" and the actual technical definition of "workspace" while instructing the reader about the actual technical differences:   a colloquial workspace, aka a folder, has settings in a different location than an actual workspace, so if you don't know which one you're dealing with, you won't know where to go to view or revise those settings.  

Perhaps it would be easier and/or better to do away with that difference in settings storage than it would be to try to deprecate the colloquial meaning of workspace.  However, the former would require a code change rather than or in addition to a documentation change.

PR #2535, opened 2019-04-02
Add "statusBarItem.prominentForeground" to theme color docs

misolori:
We will be introducing a new color theme token, `statusBarItem.prominentForeground` and this is to update the docs to reflect this new color.

**Note:** this should only be merged _AFTER_ 1.33 is released since this will only be available in 1.34

FYI @aeschli


rediffusion:
@misolori 

 What this feature can do? I don't understand how it works?

> "statusBarItem.prominentForeground": "#5632b9"

🚨 Can you help?

PR #2536, opened 2019-04-03
changed version number

ramlaxman:


gregvanl:
@JoshuaPartlow I took this PR so it is up to date with the python site but you may want to update the topic to avoid version churn.

PR #2539, opened 2019-04-04
Information about the NVDA screen reader add-on

pawelurbanski:
Contains a paragraph in the screen reader section that points to the NVDA add-on I created. It is related to fixes added in the 1.33 insider edition.
The add-on enhances the following use cases:
* Reading the item completed from the intelisense code completion without reading back the entire line of code,
* Fixes losing focus when coding in the editor and using the escape key to leave the intelisense suggestion popup,
* Provides a settings.json file that can be useful when working with a screen reader.

The add-on with a redme file can be found here: https://github.com/pawelurbanski/nvda-for-vs-code

@jrieken - this pull request for the documentation makes our fixes complete, and makes working with VS Code much better.


gregvanl:
@pawelurbanski Thank your for the add-on and this PR. I'm made some small edits since 1.33 is now in stable. https://github.com/Microsoft/vscode-docs/commit/86d030030b9f637556dd1edfdb621134c7765c50

PR #2543, opened 2019-04-07
clarified what priority means

johnpapa:


gregvanl:
Hi @johnpapa The vscode API reference is auto-generated from https://github.com/Microsoft/vscode/blob/master/src/vs/vscode.d.ts. You can submit a PR in the vscode product repo to change the comment and the API reference will be updated next release.

PR #2547, opened 2019-04-09
Update faq.md

mvdobrinin:
Proposing "OS's" change to "OSs". Most discussions seem to be in agreement about the current form being incorrect - should only be used for possessives.


mvdobrinin:
Looks to me like the preferred forms should be `OSes` or `OSs` since it is an acronym. There is some good debate on the topic here - http://technotes.whw1.com/computer-related/operating-systems/22-what-is-the-plural-of-os-operating-system.

mvdobrinin:
These two links would seem to confirm this as well.
- https://en.wiktionary.org/wiki/OSes
- https://blog.apastyle.org/apastyle/abbreviations/#Q13

Happy to hear your thoughts on this.

gregvanl:
Agreed, Thanks!

Matches Microsoft Cloud Style Guide as well:

image

PR #2550, opened 2019-04-09
Added Windows 10 OpenSSH installation

oogetyboogety:
https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse


kraigb:
Thanks for the contribution!

PR #2555, opened 2019-04-11
Update userdefinedsnippets.md

rongjie:
According to For_Loop sample, seems For_Loop is the snippet name rather than For Loop.


gregvanl:
Thanks!

PR #2559, opened 2019-04-12
clarified instructions for Windows command prompt on WSL

mikeblome:


docs/cpp/config-wsl.md
```diff
@@ -57,16 +57,20 @@ To successfully complete this tutorial, you must do the following steps:
 
 ## Create a workspace
 
-At a Windows command prompt, create an empty folder called `projects` (just like you did on Ubuntu) where you can place your helloworld project, and other VS Code projects you might create. Then create a subfolder called `helloworld`, navigate into it, and open VS Code (`code`) in that folder (`.`) by entering the following commands:
+In Windows, you will need an empty folder called `projects` (just like you now have on Ubuntu) where you can place your helloworld project, and other VS Code projects you might create. 
 
-    ```bash
```
gregvanl: The code block language bash should not have been displayed since it is just for syntax highlighting but it looks like 4 leading spaces versus 3 confused the Markdown parser. We usually use bash for all command line code blocks and I didn't know cmd was an allowed language identifier. cmd seems to render well (commands in red).

image

docs/cpp/config-wsl.md
```diff
@@ -57,16 +57,20 @@ To successfully complete this tutorial, you must do the following steps:
...
+
+   ```cmd
+   mkdir projects
+   cd projects
+   mkdir helloworld
+   cd helloworld
+   code .
+   ```
+
+The **code .** command opens VS Code for Windows in the current working folder, which becomes your *workspace*. When you first open the folder, VS Code creates a subfolder called `.vscode` and adds a file called `settings.json` to it. You can look at that file if you like but we won't be concerned with it in this tutorial. Our task is to add three additional files to the `.vscode` folder that will tell VS Code how to compile and debug our program:
```
gregvanl: You can just say "VS Code" and don't need to specify the platform.

gregvanl: VS Code won't actually create a .vscode folder when you first open a folder. The user will need to create some .json file like settings.json, task.json, or launch.json first.

gregvanl:
@mikeblome I've merged the PR and just left you a few comments for later.

PR #2560, opened 2019-04-12
Update keybindings.md

cored:
The file contains the wrong menu path to get to the preferences option: 

Screen Shot 2019-04-12 at 5 58 36 PM

### DIFF
--- a/docs/getstarted/keybindings.md
+++ b/docs/getstarted/keybindings.md
@@ -26,7 +26,7 @@ For doing more advanced keyboard shortcut customization, read [Advanced Customiz
 
 ## Keymap extensions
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @cored File > Preferences > Keymaps is correct for Windows and Linux. In the paragraph above about Keyboard Shortcuts, we mention that it is Code > Preferences > Keyboard Shortcuts on macOS. Do you think we need to add the same parenthetical statement for Keymaps? We assumed someone reading the article would have understood from the earlier example.

cored:
@gregvanl Oh. I was reading too fast I guess and didn't read the Mac OSX part. It was my mistake, thanks for the fast response. You can close this one.

PR #2563, opened 2019-04-13
Update config-mingw.md

MKcKr0:
I followed this doc and used a gi-bash, but at the second time of debugging, there is a problem "Unable to start connection to GDB"。Now I use the default PowerShell,  everything is OK at present.


WardenGnaw:
@MKcKr0 Thank you for your contribution.

Can you also post an issue with what you are encountering with you use git-bash at https://github.com/Microsoft/vscode-cpptools/issues?

For the current changes, Powershell is a shell, so there is no need to explicitly state using that shell in the docs.

gregvanl:
Closing based on dev team comment

PR #2565, opened 2019-04-14
add dotnet core razor to language identifiers

haydenbr:
I stumbled across this while trying to enable emmet for cshtml files for .Net Core razor. Thought I would add it to the list in case it helps anyone else.


gregvanl:
Hi @haydenbr Thank you for your PR. Given this language id is provided by the C# extension, I think the language id is already discoverable from the extension Contributions Languages section.

image

PR #2581, opened 2019-04-17
Update "new terminal" command

luabud:
Now it shows "New Terminal" instead of "Create New Integrated Terminal":
image


gregvanl:
Hi @luabud Thank you for submitting a PR. The Command Palette command is still "Terminal: Create New Integrated Terminal" so the documentation is correct since it is mentioning the Command Palette and not the Terminal menu.

image

PR #2583, opened 2019-04-18
Add tutorial for Azure Functions

kraigb:


docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+    ![Selecting the virtual environment created with the project](images/functions/select-venv-interpreter.png)
+
+> **Tip**: Whenever you want to create another function in the same project, use the **Create Function** command in the **Azure: Functions** explorer, or use the **Azure Functions: Create Function** command from the Command Palette. Both commands prompt you for a function name (which is the name of the endpoint), then creates a subfolder with the default files.
+>
+> ![New Function command in the Azure: Functions explorer](images/functions/function-create-new.png)
+
+## Examine the code files
+
+In the newly-created function subfolder you see three files: `__init__.py`, which contains the function's code, `functions.json`, which describes the function to Azure Functions, and `sample.dat`, which demonstrates that a function subfolder can contain additional files if needs be. You won't be using `sample.dat`, so you can delete it if you want.
```
sptramer: EDIT (clarity suggestion): Change after the `:` to:
```
The Azure Function's code in __init__.py, the Function configuration in `functions.json`, and a sample data file `sample.dat`.
The sample data only exists to show other files can be added to a function subfolder, and can be deleted for this project.
```

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+Let's look at `functions.json` first, then the code in `__init__.py`.
+
+### functions.json
+
+The functions.json file provides the necessary configuration information for the function itself:
```
sptramer: In general make sure that there's some kind of way to distinguish between `Function` (Azure Function) and `function` (Python function object). This seems to be a problem we encounter a lot when writing Functions documentation, maybe check with Glenn for style advice?

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "$return"
+    }
+  ]
+}
+```
+
+You can see that `scriptFile` identifies the startup file for the code, which must contain an function named `main`. You can factor your code into multiple files so long as the file specified here contains a `main` function.
```
sptramer: COPY (edit): Remove `You can see that`
COPY (typo): `an function` -> `a function`

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+The `bindings` element the contains two objects, one to describe incoming requests, and the other to describe the HTTP response. You can see that for incoming requests (`"direction": "in"`), the function responds to HTTP get or post requests and doesn't require authentication. The response (`"direction": "out"`) is an HTTP response that returns whatever value is returned from the `main` function.
```
sptramer: COPY (passive): Remove `You can see that`
COPY (style): Use `GET` and `POST` when referring to HTTP methods (would recommend stylizing inside `` ` ``)

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+             "Please pass a name on the query string or in the request body",
+             status_code=400
+        )
+```
+
+The important parts of the code are as follows:
+
+- You must import `func` from `azure.functions`; importing the logging module is optional but recommended.
+- The required `main` function receives a `func.request`  object named `req`, and returns a value of type `func.HttpResponse`. You can learn more about the capabilities of these objects in the [func.HttpRequest](https://docs.microsoft.com/python/api/azure-functions/azure.functions.httprequest?view=azure-python) and [func.HttpResponse](https://docs.microsoft.com/python/api/azure-functions/azure.functions.httpresponse?view=azure-python) references.
+- The body of `main` then does whatever it wants to process the request and generate a response. In this case, the function looks for a `name` parameter in the URL. Failing that, it checks if the request body contains JSON (using the `func.HttpRequest.get_json` function) and that the JSON contains a `name` value (using the `get` method of the JSON object returned by `get_json`).
```
sptramer: COPY (clarity): Remove `then does whatever it wants to`; Change `process` -> `processes`; Change `generate` -> `generates`

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+    ![Result of creating a new Python functions project](images/functions/project-create-results.png)
+
+1. When `__init__.py` opens, VS Code might tell you that you don't have a Python interpreter selected. In that case, use the **Python: Select Interpreter** command from the Command Palette and select the virtual environment in the local `.env` folder (which was created as part of the project).
+
+    ![Selecting the virtual environment created with the project](images/functions/select-venv-interpreter.png)
+
+> **Tip**: Whenever you want to create another function in the same project, use the **Create Function** command in the **Azure: Functions** explorer, or use the **Azure Functions: Create Function** command from the Command Palette. Both commands prompt you for a function name (which is the name of the endpoint), then creates a subfolder with the default files.
+>
+> ![New Function command in the Azure: Functions explorer](images/functions/function-create-new.png)
+
```
sptramer: Also requires that you manually install the `azure.functions` module (!) which is not `pip install`ed by the project creation phase. This can just be done directly in the terminal once the virtualenv is loaded:

```
pip install azure.functions
```

sptramer: Looking into this some more, as part of running the debug config later in the article: Even then it reports an error about not being able to import azure.functions:

image

Clicked 'continue debugging anyway' - and it worked (????).

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+    Content root path: d:\Examples\Python\AzureFunctions
+    Now listening on: http://0.0.0.0:7071
+    Application started. Press Ctrl+C to shut down.
+
+    Http Functions:
+
+            HttpExample: [GET,POST] http://localhost:7071/api/HttpExample
+    ```
+
+1. Ctrl+click the URL in the output window to open a browser to that address, or start a browser and paste in the same URL. In either case, you can see that the endpoint is `api/<function_name>`, in this case `api/HttpExample`. However, because that URL doesn't include a name parameter, the browser window should just show, "Please pass a name on the query string or in the request body" as appropriate for that path in the code.
```
sptramer: Platform: Is this a different gesture on macOS? Not familiar enough with vscode on that platform to know.

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+        --data {"name":"VS Code"} http://localhost:7071/api/HttpExample
+    ```
+
+    ```ps
+    # Windows (escaping on the quotes is necessary)
+    curl --header "Content-Type: application/json" --request POST \
+        --data {"""name""":"""VS Code"""} http://localhost:7071/api/HttpExample
+    ```
+
+    Alternately, create a file like `data.json` that contains `{"name":"VS Code"}` and use the command `curl --header "Content-Type: application/json" --request POST --data @data.json http://localhost:7071/api/HttpExample`.
```
sptramer: Remove? Not sure if this is beneficial.

docs/python/tutorial-azure-functions.md
sptramer: The file is `function.json` - needs to be fixed throughout

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+    ```
+    Http Functions:
+
+            DigitsOfPi: [GET] http://localhost:7071/api/DigitsOfPi
+
+            HttpExample: [GET,POST] http://localhost:7071/api/HttpExample
+    ```
+
+1. In a browser, or from curl, make a request to `http://localhost:7071/api/DigitsOfPi?digits=125` and observe the output. (You might notice that the code algorithm isn't entirely accurate, but we'll leave the improvements to you!) Stop the debugger when you're finished.
```
sptramer: COPY (clarity, style): Remove parenthetical

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+1. Redeploy the code by using the **Deploy to Function App** in the **Azure: Functions** explorer. When prompted, select the Function App created previously.
+
+1. Once deployment finishes (it takes a few minutes!), the output window shows the public endpoints with which you can repeat your tests.
```
sptramer: COPY (style): Remove parenthetical. If you want to call out that deployment might take some time, add a followup sentence.

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Clean up resources
+
+The Function App you created includes resources that can incur minimal costs (refer to [Functions Pricing](https://azure.microsoft.com/pricing/details/functions/)). To clean up the resources, right-click the Function App in the **Azure: Functions** explorer and select **Delete Function App**. You can also visit the [Azure portal](https://portal.azure.com), select **Resource groups** from the left-side navigation pane, select the resource group that was created in the process of this tutorial, and then use the **Delete resource group** command.
+
+## Next steps
+
+Congratulations on completing this walkthrough of deploying Python code to Azure Functions! You're now well positioned to create many more serverless functions.
```
sptramer: COPY (style?): Is this a common writing style for the VS Code docs? The tutorials on Azure tend to avoid using kudos like this, making direct appeals to the user, etc.

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+As noted earlier, you can learn more about the Functions extension by visiting its GitHub repository, [vscode-azurefunctions](https://github.com/Microsoft/vscode-azurefunctions). Issues and contributions are also welcome.
```
sptramer: COPY (passive, clarity): Remove `As noted earlier, you can` - replace with `To`

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+To learn more about Azure Functions, browse the [Azure Functions](https://docs.microsoft.com/azure/azure-functions/functions-create-first-function-python) documentation, and especially explore the different triggers you can use.
+
+To learn more about Azure services that you can use from Python, including data storage along with AI and Machine Learning services, visit [Azure Python Developer Center](https://docs.microsoft.com/python/azure/?view=azure-python).
+
+There are also other Azure extensions for VS Code that you may find helpful. Just search on "Azure" in the Extensions explorer:
```
sptramer: COPY (clarity): Change to `To find other VS Code extensions that help you work with Azure, search for 'Azure' in the Extensions explorer:`

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+![Azure extensions for VS Code](images/deploy-containers/azure-extensions.png)
+
+A few favorites include:
```
sptramer: COPY (style?): Maybe change to something like `Some popular extensions are...` to avoid a kind of value judgement here?

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+- [Cosmos DB](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-cosmosdb)
+- [Azure App Service](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice)
+- [Azure CLI Tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.azurecli)
+- [Azure Resource Manager (ARM) Tools](https://marketplace.visualstudio.com/items?itemName=msazurermtools.azurerm-vscode-tools)
+
+And again, if you've encountered any problems in the course of this tutorial, feel free to file an issue in the [VS Code docs repo](https://github.com/Microsoft/vscode-docs/issues).
```
sptramer: COPY (redundancy, passive): Remove `And again,` Change `you've` -> `you`

docs/python/tutorial-azure-functions.md
```diff
@@ -0,0 +1,348 @@
...
+
+Install the following software:
+
+- [Visual Studio Code](https://code.visualstudio.com/).
+- Python and the [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python) extension as described on [Python Tutorial - Prerequisites](/docs/python/python-tutorial.md).
+- [Azure Functions extension](https://marketplace.visualstudio.com/itemdetails?itemName=ms-azuretools.vscode-azurefunctions). For general information, visit the [vscode-azurefunctions GitHub repository](https://github.com/Microsoft/vscode-azurefunctions).
+
+### Install the Azure Functions core tools
+
+Follow the instructions for your operating system on [Work with Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local#v2) in the Azure documentation. The core tools themselves are written in .NET Core and Node.js, which is why it's presently necessary, believe it or not, to install .NET Core and Node.js to work with Azure Functions in Python! Fortunately, you need install these components only once, after which VS Code automatically prompts you to install any updates.
```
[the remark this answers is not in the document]
kraigb: With the last comment, "only" is in the correct place already, toward the end. I won't belabor the point other than to say it's something my MS Press editor beat into me through the process of about 2400 page of books. :)

docs/python/tutorial-azure-functions.md
kraigb: Yes, the VS Code docs are more casual, hence also the tongue-in-cheek remark earlier (which I need to remove anyway for other reasons).

docs/python/tutorial-azure-functions.md
kraigb: Again, the tone in VS Code is more casual, so I'm going to leave it as-is.

docs/python/tutorial-azure-functions.md
kraigb: I'll change "you've" but am leaving the And again, as it matches the other Azure tutorials and has a lighter tone.

docs/python/tutorial-azure-functions.md
kraigb: That works...I think the original language is an old carry-over from Node.js tutorials.

docs/python/tutorial-azure-functions.md
kraigb: Keeping this one for the casual VS Code tone.

docs/python/tutorial-azure-functions.md
kraigb: There are far too many occurrences of "function" that means "Azure Function" to use the proper name without being overburdensome. Instead, there are only a handful of places where I can distinguish Python functions specifically.

docs/python/tutorial-azure-functions.md
kraigb: For casual tone, leaving the opening bit.

docs/python/tutorial-azure-functions.md
kraigb: There's actually a keyboard encoding I need to use here. Thanks for catching it.

docs/python/tutorial-azure-functions.md
kraigb: I wanted to include that detail to complete the thought of using CURL, so if you create your own function that expects more complex JSON, you don't have to go figure out how to use a file instead of the inline data. It's minimal impact with a potentially large benefit for a decent number of people.

docs/python/tutorial-azure-functions.md
kraigb: With casual VS Code tone, I think it's appropriate because otherwise someone's going to pick on that particular detail. If we don't note if ourselves, people are left to ask whether we noticed at all. This way is better, then.

docs/python/tutorial-azure-functions.md
kraigb: Casual VS Code tone.

docs/python/tutorial-azure-functions.md
kraigb: I think this might be a bug somewhere. I never did an explicit pip install and things worked fine...

kraigb: Also, https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-python doesn't say anything about that, so I think the issue is in the language server and not in Functions.

sptramer:
LGTM, no longer seeing any obvious issues.

docs/python/tutorial-azure-functions.md
kraigb: Also, Rabeea didn't report this as an issue, so I'm going to leave it be for now.

PR #2585, opened 2019-04-18
updates to WSL  to fix shell issue

mikeblome:
@gregvanl I'd like to stage this internally only until PM has a chance to validate this fixes the problem.


gregvanl:
@mikeblome Merged and available on the staging server. Let me know when you'd like the changes published.

PR #2586, opened 2019-04-18
updates to linux install steps

mikeblome:
@gregvanl if these look good we can  go ahead and publish.

### DIFF
--- a/docs/cpp/config-wsl.md
+++ b/docs/cpp/config-wsl.md
@@ -49,20 +49,24 @@ To successfully complete this tutorial, you must do the following steps:
 
 1. Although you will be using VS Code to edit your code on Windows, you'll be compiling the code on Linux using the g++ compiler. You'll also debug on Linux using GDB. These tools are not installed by default on Ubuntu, so you have to install them. Fortunately, that task is quite easy!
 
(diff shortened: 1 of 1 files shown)


docs/cpp/config-wsl.md
```diff
@@ -49,20 +49,24 @@ To successfully complete this tutorial, you must do the following steps:
...
 
    ```bash
-   cd $HOME
-   cd ../../usr/bin
-   ls g++ gdb
+   sudo apt-get install build-essential gdb
    ```
+1. Verify that the install succeeded by navigating to the /user/bin directory and locating g++ and gdb. If the filenames are not returned from the following command, try running the update and upgrade command again.
 
+```bash
```
gregvanl: Can you indent this code block 3 spaces? Otherwise it won't be indented correctly under the step text.

gregvanl: I can do it to get this out today.

mikeblome: @gregvanl thanks I missed that. Done.

mikeblome: ok never mind i see you took care of it. thanks!

gregvanl:
Published

PR #2589, opened 2019-04-20
Correct base template path generated by snippet.

chrzhang:
Old

versus

New


kraigb:
@chrzhang Thanks for catching that! It was a detail I missed when I moved the template folder.

PR #2606, opened 2019-04-28
Update java-spring-boot.md

marcnewton:
Suggest link to install all recommended extensions from an extension package maintained by pivotal (https://marketplace.visualstudio.com/publishers/Pivotal).


hexiaokai:
Thank you for your PR and your interest in the VS Code documentation. While the referenced extension pack may be useful for many developers, it is probably too specific for general Spring Boot development.

PR #2607, opened 2019-04-28
Fix typo: change "pycodestyle" to "pydocstyle"

PeteH32:
The section on **pydocstyle** mistakenly says "pycodestyle" in one place.


gregvanl:
Thanks!

PR #2609, opened 2019-04-30
Remove ref to loading VSC from activated env

DonJayamanne:
Loading VSC from an activated Python Environment doesn't result in Python Extension using that environment.


DonJayamanne:
@kraigb /ccc

PR #2610, opened 2019-04-30
Edits to close a few issues

kraigb:


docs/python/debugging.md
```diff
@@ -181,17 +181,27 @@ As an example, say `${workspaceFolder}` contains a `py_code` folder containing `
 | `${workspaceFolder}/py_code` | `../data/salaries.csv` |
 | `${workspaceFolder}/data` | `salaries.csv` |
 
-### `debugOptions`
+### `redirectOutput`
 
-An array of additional options:
+When omitted or set to `True` (the default), causes the debugger to print all output from the program into the VS Code debug output window. If omitted, all program output is not displayed in the debugger output window.
```
gregvanl: @kraigb Is this second sentence correct? Should it be "If set to `False`, ..."

kraigb: Yes, you're correct. Can you make that tweak?

gregvanl: Will do.

PR #2620, opened 2019-05-03
Fix trivial typo.

DavidAnson:


gregvanl:
Every little bit helps.

PR #2624, opened 2019-05-04
Clarify x86_64 Linux for SSH

roblourens:
Microsoft/vscode-remote-release#126

Is this the best way to describe it, or x64, or Intel-based, etc?


gregvanl:
Published

PR #2626, opened 2019-05-05
Update Java documents with new updates from April

hexiaokai:


docs/java/java-debugging.md
```diff
@@ -154,21 +157,26 @@ Below are all the configurations available for `Launch` and `Attach`, for more i
 
 ### Launch
 
-- `mainClass` (required) - The fully qualified class name (for example [java module name/]com.xyz.MainApp) or the java file path of the program entry.
-- `args` - The command-line arguments passed to the program. Use `"${command:SpecifyProgramArgs}"` to prompt for program arguments. It accepts a string or an array of string.
+- `mainClass` (required) - The fully qualified class name (e.g. [java module name/]com.xyz.MainApp) or the java file path of the program entry.
```
gregvanl: It looks like you might not be current with the latest version since you are reverting editing changes.
For example:
we don't use e.g.
command-line arguments is correct (with a hyphen)
Java should be capitalized
The feature is CodeLens

docs/java/java-debugging.md
```diff
@@ -1,15 +1,15 @@
...
-PageTitle: Java Debug in Visual Studio Code
+PageTitle: Run and Debug Java in Visual Studio Code
 DateApproved: 1/2/2019
-MetaDescription: See how you can debug your Java code locally, and in the cloud.
+MetaDescription: See how you can run and debug your Java code locally, and in the cloud.
 MetaSocialImage:
 ---
 
-# Debugging Java in Visual Studio Code
+# Running and Debugging Java in Visual Studio Code
```
gregvanl: Are you sure this is neccesary? This will cause the topic title to wrap which doesn't look very nice.

docs/java/java-debugging.md
```diff
@@ -1,15 +1,15 @@
 ---
 Order: 3
 Area: java
-TOCTitle: Debugging
+TOCTitle: Running and Debugging
```
gregvanl: This will cause the TOC to wrap which doesn't look very nice. Are you sure you need to make this change?

docs/java/java-debugging.md
hexiaokai: I see. If that's too long, can we shorten it to run and debug? Lots of students don't really understand debug but they do want to run their Java code.

docs/java/java-debugging.md
hexiaokai: You're right, I just copied over from another source with new configurations we support now.

PR #2631, opened 2019-05-06
Update ssh.md

ahnm:
Add note that realpath needs to be installed on the remote machine.


roblourens:
Thanks, but we are removing the dependency on `realpath`, so we aren't going to take this.

PR #2633, opened 2019-05-07
Added security documentation for when you're extending docker-compose

AntonNguyen:
I also made a change in the examples here: https://github.com/microsoft/vscode-dev-containers/pull/36

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -512,6 +512,11 @@ version: '3'
 
       # Overrides default command so things don't shut down after the process ends.
       command: sleep infinity
+      # Allows vscode's debugger to connect to the container
+      cap_add:
+        - SYS_PTRACE
+      security_opt:
+        - seccomp:unconfined
 ```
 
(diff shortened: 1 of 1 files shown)


Chuxel:
This is good suggestion!  However, this is only required for C++, Go, and Rust which is why it's not included everywhere.  So, it's important to note that it is only required for certain languages. We probably should mention it in the section immediately before this one since it would also apply if you weren't extending.

docs/remote/containers.md
```diff
@@ -512,6 +512,11 @@ version: '3'
 
       # Overrides default command so things don't shut down after the process ends.
       command: sleep infinity
+      # Allows vscode's debugger to connect to the container
```
chrmarti: "Options required for debuggers based on ptrace."?

AntonNguyen:
> However, this is only required for C++, Go, and Rust which is why it's not included everywhere.

TIL! 

> We probably should mention it in the section immediately before this one since it would also apply

I can make this change.

PR #2638, opened 2019-05-08
Update troubleshooting.cmd - PATH

virgilwashere:
Changes:
- [x] consistent Windows username references
- [x] add path for System Installer version
- [x] add warning about quoting spaces


Chuxel:
LGTM!  Those are good edits!

PR #2639, opened 2019-05-08
Replace outdated git cheat sheet url

mbelsky:
Hey,
Current url returns:
```
Error
Failed to load PDF document.
```

There is a simple url update.


gregvanl:
Thanks!

PR #2664, opened 2019-05-09
Added a description how specify a custom remote machine ssh port.

TaQk:


roblourens:
This is not true. If you want to use a different port, you'll have to put it in your ssh config.

PR #2675, opened 2019-05-14
Update ssh.md

piotrgredowski:
Unified name of settings property indicating default extensions installed on SSH host


gregvanl:
Thanks!

PR #2676, opened 2019-05-14
Distinguish Ipython and Ipython Notebook

navneethc:
Changed IPython to IPython Notebook. I realise that it's a minor typo in terms of changes, but it's an important distinction to make as IPython is an on-going project on its own.


gregvanl:
Thanks for the clarification.

PR #2678, opened 2019-05-15
Update the install command for functions tools.

fiveisprime:
The tag for version 2 has been removed so this installation command is
failing for users coming to the tutorial.


gregvanl:
Published

PR #2679, opened 2019-05-15
Containers advanced topics and misc cleanup

Chuxel:
This PR:
1. Adds an "Advanced configuration" section in the main containers article that now includes:
    - Docker-in-Docker/Kubernetes (existed before)
    - Connecting to multiple containers at once (new)
    - Using SSH to connect to a remote Docker host (new)
2. Adds tips and tricks topics on:
   - Setting up a SSHFS mount. (Referenced in "Using SSH to connect to a remote Docker host")
   - Using rsync to move content to/from a remote host.  (Referenced in "Using SSH to connect to a remote Docker host")
   - Two more SSH tips given commonly raised issues.
   - Other cleanup

I could see the "Advanced configuration" section becoming a separate article or breaking up troubleshooting into separate articles per-extension that then includes advanced configuration.  However, I held off on making those changes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+* text=auto eol=lf
 *.gif filter=lfs diff=lfs merge=lfs -text
 *.mp4 filter=lfs diff=lfs merge=lfs -text
 *.jpg filter=lfs diff=lfs merge=lfs -text
--- a/docs/remote/containers-advanced.md
+++ b/docs/remote/containers-advanced.md
@@ -1,0 +1,413 @@
+---
+Order: 6
+Area: remote
+TOCTitle: Advanced Containers
+PageTitle: Advanced Container Configuration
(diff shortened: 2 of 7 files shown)


gregvanl:
@Chuxel That is a lot of new content and it looks like we should created an Advanced Containers topic before this goes in. It will help with the headings and the right nav as you can avoid those H4s which are now bold. Do you want to create the new topic with this PR?

Chuxel:
@gregvanl Yep - Can do.  Yeah I was pretty nervous about the size which is why I'd brought up the refactoring.

Chuxel:
@gregvanl Okay - added the new topic with the new content and moved a few things out of tips and tricks that seemed to fit under the advanced configuration topic.

gregvanl:
Published after a quick edit pass

PR #2689, opened 2019-05-21
Document disabling extensions when using npm test

TomasHubelbauer:
Googling *vs code extension npm test disable other extensions* takes one to this document, which is only about launching VS Code debugger configurations, but it seems to me like this information should be here as well for that reason.

I wanted to keep this short but felt like I should mention `cross-env` since that's what people will likely want anyway in case they develop on Windows but have CI run their extension tests on Linux.

### DIFF
--- a/api/working-with-extensions/testing-extension.md
+++ b/api/working-with-extensions/testing-extension.md
@@ -82,6 +82,17 @@ By default, the debug instance of VS Code will load any extension you've previou
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @TomasHubelbauer I think your PR will no longer be applicable with test runner changes coming in the next release 1.35. Have a look at the vnext documentation at https://github.com/microsoft/vscode-docs/blob/vnext/api/working-with-extensions/testing-extension.md and see how your suggestion would work.

TomasHubelbauer:
Cool! This will also make it possible to pass --inspect to the extension host which is something I had a use case for before and had to hack around a little! Thanks

PR #2694, opened 2019-05-23
Documentation to make it easier to connect to Hyper-V hosts

jovinbm:
Hyper-V tends to assign new IP addresses when the computer is restarted or is woken up from sleep. This details how to use `mshome.net` to make it easier to connect to Hyper-V hosts.


roblourens:
Thanks for the PR. This isn't really a Hyper-V specific tip because most machines will get assigned a new IP address when they start up and connect to the network unless they have a static IP. So this boils down to just "use a domain name instead of an IP address if they have one". So I don't really want to take this because I don't really think it adds that much.

jovinbm:
Okay, thanks. I'm closing this one.

PR #2699, opened 2019-05-24
Location of wsgi.ini file is incorrect

jainr:


kraigb:
Thanks for taking the time to fix this detail!

PR #2704, opened 2019-05-27
Fixes #2702

### LINKED ISSUE #2702: Uneccesary <p></p> tags in Django Tutorial
I was looking through the Django Tutorial as I am currently thinking of switching to using VS Code instead of Pycharm, and I noticed in [this section](https://github.com/microsoft/vscode-docs/blob/63214cd2616f9edca20adefe4593859617720b78/docs/python/tutorial-django.md#use-the-code-snippet-to-add-pages) that it instructs users to enter `<p>Home page for the Visual Studio Code Django tutorial.</p>` in the `{% block title %}` section, but the `<p><\p>` tags are not necessary because the `{% block title %}` section is already included inside of the `<title></title>` tags in the template. Doing as instructed, the `<p></p>` tags are rendered in the tab header as plain text.

endormi:
I fixed Flask's issues as well.

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -582,11 +582,11 @@ With the code snippet in place, you can quickly create templates for the Home, A
 
     ![Django tutorial: insertion of the djextlayout code snippet](images/django-tutorial/code-snippet-inserted.png)
 
-1. At the insertion point in the "title" block, write `Home`, and in the "content" block, write `<p>Home page for the Visual Studio Code Django tutorial.</p>`, then save the file. These lines are the only unique parts of the extended page template:
+1. At the insertion point in the "title" block, write `Home`, and in the "content" block, write `Home page for the Visual Studio Code Django tutorial.`, then save the file. These lines are the only unique parts of the extended page template:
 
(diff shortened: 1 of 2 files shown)


endormi:
Hmmm, is there a reason why you haven't merged my PR? @gregvanl jus wondering, since it is a fix.

gregvanl:
Hi @endormi The python tutorials are owned by another writer but it looks like @JoshuaPartlow is currently out. I'll try to have a look at your PR shortly. Sorry for the delay.

endormi:
Great, thanks!

gregvanl:
Hi @endormi, I had a look at issue #2702 and the Python Django tutorial and I think the issue creator was confused and put the paragraph into the title block rather than the content block. I'm not sure this PR is necessary. While the <p> aren't strictly necessary, they do force a new line before displaying the content.

endormi:
@gregvanl I can do a quick fix, adding the tags inside {% block content %]

gregvanl:
Also looking quickly at Django documentation, it looks like the convention is to use HTML tags within elements inside content blocks.

endormi:
Okay, that's fine 👍

gregvanl:
Thanks again for taking the time to submit a PR!

endormi:
@gregvanl No problem! I'll be looking for possible fixes.

PR #2707, opened 2019-05-28
Update Windows.md

Mohit-Nain:
There is a typo in line 28, privilege should be replaced with privileges, have made that change. Please review.


gregvanl:
Thanks

PR #2720, opened 2019-05-30
docs: variables-reference.md include settings.json

janosh:
-  make it clear that besides `launch.json` and `tasks.json`, variable substitution is also available in `settings.json`

### DIFF
--- a/docs/editor/variables-reference.md
+++ b/docs/editor/variables-reference.md
@@ -9,7 +9,7 @@ MetaDescription: Visual Studio Code variable substitution reference
 ---
 # Variables Reference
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @janosh Thank you for your interest in the VS Code documentation and taking the time to submit a PR. The variables enumerated in the Variables Reference topic are not available for settings.json. There are some variables for some settings, such as for windows.title, but these are setting specific and not the general launch.json and task.json shared variables.

janosh:
That's weird. I used `${workspaceFolder}` in `settings.json` as described in [this comment](https://github.com/microsoft/vscode-python/issues/5840#issuecomment-497321419) and it's definitely working. This information would have been valuable so even if only some variables are supported in `settings.json` I think it would be worth listing those.

gregvanl:
In the example you show, ${workspaceFolder} is passed as a string to the Python extension to evaluate and the extension knows how to expand ${workspaceFolder}. The variable is not expanded by VS Code through its own settings. The Variable Reference topic shows which variables will be evaluated by the VS Code debugger and task features.

janosh:
Oh I see, sorry for the confusion.

PR #2721, opened 2019-05-30
Linux prerequistes

Chuxel:
This is an initial pass at being a bit more explicit about our Linux requirements and what is/is not known to work.  //cc: @kieferrm @roblourens - LMK if the info looks accurate or if you want to include / exclude something.


gregvanl:
Published after a few updates.

PR #2726, opened 2019-06-03
VS Code Insiders => VS Code Stable

Chuxel:
Queuing up doc tweaks for when we are ready.


gregvanl:
Hi @Chuxel If you move this PR to target the vnext branch, we can merge now and it will go to master along with the 1.35 release notes and updates when we ship 1.35. Any reason these updates should be a PR to master?

Chuxel:
@gregvanl Ahh good point!

PR #2728, opened 2019-06-04
Update userinterface.md

Mohit-Nain:
Correcting the verb form to be used at Line79


gregvanl:
Hi @Mohit-Nain I think the current text reads better.

PR #2742, opened 2019-06-10
Update settings.md

Mohit-Nain:
Modified Line23 to "Configuring user and workspace settings" , configure is the right verb while working with settings. Creating user and workspace settings does not make correct sense.


gregvanl:
Hi @Mohit-Nain Your suggestion is valid, but I don't think it warrants changing the H3 which acts as a link to the section. This would break anyone who has created a link to that section

PR #2744, opened 2019-06-10
Update settings.md

Mohit-Nain:
Corrected line90 to use example, as we have provided single example with two different settings.


gregvanl:
Hi @Mohit-Nain, not sure this is worth changing since there are two setting examples. The verb form would need to be changed as well.

Issue #2787, opened 2019-06-22
[Python] Update arguments to run parallel tests with pytest-xdist


luabud:
Having ["-n", "4"]  in pytestArgs not working properly in some cases due to a bug. Current workaround is to change ["-n", "4"] to ["-n 4"]. Can we make this minor change in the docs? 

Related: https://github.com/microsoft/vscode-docs/issues/2551

gregvanl:
@luabud PRs welcome. @JoshuaPartlow can help you fork vscode-docs and submit a PR.
https://help.github.com/en/articles/fork-a-repo

gregvanl:
Fixed with https://github.com/microsoft/vscode-docs/commit/2d5f49cbe6c63b7ec2cb8bada828a58109130fe8

PR #2804, opened 2019-06-24
[Python] Update args for running tests in parallel with pytest-xdist

luabud:
For [#2787](https://github.com/microsoft/vscode-docs/issues/2787)


gregvanl:
Thanks!

PR #2834, opened 2019-06-29
key presses

Mohit-Nain:
line 192 has statements which reads like "The widget listens for key presses", it should either be "listens to " or "looks for ". Listens for key presses does not make sense.


gregvanl:
Current text is fine and matches how developers commonly talk about event listeners.

PR #2835, opened 2019-06-29
when clause

Mohit-Nain:
Re-framed line 219 to make better sense, previously it read like: 
VS Code gives you fine control over when your key bindings are enabled through the optional `when` clause.

when looks a bit misplaced here, hence re-framing the line to make better sense and easier to understand


gregvanl:
Current text is fine. We are describing how VS Code allows you to control when keyboard shortcuts are enabled.

PR #2837, opened 2019-06-29
Line 590

Mohit-Nain:
correcting the typo in line 590, it should be end to end node.js scenarios.


gregvanl:
I think it is accurate to describe the tutorial as a single scenario.

PR #2840, opened 2019-06-29
Fix a typo.

Maviza101:
This PR fixes what I believe is a typographical error.


gregvanl:
Hi @Maviza101 Thank you for your interest in the VS Code documentation and taking the time to submit a PR. The current text is correct, both the VS Code product and it's documentation are built using VS Code at the team's editor for JavaScript, TypeScript, and JSON for the product and Markdown for the documentation.

PR #2841, opened 2019-06-29
Link headers to their target docs section

Maviza101:
This PR links some headers in the docs for the VS Code Extensions API to the actual sections of the docs that those headers are talking about.


gregvanl:
Thank you for the suggestion. Using section links was brought up earlier and we opted to keep readers on the introductory topic until they had finished. Readers can use the left side TOC for navigation.

PR #2848, opened 2019-07-02
Zoom Level

Mohit-Nain:
Along with incremental knowledge of Zoom levels , we should also mention decrement values as well.


gregvanl:
Thanks

PR #2849, opened 2019-07-02
Added links for 64bit windows

NiteshKoushik:
Added VS code downloadable links for 64 bit windows


gregvanl:
@NiteshKoushik Thank you for submitting a PR but since 64-bit is the default so I don't think we need to call this out specifically.

PR #2850, opened 2019-07-02
Added link for Jade template engine

NiteshKoushik:
To know more on Jade template engine and API helps to render jade using the JavaScript API in node.js


gregvanl:
Jade has been renamed to pug so I'm reluctant to add this link. We may want to remove that mention of Jade.

PR #2851, opened 2019-07-02
Added other language features

NiteshKoushik:
Added other language features for knowing different types of supportable language features


gregvanl:
Hi @NiteshKoushik I don't think sending a regular VS Code user to the API documentation will be helpful to learn about other language features. I think they will be lost.

PR #2860, opened 2019-07-04
Recommend to always use strict host key checking

terroirman:
I've searched for a long time before finding the cause of my timout at "Waiting for ssh tunnel to be ready". So I thought I'd share here. Not sure if this is a bug or something the users should be aware of. It would be nice if vscode intercept that use case where the  target host key has changed (ie.: "Port forwarding is disabled to avoid man-in-the-middle attacks." warning). Let me know if a enhancement request would be a better way to go here.
Also, as an addition to what I put here, it would be nice to provide an external guide on how to properly add trusted host key. I didn't include one as I don't know if :
1- you already have one somewhere that I'm not aware of
2- what is your policy for external link in the documentation

Thanks for considering this!

### DIFF
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -134,6 +134,24 @@ If you do see that message, follow these steps to update your SSH server's [sshd
 3. Restart the SSH server (on Ubuntu, run `sudo systemctl restart sshd`).
 4. Retry.
 
+**Always use strict host key checking**
+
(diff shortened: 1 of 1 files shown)


roblourens:
Hey @terroirman thanks for pointing this case out and for the PR. But I could actually just handle this in the extension - if we detect that message, we can stop connecting and show an error message. I would rather do that than add a section to the docs. If you file an issue at https://github.com/Microsoft/vscode-remote-release/issues I can handle it this month.

PR #2864, opened 2019-07-06
Update and clarify snippet files part of the snippet user guide

### LINKED ISSUE #2450: Not mentioned project-level snippets
https://code.visualstudio.com/docs/editor/userdefinedsnippets

https://github.com/Microsoft/vscode/issues/8102#issuecomment-423476360

### LINKED ISSUE #2855: Missing documentation for the "snippet prefix arrays" feature
Snippet prefix arrays were requested in issue microsoft/vscode#42838, implemented in pull request microsoft/vscode#59092, and released in version [microsoft/vscode/tags/1.28.0](https://github.com/microsoft/vscode/releases/tag/1.28.0). However, the documentation on creating user-defined snippets ("[Creating your own snippets](https://code.visualstudio.com/docs/editor/userdefinedsnippets)") does not state, demonstrate, or otherwise reference snippet prefix arrays.

travisclagrone:
High-level summary of changes:

- Fix #2855 "Missing documentation for the 'snippet prefix arrays' feature"
- Fix #2450 "Not mentioned project level snippets"
- Standardize document title and first couple headers for clarity's sake
- Split--and then lightly extend--the subsection "Creating your own snippets" to improve SoC/SRP per section.
- Separate, reflow, and add detail to the cursory snippets files documentation with a new section that better addresses the language-vs-global and user-vs-project -scoping of snippets files.
- Add a new section "Snippet schema" that explicitly and declaratively provides the structure of a snippets file and snippet definition. (this was not something that existed before, outside of implicit and only-cursorily-documented VS Code intellisense)

### DIFF
--- a/docs/editor/userdefinedsnippets.md
+++ b/docs/editor/userdefinedsnippets.md
@@ -29,61 +29,79 @@ Below are some popular extensions which include snippets in their language suppo
 
 ## Create your own snippets
 
-You can define your own snippets, either global snippets or snippets for a specific language. To open up a snippet file for editing, select **User Snippets** under **File** > **Preferences** (**Code** > **Preferences** on macOS) and select the language (by [language identifier](/docs/languages/identifiers.md)) for which the snippets should appear or create a new global snippet (**New Global Snippets file**).
+You can easily define your own snippets without any extension. To create or edit your own snippets, select **User Snippets** under **File** > **Preferences** (**Code** > **Preferences** on macOS), and then select the language (by [language identifier]) for which the snippets should appear, or the **New Global Snippets file** option if they should appear for all languages. VS Code manages the creation and refreshing of the underlying snippets file(s) for you.
 
 ![snippet drop-down](images/userdefinedsnippets/snippet-dropdown.png)
 
-Snippets are defined in a JSON format. The example below is a `For Loop` snippet you would use for JavaScript:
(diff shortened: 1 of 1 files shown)


gregvanl:
@travisclagrone Great stuff!
I'm not sure users need the schema sections and there are a couple of corrections needed (for example, the global snippets file does not need to be named `global.code-snippets`. I'll create a follow up commit and link it here so you can have a look after I'm done.
Thanks again!

travisclagrone:
@gregvanl Awesome! And thanks for the correction. I would argue for the schema section though (at least the TypeScript-expressed snippet definition, anyway), as it makes plainly clear the fact that each of `prefix` and `body` can be either a single string or an array of strings, and what it means in each case.

gregvanl:
Hi @travisclagrone Have a look at https://github.com/microsoft/vscode-docs/commit/d3ee87557e7b934c615bc17ef740b90791aeba1a.
I needed to revert the reference link definition since that doesn't work with our website page generation and the `kb(commanded)` trick only works if there is a default keybinding (otherwise shows as 'unassigned'). 
I think including a TS schema would be too confusing for some readers and your text describing the `prefix` and `body` does a nice job mentioning one or more strings can be supplied.
I also tried to simplify the text as we try to keep content brief and straightforward as many of our readers have English as a second language.
Feel free to let me know if I went too far.
Really appreciate you taking the time to give this topic a refresh!

PR #2866, opened 2019-07-08
Corrected Line 19

Mohit-Nain:
Or should be used in place of And, as we are informing about any language that get transpiled in to Java Script. And is used when we are restricting the options to selected few, here we are giving options of any language.


gregvanl:
Thanks

PR #2867, opened 2019-07-08
Removed All from Line 33

Mohit-Nain:
All in line 33 does not make any sense, removed it for better understanding


gregvanl:
Thanks

PR #2868, opened 2019-07-08
Verb correction in Line 97

Mohit-Nain:
used correct verb form in line 97, where Opened browser tab is correct rather than open browser tab


gregvanl:
Current text reads fine.

Mohit-Nain:
I think this might need a revisit, currently open browser is incorrect as per me, it should always be currently opened browser. 
Request you to please rethink on this . @gregvanl

PR #2869, opened 2019-07-08
Preposition missing in Line 99

Mohit-Nain:
Added "to" in Line 99 to make it accurate as per standards. 
The best way to explain <something> is to think of a <way>


gregvanl:
Thanks

PR #2870, opened 2019-07-08
Correcting Line 276

Mohit-Nain:
We are giving option to use User setting across the workspace, therefore "can" makes better sense than "will"


gregvanl:
Current text is correct. The setting will be shared across workspaces without user interaction.

PR #2874, opened 2019-07-08
Add configuration example to keep trailing whitespace

rowild:
[Suggestion]
Since the question how to keep trailing whitespace in markdown files (to create simple line breaks) pops up at various different corners, it might be helpful to provide a configuration example in the documentation.


gregvanl:
Thanks! I've updated the text slightly since VS Code by default won't remove trailing whitespace but as you point out, it is not uncommon for user or workspace settings to be configured to remove trailing whitespace.

rowild:
Thank you very much!!

PR #2882, opened 2019-07-11
Revise instructions on copying SSH keys into the container

Chuxel:
While testing, I realized the instructions for this did not work on Windows due to Docker limitations.  I instead went for a less elegant, but consistently working approach instead.

Ultimately, we need a feature to streamline what I describe.

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -383,8 +383,6 @@ However, in some cases you may be cloning your repository using SSH keys instead
     "postCreateCommand": "mkdir -p ~/.ssh && cp -r ~/.ssh-localhost/* ~/.ssh && chmod 700 ~/.ssh && chmod 600 ~/.ssh/*"
     ```
 
-You can modify the mount
-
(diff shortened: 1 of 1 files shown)


docs/remote/containers.md
```diff
@@ -353,27 +353,40 @@ git config --global user.email "your.email@address"
...
+          - ~/.ssh:~/.ssh-localhost:ro
     ```
 
+    And then add this to `devcontainer.json` to copy the keys to the correct spot with the right permissions:
+
+    ```json
+    "postCreateCommand": "mkdir -p ~/.ssh && cp -r ~/.ssh-localhost/* ~/.ssh && chmod 700 ~/.ssh && chmod 600 ~/.ssh/*"
+    ```
+
+You can modify the mount
```
gregvanl: @Chuxel Is there more content to go here or can this sentence fragment be removed?

Chuxel: Whoops!  Good catch.  Fixed.

PR #2888, opened 2019-07-15
Update config-clang-mac.md

nico:
clang is at `/usr/bin/clang`, not at `usr/bin/clang/`. And since this is about macOS, there's no WSL.


mikeblome:
@nico thanks! I'm hoping to publish an update to this topic very soon which will address the WSL point.

PR #2891, opened 2019-07-16
Line 24 issue

Mohit-Nain:
In line 24, around and between are used together while referring to movement between different lines of the same file. ideally the line should read as "If you are jumping between different lines of the same file". 
Using Around and between together is in correct way of writing.


gregvanl:
The current text is more conversational which is the tone we aim for.

PR #2892, opened 2019-07-17
Update Format on save information

Edulopez:
Update Go.FormatOnSave example to the current version with its current settings.


gregvanl:
Thanks!

PR #2897, opened 2019-07-18
Typo Correction

Mohit-Nain:
Line 16 has typo , which has been fixed.


gregvanl:
The current text is correct. There are two kinds of expansions, for a snippet or an abbreviation.

PR #2898, opened 2019-07-18
Line 103 update

Mohit-Nain:
In line 103, we are referring to CSS properties where the preference has been enabled, hence those should be used instead of these, as we are referring to CSS properties having preference.


gregvanl:
Thanks

PR #2900, opened 2019-07-18
Correct helping verb for Line 378

Mohit-Nain:
Cannot use does when referring to plural snippets, either use snippet or use do not.


gregvanl:
Thanks

PR #2909, opened 2019-07-24
.venv activation and PowerShell execution policy

CodeRunRepeat:
Setting up the PowerShell execution policy to allow activation of a virtual environment and install packages


kraigb:
Thanks for adding that detail.

PR #2910, opened 2019-07-24
Modify hello_there in views.py same as app.py

kensel:
Modify the hello_there function in views.py to have the name parameter default to None as seen in app.py.  This restores the message returrned when navigating to /hello


kraigb:
Thanks, good fix.

PR #2915, opened 2019-07-26
Update bundling-extension.md

hiepxanh:
The instruction about lauch.json is not detail. That instruction cause F5 debugging with the launch.json is cannot lauch. 
My update suggest the right way that user can debug after upgrading VSCE to webpack.
Base on this [instruction by John Papa](https://johnpapa.net/is-your-vs-code-extension-slow-heres-how-to-speed-it-up/)

### DIFF
--- a/api/working-with-extensions/bundling-extension.md
+++ b/api/working-with-extensions/bundling-extension.md
@@ -90,7 +90,7 @@ Merge these entries into the `scripts` section in `package.json`:
   "vscode:prepublish": "webpack --mode production",
   "compile": "webpack --mode none",
   "watch": "webpack --mode none --watch",
-  "test-compile": "tsc -p ./",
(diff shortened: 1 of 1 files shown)


jrieken:
Sorry, I don't understand these changes. Can explain what's not working in the sample and how you want to fix it?

hiepxanh:
@jrieken old lauch.json preLauchTask point to: `"compile": "webpack --mode none",` which only build webpack but not open up the vscode Extension Development Host IDE, I could take a video record if you need

hiepxanh:
Agr... After lastest update visual studio code, I can see that it work normally, sorry for this wrong PR. I can't really get understand why, so sad

Raylin51:
I have the same question with you. It seems like the "--watch" option cause this problem. Vscode can't catch complied status correctly. The "compile" script don't have "--watch" option, so it work normally.

PR #2921, opened 2019-07-29
Minor fixes to CI page

letmaik:
The service is called Travis CI, not just Travis. That's more or less what this PR fixes.


gregvanl:
Thanks!

PR #2923, opened 2019-07-31
Correcting Line 112

Mohit-Nain:
Line 112 talks about updating the extension , therefore we need to use extension's auto-update, rather than extensions auto-update. 
In case we want to inform about multiple extensions, it has to be extensions' auto-update.


gregvanl:
Extensions auto-update is the name of the feature

PR #2924, opened 2019-07-31
Re-writing line 160

Mohit-Nain:
Section of line 160 has been re-written to make it a better read and also corrected one typo.


gregvanl:
Thanks

PR #2925, opened 2019-07-31
Line 206 re-written

Mohit-Nain:
Re-written line 206 to make it easier to understand , also included the line break where it was necessary


gregvanl:
Current text is fine

PR #2927, opened 2019-08-01
Update extension pack post

burkeholland:
1. Change `extensionDependencies` to `extensionPack`. 
2. Update Wade's Twitter to the Code twitter


burkeholland:
The `extensionDependencies` attribute prohibits the disabling of any extension that's been installed with an extension pack. The correct attribute is `extensionPack`.

Also updated the Twitter reference to the general Code account since Wade is no longer on the team.

gregvanl:
Thanks!

PR #2930, opened 2019-08-02
Line 149 change

Mohit-Nain:
Preposition correction in Line 149, with focus always on should be used.


gregvanl:
Developers usually talk about focus being in a UI element rather than on.

PR #2932, opened 2019-08-02
Line 251 to use powerful

Mohit-Nain:
powerful user tips for the terminal makes better sense


gregvanl:
A "power user" is a common term for an advanced software user.

PR #2934, opened 2019-08-02
Fix one note in "Variables Reference"; add another

SamB:
If this is too much detail, I guess you could just take the whole note out; however, suggesting that `${env:Path}` is the correct casing is probably the worst possible thing you could do (you guys are pretty smart, so there aren't that many other mistakes for you to make here).

You know, in theory. If there were actually a reason to use `${env:PATH}` for anything, where it didn't matter that the syntax is different on different platforms.

Maybe this portability stuff would go better somewhere else in the manual?

### DIFF
--- a/docs/editor/variables-reference.md
+++ b/docs/editor/variables-reference.md
@@ -72,7 +72,11 @@ You can also reference environment variables through the **${env:Name}** syntax 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @SamB Thanks for pointing out that the note about casing is unnecessary. We will remove it.

PR #2936, opened 2019-08-05
Line 26 re-arrangement

Mohit-Nain:
It would be better to re-arrange line  26 in following manner:
If you continue typing characters, the list of members is filtered to only include members containing the typed characters. 
One other option that can be used goes like this:
If you continue typing characters, the list of members is filtered to include only the members containing the typed characters.


gregvanl:
Thanks

PR #2938, opened 2019-08-05
Line 149 update

Mohit-Nain:
this behavior would make a better sense than the behavior, since we are referring to the behavior which has been introduced in the early part of the sentense


gregvanl:
"the behavior" of the widget in general,

PR #2946, opened 2019-08-06
Update bundling-extension.md

Raylin51:
Do not use "--watch" option in watch script. Vscode can not catch compiled status correctly. It will cause the dev environment can't wake up.

### DIFF
--- a/api/working-with-extensions/bundling-extension.md
+++ b/api/working-with-extensions/bundling-extension.md
@@ -89,7 +89,7 @@ Merge these entries into the `scripts` section in `package.json`:
 "scripts": {
   "vscode:prepublish": "webpack --mode production",
   "compile": "webpack --mode none",
-  "watch": "webpack --mode none --watch",
+  "watch": "webpack --mode development",
   "test-compile": "tsc -p ./",
 },
 ```


jrieken:
Sorry, but I don't understand what this is for. With this change `watch` isn't watching anymore, use the compile task then. Also, using none as mode as proven to produce the best source maps.

Raylin51:
@jrieken The default prelaunch task config is "npm: watch". So when we press F5 to start debug mode, it will run `npm run watch`. But if you add `--watch` option in watch script, the VS Code Extension Development Host IDE will not open up unless watch task is interrupted. It seems like VS Code can't catch webpack compile status correctly. According to default config, watch script is used to develop. So development mode should be used in watch script. Thanks.

jrieken:
Please use this as reference: https://github.com/Microsoft/vscode-extension-samples/blob/5fb626a4e383c41b8878f94045cff987b92b6bca/webpack-sample/.vscode/launch.json#L16

The launch-config doesn't trigger the watch-version but the normal webpack version. What I have realised tho is that the script in the doc was out of sync with the script (names) in the sample repo. I have pushed a change to align those names.

Nothing else left to do

Raylin51:
> Please use this as reference: https://github.com/Microsoft/vscode-extension-samples/blob/5fb626a4e383c41b8878f94045cff987b92b6bca/webpack-sample/.vscode/launch.json#L16
> 
> The launch-config doesn't trigger the watch-version but the normal webpack version. What I have realised tho is that the script in the doc was out of sync with the script (names) in the sample repo. I have pushed a change to align those names.
> 
> Nothing else left to do

But the question is this demo's prelaunch task is also without `--watch` option. https://github.com/microsoft/vscode-extension-samples/blob/5fb626a4e383c41b8878f94045cff987b92b6bca/webpack-sample/package.json#L27

I am very sure about the webpack's watch option shouldn't be used in prelaunch task based on current version VS Code.

octref:
> But if you add --watch option in watch script, the VS Code Extension Development Host IDE will not open up unless watch task is interrupted.

Are you sure? AFAIK first launch would do prelaunch + launch, but if prelaunch task is watching, subsequent launches will be immediate.

Raylin51:
> > But if you add --watch option in watch script, the VS Code Extension Development Host IDE will not open up unless watch task is interrupted.
> 
> Are you sure? AFAIK first launch would do prelaunch + launch, but if prelaunch task is watching, subsequent launches will be immediate.

I'm quite sure. Here is the same problem. https://github.com/microsoft/vscode-docs/pull/2915.
I think it is a bug. But I have no time to fix it. So I just fix document. My system version is MacOS Mojave 10.14.6. I don't know whether this problem will happen on Windows. But in MacOS when I use watch option and start debug, VS Code will always pending and never launch.

PR #2948, opened 2019-08-07
Correction in line 135

Mohit-Nain:
Typo correction in line 135


gregvanl:
I think "OPEN EDITORS and Quick Open lists include" will read better. I'll update it.

PR #2951, opened 2019-08-08
Line 14 Update

Mohit-Nain:
Avoid when referring to technical things reads better than Prevent, hence making that change


gregvanl:
Current text is correct. Some corporate environments don't let employees install extensions.

PR #2952, opened 2019-08-08
Line 35 Change

Mohit-Nain:
"Newly extracted version" reads much better than "newer extracted version"


gregvanl:
Current text is correct. The extracted version is newer.

PR #2953, opened 2019-08-09
📦 NEW: Background color of activated input fields

ahmadawais:
Documented the new color from version [1.37 New theme color for active input options](https://code.visualstudio.com/updates/v1_37#_new-theme-color-for-active-input-options)

Peace! ✌️


gregvanl:
Thanks!

PR #2955, opened 2019-08-11
Correct a spelling error (Changed "cra" to "crA")

sobiodarlington:
Changed "cra" to "crA"

> **Tip:** The suggestions widget supports CamelCase filtering, meaning you can type the letters which are upper cased in a method name to limit the suggestions. For example, "cra" will quickly bring up "createApplication".

> **Tip:** The suggestions widget supports CamelCase filtering, meaning you can type the letters which are upper cased in a method name to limit the suggestions. For example, "crA" will quickly bring up "createApplication".


gregvanl:
Hi @sobiodarlington Thank you for taking the time to submit a PR. Users don't need to match case for CamelCase filtering to work.

image

PR #2958, opened 2019-08-12
Line 88 Re-phrased

Mohit-Nain:
Re-written line 88 to make is easier to understand and read


gregvanl:
Current text is correct. It could use a comma but there is no conditional in the previous sentence.

PR #2960, opened 2019-08-12
Re-Arrangement of line 137

Mohit-Nain:
Line 137 has been modified to introduce line break and merge one sentence to help user understand it better


gregvanl:
Current text is fine.

PR #2962, opened 2019-08-12
Removing duplicate text in line 220

Mohit-Nain:
For example was duplicated in line 220, the same has been removed


gregvanl:
Thanks

PR #2978, opened 2019-08-14
know -> learn

davidthewatson:
While the usage of "know" here is not technically incorrect, it is unusual. The more idiomatic usage is "learn".


gregvanl:
Thanks!

PR #2979, opened 2019-08-15
Update python-tutorial.md

rich-TIA:
Added note to select the correct interpreter when running in an Anaconda environment. Only tested on VSCode 1.37.0 on MacOSX High Sierra.


kraigb:
Thanks, that's helpful.

PR #2988, opened 2019-08-16
Clarifications on limitations of mounting .vscode-server, WSL tip clarification

Chuxel:
Partial resolution to https://github.com/microsoft/vscode-remote-release/issues/1153

Also drops mention of adding Windows system path into WSL path in the troubleshooting article as this is no longer required.


docs/remote/containers-advanced.md
```diff
@@ -151,48 +151,65 @@ If you've already built the container and connected to it, run **Remote-Containe
...
+* When you rebuild the container:
+  * New extensions added to devcontainer.json will **not** be automatically installed.
+  * Any `postCreateCommand` in `devcontainer.json` will **not run**.
+* Deleting the container will not automatically delete the named volume.
 
-**Dockerfile or image**:
+To create the named local volume, follow these steps:
 
-Replace `your-volume-name-goes-here` with a unique volume name for the container in `devcontainer.json` as follows:
+1. **If you are using a non-root user**, you'll need to ensure your Dockerfile creates `~/.vscode-server` and/or `~/.vscode-server-insiders` in the container.  If you do not do this, the folder will be owned by root and your connection will fail with a permissions issue. See [Adding a non-root user to your dev container](#adding-a-non-root-user-to-your-dev-container) for full details, but you can use this snippet in your Dockerfile to create the folders for a `vscode` user:
```
chrmarti: Maybe refer to "the non-root user" instead of "a `vscode` user", otherwise users might think that naming a user `vscode` makes it somehow special?

docs/remote/containers-advanced.md
```diff
@@ -151,48 +151,65 @@ If you've already built the container and connected to it, run **Remote-Containe
...
 
-```json
-"runArgs": ["-v","your-volume-name-goes-here:/root/.vscode-server"]
-```
+    ```Dockerfile
+    ARG USERNAME=vscode
+    RUN mkdir -p /home/$USERNAME/.vscode-server /home/$USERNAME/.vscode-server-insiders \
```
chrmarti: Would:
```
USER someuser
RUN mkdir -p ~/.vscode-server ...
```
work and make it easier to apply?

docs/remote/containers-advanced.md
```diff
@@ -151,48 +151,65 @@ If you've already built the container and connected to it, run **Remote-Containe
...
 
-**Docker Compose**:
+    Add the following to `devcontainer.json`, replacing `/root` with the home directory of in the container if not root (e.g. `/home/vscode`) and `unique-vol-name-here` with a unique name for the volume:
 
-Update (or [extend](/docs/remote/containers.md#extending-your-docker-compose-file-for-development)) your `docker-compose.yml` with the following. Replace `your-volume-name-goes-here` with a unique volume name for the container.
+    ```json
+    "runArgs": [
+        "-v", "unique-vol-name-here:/root/.vscode-server",
+        // And/or for VS Code Insiders
+        "-v", "unique-vol-name-here-insiders:/.vscode-server-insiders",
```
chrmarti: Would it be easier to only show the code for stable and add a note in the text that the same applies for the insiders folder?

docs/remote/containers-advanced.md
Chuxel: That would cause the commands that follow to run as the user specified.  If you've opted not to install `sudo` for security reasons, switching to the other user would be the last step.  This is also a partial of the full code that creates the user.

docs/remote/containers-advanced.md
Chuxel: I had it that way, but then you need to say "also insiders" in each spot where this is needed, which actually ended up taking up more space and was easy to miss. This was a bit more concise and allows you to use both versions - cut and paste also covers both use cases so people don't have to think as much.

docs/remote/containers-advanced.md
Chuxel: Yeah I just noticed a typo in the later section - I changed this to `user-name-goes-here` for exactly the reason you mention, but didn't do it everywhere.  Will submit a PR for that correction.

PR #2989, opened 2019-08-16
Update config-wsl.md

pieandcakes:
Add a note that if users are using the **Remote - WSL** option, to not use the below launch.json for debugging.


pieandcakes:
@WardenGnaw

PR #3001, opened 2019-08-20
clarifying some of the storage steps

johnpapa:
- clarified that the storage account will take a moment to be created.
- renamed Index.html to index.html
- added Angular and Vue to the note


johnpapa:
Thank you for reviewing and merging!

PR #3045, opened 2019-08-30
Update tutorial-django.md

jpapp25:
fixed issue with hello_there template using single quotes when it needed double


jpapp25:
minor fix to hello_there.html file in tutorial

kraigb:
@jpapp25 Can you also submit a PR to the https://github.com/Microsoft/python-sample-vscode-django-tutorial repo to make sure the sample matches the changes to the article? Please tag me in that one and then I can do both PRs together.

PR #3051, opened 2019-09-03
Add remote highlights to 1.38 release notes

Chuxel:
As an experiment, I'd like to add a short list of 1.38 highlights to the remote section of the core VS Code release notes.  Its presence here seems to drive more uptake than their presence in the release notes.


Chuxel:
@gregvanl - Merging since I see you're OOF good chunks of the day.  LMK if you have major objections and I can revert as needed.

PR #3056, opened 2019-09-04
Update jupyter-support.md

WieserSoftware:
Suggest that the code snippet above is saved with the correct .py extension to enable the functionality


JoshuaPartlow:
Thank you for your contribution.

PR #3060, opened 2019-09-08
Update keyboard-shortcuts.md

moisesejimenezg:
Update wrong actions


gregvanl:
Thanks.
That page is not actually displayed on the VS Code website as it was for generating the keyboard shortcuts cheat sheets but still nice to have it be correct.

PR #3066, opened 2019-09-10
Changes to Line 43

Mohit-Nain:
Add in a tsconfig.json file does not make correct sense. 
If we are adding a new file it should Add a tsconfig.json file, or if we are adding to already existing file it should be Add in tsconfig.json file.


gregvanl:
Thanks

PR #3067, opened 2019-09-10
Line Break in line 63

Mohit-Nain:
Line break in line 63 will help reading the file better


gregvanl:
Current sentence is fine without comma.

PR #3072, opened 2019-09-10
docs: Remove inaccurate number

grant:
There are 17k questions right now. But there will be more in the future.

State something generic like `thousands`.


gregvanl:
Indeed. Thanks!

PR #3091, opened 2019-09-15
updated documentation to securely add gpg key to system

jahabibi:
Small change to documentation on how to add the trusted key into the system's keyring securely.


Tyriar:
apt-key is deprecated, that's why it's like this: https://github.com/Microsoft/vscode/issues/2973#issuecomment-266815997

PR #3096, opened 2019-09-18
Remove automatic indentation docs for Python

brettcannon:
Feature was backed out.


the0demiurge:
Identation has not been fixed yet:
image

Expected:

```python
if true:
    print('true')
else:
    print()
```

brettcannon:
@the0demiurge no it has not. The best place to follow this is https://github.com/microsoft/vscode-python/issues/481 and the microsoft/vscode-python repo itself.

PR #3097, opened 2019-09-19
Add markers "# region" and "# endregion" for Python

nuno-andre:
In Python spaces between "#" and "region" / "endregion" are allowed
https://github.com/microsoft/vscode/pull/46885/files


gregvanl:
Thanks!

PR #3106, opened 2019-09-25
Update containers-advanced.md

marckassay:
Changed referenced named volume to `try-node-node_modules` as this is the only declared volume. And this name is used in the proceeding example.

### DIFF
--- a/docs/remote/containers-advanced.md
+++ b/docs/remote/containers-advanced.md
@@ -126,7 +126,7 @@ services:
     volumes:
        # Or wherever you've mounted your source code
       - .:/workspace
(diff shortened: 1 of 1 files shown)


Chuxel:
The rest of the instructions in that section actually refer to the same volume name - if we update it here, we should update it in the other spots too.  Generally the point is that you can name that whatever you want.

Happy to take a look at a PR update if you'd like to make those edits as well.

Chuxel:
Oops - on second glance, your update is correct.  I was looking in the wrong spot!  Sorry about that.

PR #3112, opened 2019-09-26
Change links to migrated Node.js tutorials

kraigb:


kraigb:
@gregvanl OK to merge this when the tutorials are live; check for https://docs.microsoft.com/en-us/azure/javascript/tutorial-vscode-serverless-node-01 showing up, which should be later today.

PR #3118, opened 2019-09-29
docs: don't use PS stdout redirection doing UTF-16

ComFreek:
If you do `pylint pylint --generate-rcfile >.pylintrc` in PowerShell,
then this file gets written in UTF-16. This in turn is not accepted by
`pylint`, which will throw the following error upon *any* execution:
  """
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position
    0: invalid start byte
  """
See also https://stackoverflow.com/q/52418511/603003.

Also when executing "linting" in VSCode, VSCode just does nothing. After fixing the encoding, linting works from within VSCode and from shell via `pylint`.

### DIFF
--- a/docs/python/linting.md
+++ b/docs/python/linting.md
@@ -151,6 +151,13 @@ To control which Pylint messages are shown, add the following contents to an opt
 
 You can easily generate an options file using Pylint itself:
 
+```bash
+# Using an *nix shell or cmd on Windows
+pylint --generate-rcfile > .pylintrc
+```
+
+For PowerShell you have to explictly specify a UTF-8 output encoding:
+
 ```ps
 pylint --generate-rcfile | Out-File -Encoding utf8 .pylintrc
 ```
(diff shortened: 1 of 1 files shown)


JoshuaPartlow:
Hi @ComFreek - Thanks for your PR contribution. Given your issue is with PowerShell, would it be possible for you to break that out as a second code block with the ```ps identifier in your PR?

ComFreek:
@JoshuaPartlow Done! I would have found

> ```bash
> # Using an *nix shell or cmd on Windows
> pylint --generate-rcfile > .pylintrc
>
> # Using PowerShell
> pylint --generate-rcfile | Out-File -Encoding utf8 .pylintrc
> ```

slightly better juse because it spares us one "natural language" text line. (For people like me who just skim through docs this is more efficient.) Anyway, I sticked to your suggestion.

JoshuaPartlow:
Thanks @ComFreek. We appreciate your contribution to the docs!

PR #3122, opened 2019-10-01
Remove 'half-width kana' from 1.39 release note

smikitky:
Related: https://github.com/microsoft/vscode/pull/79735

I believe we will (and should) keep using full-width kana. Just updating CSS will not make katakana half-width. ["Half-width kana"](https://en.wikipedia.org/wiki/Half-width_kana#Encoding) normally refers to special katakana characters that are assigned [different Unicode code points](https://en.wikipedia.org/wiki/Halfwidth_and_Fullwidth_Forms_(Unicode_block)). Fonts like Meiryo UI have **full**-width katakana that are narrower than average.


octref:
You are right. Thanks for the correction!

PR #3126, opened 2019-10-02
Update the Java jdks

amamounelsayed:
We need to point to the Azul java jdks, aligned with https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-java-maven


gregvanl:
Thanks

PR #3127, opened 2019-10-03
VS Code Remote Updates for v1.39

Chuxel:
This PR covers:

1. SSH: Adding Windows as an experimental platform
2. Containers:
    1. New support for persisting container settings when attaching
    2. The new "Open Repository in Container" flow
    3. A clarification on how to deal with SSH keys in a container when you are not running as root (from customer feedback)
4. Advanced containers article updates:
    1. Clarifications and simplifications for connecting to a remote docker host now that the "docker.certPath" and "docker.tlsVerify" settings are supported.
    2. Corrections, clarifications to "Use a targeted named volume" based on customer feedback
    3. Mentioning ".env" files in the environment variables section, since we've seen issues with people being confused about how to use them.


Chuxel:
Meant to do this to vnext.

PR #3128, opened 2019-10-03
VS Code Remote Updates for v1.39 

Chuxel:
This PR covers:

1. SSH: Adding Windows as an experimental platform
2. Containers:
    1. New support for persisting container settings when attaching
    2. The new "Open Repository in Container" flow
    3. A clarification on how to deal with SSH keys in a container when you are not running as root (from customer feedback)
4. Advanced containers article updates:
    1. Clarifications and simplifications for connecting to a remote docker host now that the "docker.certPath" and "docker.tlsVerify" settings are supported.
    2. Corrections, clarifications to "Use a targeted named volume" based on customer feedback
    3. Mentioning ".env" files in the environment variables section, since we've seen issues with people being confused about how to use them.

This time merging to vnext instead of master. 😊


docs/remote/containers-advanced.md
```diff
@@ -82,68 +120,74 @@ If you've already built the container and connected to it, run **Remote-Containe
...
 
-```json
-"workspaceFolder": "/workspace"
-```
+    volumes:
+      - try-node-node_modules:
+    ```
 
-If you're running in the container with a [user other than root](#adding-a-non-root-user-to-your-dev-container), add a postCreateCommand to update the owner of the folder you mount since it may have been mounted as root.
+2. Next, be sure the `workspaceFolder` property in `devcontainer.json` matches the place your actual source code mounted:
```
chrmarti: "is" mounted?

docs/remote/containers.md
```diff
@@ -228,6 +249,29 @@ Run **Remote-Containers: Attach to Running Container...** command from the Comma
...
+```json
+{
+    // Default path to open when attaching to a new container.
+    "workspaceFolder": "/path/to/code/in/container/here",
+
+    // An array of extension IDs that specify the extensions to
+    // install inside the container when you first attach to it.
+    "extensions": [
+        "dbaeumer.vscode-eslint"
+    ]
```
chrmarti: Additionally "forwardPorts" can be set to an array of numbers, each representing a port on the container's side that will be forwarded automatically.

docs/remote/containers-advanced.md
Chuxel: Yep - made this change in vnext.  Thanks.

docs/remote/containers.md
Chuxel: Great - added to vnext.  Missed that one.

PR #3130, opened 2019-10-03
docs: add "Automated publishing" section to "Continuous Integration"

### LINKED ISSUE #3125: Extension continuous integration should also instruct how to publish to the Marketplace
Related: https://github.com/microsoft/vscode-vsce/issues/361

From @Raagh:

> Hi @joaomoreno thanks for your response, already read that unfortunately only covers the part about CI, not the CD. CI was not the problem. The problem was how to make a release pipeline upload the extension to the marketplace using proper publisher id and personal access token.
>
> in case somebody comes here and its interested I manage to solve it by after a build pipeline that publish the packaged vsix as an artifact. Adding 2 artifact references(github repo and packaged vsix) to the release pipeline and having 2 tasks on this pipeline, one npm for installing vsce globaly and a command line for running vsce publish -p {{PERSONAL ACCESS TOKEN}}.

From @moshfeu:

> And in case your extension is public and you don't want to expose PAT to the world, follow this answer: https://stackoverflow.com/q/58192624/863110

moshfeu:
resolve #3125

### DIFF
--- a/api/working-with-extensions/continuous-integration.md
+++ b/api/working-with-extensions/continuous-integration.md
@@ -102,39 +102,14 @@ You can configure the CI to publish a new version of the extension automatically
 
 The publish command is similar to the command of publishing from local environment using [`vsce`](https://github.com/Microsoft/vsce) service but this time the command needs to pass also the the Personal Access Token ("PAT").
 
-So, to configure the CI to also publish the extension, follow those steps:
-
-1. Install `vsce` as `devDependencies` (`npm install vsce --save-dev` or `yarn add vsce --dev`)
-2. Add a `deploy` script in `package.json`
-
-```json
-"scripts": {
-  "deploy": "vsce publish -p [your-personal-access-token]"
-}
-```
-
-3. Add the publish step in `azure-pipelines.yml`
-```yaml
-- bash: |
-    echo ">>> Publish"
-    yarn deploy
-  displayName: Publish
-  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Agent.OS'], 'Linux'))
-```
-
-You can use the [`condition`](https://docs.microsoft.com/azure/devops/pipelines/process/conditions) to tell the CI not to publish in certain cases. For example, publishing only if the tests are passed, only for `master` builds or only when the build is running on a specific platform.
-
-### Publishing a public project
-
-In case that the project is [public](https://docs.microsoft.com/azure/devops/pipelines/repos/github#public-github-repository-and-azure-pipelines-in-a-public-project) the PAT probably shouldn't be exposed with the rest of the code.
-
-In this case, you can store the PAT in a "secret variable". The value of that variable will not be exposed and you can use it in the `azure-pipelines.yml` file.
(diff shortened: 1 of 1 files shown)


api/working-with-extensions/continuous-integration.md
```diff
@@ -95,3 +95,59 @@ script:
...
+  "deploy": "vsce publish -p"
+}
+```
+
+2. Use it in the `publish` step in `azure-pipelines.yml`
+
+```yaml
+- bash: |
+    echo ">>> Publish"
+    yarn deploy $([the-name-of-the-secret-variable])
```
joaomoreno: Let's just have `yarn deploy $(VSCODE_MARKETPLACE_TOKEN)` here, and alude to that name above so people do not get confused.

api/working-with-extensions/continuous-integration.md
```diff
@@ -95,3 +95,59 @@ script:
...
+  displayName: Publish
+  condition: and(succeeded(), eq(variables['Build.SourceBranch'], 'refs/heads/master'), eq(variables['Agent.OS'], 'Linux'))
```
joaomoreno: We should explain why `eq(variables['Agent.OS'], 'Linux')` is here and when one might want to remove it (if one only has one agent).

joaomoreno: Also, I believe something like `eq(variables['Build.SourceBranch'], 'refs/heads/master')` is wrong, since it would simply attempt to publish every single push. A better approach is to publish whenever a tag gets pushed to the repo. We can do this by first adding triggers to the yaml:

```yaml
trigger:
  branches:
    include: ['*']
  tags:
    include: ['*']
```

And then the condition would be:

```
  condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
```

joaomoreno: Also, I believe something like `eq(variables['Build.SourceBranch'], 'refs/heads/master')` is wrong, since it would simply attempt to publish every single push.

A better approach is to publish whenever a tag gets pushed to the repo. We can do this by first adding triggers to the yaml:

```yaml
trigger:
  branches:
    include: ['*']
  tags:
    include: ['*']
```

And then the condition would be:

```
  condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/'))
```

api/working-with-extensions/continuous-integration.md
```diff
@@ -95,3 +95,59 @@ script:
...
+The publish command is similar to the command of publishing from local environment using [`vsce`](https://github.com/Microsoft/vsce) service but this time the command needs to pass also the the Personal Access Token ("PAT").
+
+So, to configure the CI to also publish the extension, follow those steps:
+
+1. Install `vsce` as `devDependencies` (`npm install vsce --save-dev` or `yarn add vsce --dev`)
+2. Add a `deploy` script in `package.json`
+
+```json
+"scripts": {
+  "deploy": "vsce publish -p [your-personal-access-token]"
```
joaomoreno: Not sure we should advocate for doing this. I would recommend simply merging these two sections and skip this step altogether. Using secret variables is the only right way to do this.

moshfeu:
Thanks!
Please review the changes.

joaomoreno:
Thanks @gregvanl!

PR #3142, opened 2019-10-09
SUSE SLE15+ requirements

alfsuse:
Added requirements to attach succesfully SLE15+ based containers.


Chuxel:
Apologies for the delay - Thanks for the PR.  

In this table, the first column should list the specific packages that can be installed to meet the requirements (for example if you're using a really minimal install).  What you have in that column then goes in the right-most column.  Would you be able to update to include the packages to install?

Chuxel:
Thanks - However, what we're looking for for a specific distro is the exact packages that meet the base requirements.  You can see that in the RedHat and Ubuntu sections, but the specific required packages.

The section below then shows what is missing from Vanilla installations of a specific distro.  This section is intended to show the underlying requirements. SUSE tends to be similar to RedHat with package names, but  they could differ.

Here's an example showing what appears for RHEL/CentOS:

| Distribution | Base Requirements | Remote - SSH Requirements | Notes |
|--------------|-------------------|------------------|-------|
| RHEL / CentOS 7+ | `glibc libgcc libstdc++ python ca-certificates tar` | `openssh-server bash` and `curl` or `wget` |   Requires kernel >= 3.10, glibc >= 2.17, libstdc++ >= 3.4.18.  RHEL / CentOS < 7 does not meet this requirement without using a [workaround to upgrade](#updating-glibc-and-libstdc-on-rhel-centos-6). |
| openSUSE Leap 15+| **PACKAGE NAMES for glibc, libgcc, libstdc++, Python, Root CAs for SSL, tar GO HERE** (Python is usally installed by these, but if not, include that too) | **PACKAGE NAME FOR OPEN SSH SERVER** and `curl` or `wget` | Docker images are missing `tar`. |

yellowfish7:
Safari extension

gregvanl:
Thanks!

PR #3157, opened 2019-10-14
Re-written line 14

Mohit-Nain:
Introduced a line break and , re phrased sub sentence in line 14 to make readability better


gregvanl:
Your proposed change suggests that other extensions have better support than the VS Code built-in language support.

PR #3159, opened 2019-10-14
Update nodejs-debugging.md

samtstern:
When I edit `launch.json` in VSCode it tells me that `autoAttachChildProcesses` is not available for `attach` configurations but only for `launch` configurations.

I am actually not sure if this page is wrong or if the linter is wrong.


gregvanl:
You are correct. Thanks!

PR #3160, opened 2019-10-14
Reformatted region section to a Table

Alipoodle:
Reformatted the for creating reigon markers into a Table for better visiablility of it.

Looking at it as a List I found it very hard to read and understand.
With it formatted into a table it makes it clear markers of which are the Starting regions and ending regions.

Additionally I have removed the HTML from it due to not having any, although it uses the same region markers as Markdown.

VS Code Image


gregvanl:
Nice work. Thanks!

PR #3162, opened 2019-10-15
Typo in Line 22

Mohit-Nain:
corrected the typo in line 22


gregvanl:
The current text is correct. HTML "include" is the correct term for bringing in outside styles and scripts from other files.

PR #3176, opened 2019-10-21
Updating hello - httptrigger-java

ryancrawcour:
In the example of running Func after deploying to Azure the code snippet says to call api/hello, but the code that the maven artefact creates creates httptrigger-java

Either the instructions need updating, as per this PR, or, the instructions need to change to tell the reader to change the Function name before deploying.


gregvanl:
Thanks! Good catch.

PR #3187, opened 2019-10-25
Webview guide fixes

IllusionMH:
First of all - thank you for such fun and detailed guide about Webviews. It really helped me to clarify some questions I had.

This PR has 2 small corrections
1. Fixed setting ID that was a bit off.
1. In my opinion 'global scope' is term that is often used in plain JS and clearly communicates variables scope where malicious scripts can access saved reference. However I have no hard filing about this terminology and will drop that commit if needed.


mjbvz:
Thanks!

PR #3189, opened 2019-10-27
Made it clearer this is just for Linux containers

gabrielmccoll:
I think it's easy to miss as it is now and there will be an idea that it will work since it's Windows and it's Microsoft.

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -9,7 +9,7 @@ DateApproved: 10/14/2019
 ---
 # Developing inside a Container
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @gabrielmccoll Good idea for the clarification. To keep the first paragraph flow general, I'd rather not add the Linux qualifier there but think it's a great idea to add your clarification on the first mention of Windows. Can you change that sentence to read "Windows container images are not supported"? Thanks

gabrielmccoll:
Love to yep did think that was maybe too much.  First time doing this so not sure how to change and try again .

gabrielmccoll:
Think I did it @gregvanl

gregvanl:
Perfect. Thanks!

PR #3192, opened 2019-10-28
Add step to copy nginx.conf file

kraigb:


kraigb:
@qubitron Dan, let me know if there's more needed than what I added here. I'm going ahead and merging, but we can make another PR if needed.

qubitron:
Looks good to me!

PR #3193, opened 2019-10-28
Fix typo in java lang intro.  Be very explict about how to open getting started from the Command Palette

edburns:
…

@gregvanl  Fix a typo.


gregvanl:
Thanks!

PR #3194, opened 2019-10-28
Remove extra isFileSystemResource

dlech:
For some reason, isFileSystemResource was duplicated twice in a row.


gregvanl:
Thanks!

PR #3197, opened 2019-10-30
Add note to dev containers doc emphasising the importance of extension-specific dependencies 

masoncusack:
Came across this issue when trying to use Live Share within dev containers. Live Share has system-level .NET Core native dependencies, which need installing with apt-get (in an Ubuntu image specifically). As far as I can see, the current documentation doesn't highlight this as an issue when discussing extension installation in general, and doesn't mention the fix: to manually add any required install commands to the dev container's Dockerfile.

I've added a note to address this in the hope it will steer others away from time-consuming and non-instructive errors like:

- "Agent terminated with exit code: 0 and signal null"
- "Command liveshare.startFromActivityBar not found"


gregvanl:
Thanks! Sorry for the delay in merging your PR. We've also moved the content to an existing Common Question "I'm seeing an error about a missing library or dependency".

masoncusack:
Thanks Greg. I realise this is small so thanks for the eyes. Hope it will help others. 👍

PR #3202, opened 2019-11-01
Fix incoherence in Virtual Documents guide

### LINKED ISSUE #3183: Virtual Documents does not document usage of onDidChange
https://code.visualstudio.com/api/extension-guides/virtual-documents
has the text: "That's all what's needed to make VS code listen for changes of virtual document. The next section will add an editor action that uses the event emitter."

Then goes on to show a command that replaces/adds a new uri instead. (The reverse string)

The needed information was on how and when to call onDidChangeEmitter.fire()

I found an actual use case in the [ContentProvider](https://github.com/microsoft/vscode-extension-samples/blob/master/contentprovider-sample/src/referencesDocument.ts) example, rather than the linked [Virtual Documents](https://github.com/microsoft/vscode-extension-samples/tree/master/virtual-document-sample) example.  It's worth noting that the usage style of the ContentProvider example is much different - and finding it took a bit of hunting and luck.

A piece of information that would be useful to prevent an easy mistake I made: the argument to the fire() command needs to be a uri of the same scheme as provided to registerTextDocumentContentProvider in order to actually fire the update.

joanrieu:
The virtual documents guide talks about event emitters and introduces an example which is supposed to use them but actually doesn't. 
The text has been updated to reflect what the code really does, and the mention of the event emitter feature is now accompanied by a link to a relevant example.

Fixes #3183

### DIFF
--- a/api/extension-guides/virtual-documents.md
+++ b/api/extension-guides/virtual-documents.md
@@ -51,7 +51,7 @@ With this we have a fully functional text document content provider. The next se
 
 ### Update Virtual Documents
 
-Depending on the scenario virtual documents might change. To support that, providers can implement a `onDidChange`-event. It must be fired for an uri and the editor will then ask for the new contents - assuming the document is still in use.
+Depending on the scenario virtual documents might change. To support that, providers can implement a `onDidChange`-event.
 
 The `vscode.Event`-type defines the contract for eventing in VS Code. The easiest way to implement an event is `vscode.EventEmitter`, like so:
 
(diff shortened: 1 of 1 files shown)


jrieken:
Maybe the better change is to fire the change event

joanrieu:
That was my initial thought as well but that would require a major change to the https://github.com/microsoft/vscode-extension-samples 's virtual documents example code, and there's already one that does it in a more complete way anyway so we might as well link it. It might be worth rewriting the guide to focus on that other example from start to finish but that's a much bigger job, and we might as well fix the current guide in the mean time.

jrieken:
Yeah, that makes sense

gregvanl:
Thanks!

PR #3204, opened 2019-11-01
Added more links for Powershell and VSCode

salmanmkc:
Making it more consistent with the rest of the document.


gregvanl:
Hi @salmanmkc Thank you for submitting a PR. In general, we don't want readers to get link fatigue so we usually only link to a technology like Powershell once in an article. I also don't think it is necessary to link to the VS code website since the user is reading the topic there.

PR #3209, opened 2019-11-04
fix typo

louiechristietest:
Change to 'on macOS' for consistency with rest of documentation (not 'on the macOS').


gregvanl:
Thanks! We needed to be more careful when we did the update from "Mac" to "macOS". ;-)

PR #3213, opened 2019-11-06
Line 33 is updated

Mohit-Nain:
changing the sentence to use singular form rather than plural.


gregvanl:
Current text reads fine.

PR #3214, opened 2019-11-07
Add note on how to persist bash history

lawrencegripper:
This is a quick edit to show users how to persist bash history between dev container sessions.


gregvanl:
Thanks! Nice tip.

PR #3216, opened 2019-11-07
Line 28 modified

Mohit-Nain:
Line 28 has been modified to make better sense, version of mono once installed will not interfere .


gregvanl:
Prefer original text.

PR #3217, opened 2019-11-07
Introduced Line Break

Mohit-Nain:
Line break is required in line 40 to make readability better.


gregvanl:
Sentence is not too long without break

PR #3218, opened 2019-11-07
Typo correction in line 40

Mohit-Nain:


gregvanl:
I think we want to keep the sentence conditional.

PR #3221, opened 2019-11-07
Use MochaOption for consistency and shorter code #3220

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>


octref:
IIRC, there's some typing Mocha issues that causes `useColors` to cause type error. Both code would be valid though.

I'd prefer not to change. Otherwise I'd have to change vscode-test, the samples, the extension generator too.

PR #3223, opened 2019-11-07
Homebrew is mentioned for Python but not Scoop

Restia666Ashdoll:
Its pity that a product by Microsoft suggests Homebrew for Mac but not Scoop and Chocolatey for Windows. I find it easier to install through Scoop then to download a file and go through an installer.

### DIFF
--- a/docs/python/python-tutorial.md
+++ b/docs/python/python-tutorial.md
@@ -47,8 +47,12 @@ Along with the Python extension, you need to install a Python interpreter. Which
 
 Install [Python from python.org](https://www.python.org/downloads/). You can typically use the **Download Python** button that appears first on the page to download the latest version.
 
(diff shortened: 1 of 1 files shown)


JoshuaPartlow:
Hi @Restia666Ashdoll - Thanks very much for bringing Scoop and Chocolatey to my attention, as well as for taking the time to create a PR. At this time we don't want to expand the Interpreter installation section of the tutorial, but I'll make note of your suggestions for the future. Again, thanks for taking the time to create a PR, if you have other feedback for the docs or of your experience using the tutorial we'd love to hear it.

Restia666Ashdoll:
It's really shame that Microsoft won't mention Windows package managers but will do so for OS X/Linux.

PR #3224, opened 2019-11-07
Add NPX option for running the generator

rheimus:


octref:
Thanks for your contribution! It's the beginner tutorial, so I'd prefer to just assume basic knowledge and stick to using the basic NPM global install for yeoman.

PR #3229, opened 2019-11-09
Describes how to connect to a remote host directly

andyljones:
Source is [this comment on an old issue](https://github.com/microsoft/vscode-remote-release/issues/190#issuecomment-506411258)


gregvanl:
Hi @andyljones Thank you for the PR. We're currently discussing whether the URI format might change in the future and whether it makes sense to document it in its present form.

gregvanl:
Thanks for putting this together! 
We're going to move this to the Tips and Tricks topic.

PR #3234, opened 2019-11-11
Update docs now that the Docker extension is a workspace extension by default

Chuxel:
Partial resolution to https://github.com/microsoft/vscode-remote-release/issues/1826

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -355,12 +355,12 @@ If there are extensions that you would like always installed in any container, y
 
 Extensions are typically designed and tested to either run locally or remotely, not both. However, if an extension supports it, you can force it to run in a particular location in your `settings.json` file.
 
-For example, the setting below will force the Docker and Debugger for Chrome extensions to run remotely instead of their local defaults:
+For example, the setting below will force the Docker extension to run locally and Debugger for Chrome extension to run remotely instead of their defaults:
 
 ```json
 "remote.extensionKind": {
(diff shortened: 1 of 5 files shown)


karolz-ms:
@Chuxel @gregvanl thank you for taking care to update the docs to reflect the current Docker extension behavior.

Question though: in case if someone is working with a remote machine that has its own Docker daemon, but wants to see/manipulate Docker assets controlled by their local Docker daemon, to me by far the best option is the one that you list first in the docs, namely, to open a different VS Code window that uses local workspace/context. 

I see no advantage to forcing the extension to be loaded as "UI" extension, or going through the hoops of setting up a development container just to see local Docker assets. So why are we even describing these, arguably inferior, options in the docs?

Am I missing something?

Chuxel:
@karolz-ms That pre-supposes that working with local containers is the exception flow.  If it's primary, switching to UI makes more sense -- particularly if you prefer not to have multiple active VS Code windows open. That's why it was UI by default originally.

karolz-ms:
OK, I guess I am just so used to having multiple VS Code windows open that I haven't even imagined someone might not like it. Thanks for clarification!

Chuxel:
@karolz-ms Oh, and one other thing, on adding the Docker CLI into the container, that's actually the Docker-in-Docker and Kubernetes - Helm scenarios.  It happens when you are using a container as a development sandbox and deploying to other containers - even if just to test your non-dev config.   

The links there point to the dev containers in the other repo that we just updated. :)

PR #3241, opened 2019-11-15
Fix numbering under "Build the program" section.

michelleangela:
Fix numbering of last step of "Build the program".


gregvanl:
Hi @michelleangela Want to fix https://code.visualstudio.com/docs/cpp/config-msvc#_start-vs-code-from-the-developer-command-prompt as well?

gregvanl:
Also https://code.visualstudio.com/docs/cpp/config-clang-mac#_build-the-program

michelleangela:
Thanks for pointing out the other numbering issues. Added the fix.

gregvanl:
Thanks!

PR #3243, opened 2019-11-17
Fix typo on "make" -> "rake" files

ShMcK:
Just fixes typo on `make`, changing it to `rake`. Was a bit confused by the error.


gregvanl:
@ShMcK Thanks for pointing out the possible confusion!

PR #3245, opened 2019-11-17
docs: update custom setup with vscode-test

SimonSiefke:
I have updated the example for [custom setup with vscode-test]( https://code.visualstudio.com/api/working-with-extensions/testing-extension#custom-setup-with-vscodetest) 
1. I have changed `resolveCliPathFromExecutablePath` because `resolveCliPathFromExecutablePath` has been renamed to `resolveCliPathFromVSCodeExecutablePath`. 
2. I have also changed the imports from commonjs to esm because all the other example use esm.
3. I have put an async main function around the code so that one can just copy and paste the example without getting an error. However I do think that the other variant is more readable and I can change it back if you want :)

### DIFF
--- a/api/working-with-extensions/testing-extension.md
+++ b/api/working-with-extensions/testing-extension.md
@@ -13,7 +13,7 @@ Visual Studio Code supports running and debugging tests for your extension. Thes
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


octref:
Looks good. Thanks for helping out!

PR #3246, opened 2019-11-18
Line 46 line update

Mohit-Nain:
Use of And is not appropriate in line 46.


gregvanl:
Current text reads fine. I think a slash separator is more confusing "are/are".

Mohit-Nain:
but use of not is incorrect, it can either be are or it can be are not, i suggest we replace and with or

PR #3247, opened 2019-11-18
Line 14 modification

Mohit-Nain:
modified line 14 , as there is clickable text available and not button.


gregvanl:
It is a button.

image

Click is not accessible language as it implies a user can use a mouse. Prefer "select", "choose", or "use" for a button or link action.

PR #3256, opened 2019-11-19
Typo Correction in line 179

Mohit-Nain:


gregvanl:
The user is running in the container with root privilege. The container doesn't run as root.

PR #3257, opened 2019-11-19
further should be used in line 232

Mohit-Nain:
farther is used for physical distances, for steps further is recommended


gregvanl:
Thanks

PR #3259, opened 2019-11-21
docs: Add "icon" property to the "command" example

moshfeu:
I was thinking that command's icon can be only string, therefor can't support light/dark mode.
I was happy to find out (by other resources) that it does possible.
A short example was save me some time 😀


gregvanl:
Thanks!

PR #3267, opened 2019-11-24
Typo in command line code.

carbontax:


gregvanl:
Hi @carbontax The current line is correct. You can open VS Code scoped to a folder by passing the folder location as an argument. In the example, using `code ./helloworld` is a condensed way to do:

```bash
cd ./helloworld
code .
```

carbontax:
Thanks Greg. (So I have some other problem then. )

PR #3271, opened 2019-11-27
`env-file` doesn't work in docker-compose

mandubian:
I tested it from documentation and it was failing with:

```sh
docker-compose -f /XYZ/.devcontainer/docker-compose.yml config --services
ERROR: The Compose file '/XYZ/.devcontainer/docker-compose.yml' is invalid because:
Unsupported config option for services.remote_dev: 'env-file'
```
In docker-compose doc https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option, it's `env_file` and not `env-file`

### DIFF
--- a/docs/remote/containers-advanced.md
+++ b/docs/remote/containers-advanced.md
@@ -57,7 +57,7 @@ Next, depending on what you reference in `devcontainer.json`:
 * **Dockerfile or image**: Edit `devcontainer.json` and add a path to the `.env` file relative to the location of `devcontainer.json`:
 
     ```json
-    "runArgs": ["--env-file","devcontainer.env"]
+    "runArgs": ["--env_file","devcontainer.env"]
     ```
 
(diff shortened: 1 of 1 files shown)


docs/remote/containers-advanced.md
```diff
@@ -57,7 +57,7 @@ Next, depending on what you reference in `devcontainer.json`:
 * **Dockerfile or image**: Edit `devcontainer.json` and add a path to the `.env` file relative to the location of `devcontainer.json`:
 
     ```json
-    "runArgs": ["--env-file","devcontainer.env"]
+    "runArgs": ["--env_file","devcontainer.env"]
```
gregvanl: The docker run switch is actually --env-file. It is only dockercompose which is env_file.
I'll revert this change.

gregvanl:
Hi @mandubian Thanks for catching this and submitting a PR. I believe the docker run argument is still --env-file but you are correct for the dockercompose setting.

mandubian:
oh yes, it's still a classic arg for docker... thks

PR #3274, opened 2019-11-29
Corrected apt-get command

sjetha:
The apt-get command for Debian/Ubuntu doesn't have sudo, which is required. I updated the doc to include this, as well as add in that the command can be run for both Debian and Ubuntu.


gregvanl:
Thanks!

PR #3275, opened 2019-12-02
Add 'acceptSuggestionOnCommitCharacter' option

klettier:
I search for this option every time I switch to a new pc so it would be great to have it directly in the doc


gregvanl:
Thanks!

PR #3279, opened 2019-12-04
Update the Getting Started part for JUnit and TestNG

jdneo:
+@hexiaokai 

P.S. the wording `projects that manually manage their dependencies` might be needed to update once we have the public name for the `invisible project`


hexiaokai:
@jdneo Thank you! We've received document feedback asking how to write unit tests, this is a great addition.

PR #3281, opened 2019-12-04
Fix letter casing for default clang-format config

pastelmind:
The default clang-format configuration given in the docs includes `BreakBeforeBraces: AllMan`, which is cased incorrectly and causes clang-format 6.0 to fail when used in an actual `.clang-format` file. The correct case is `Allman` with a lowercase m.


gregvanl:
Thanks!

PR #3285, opened 2019-12-06
Mounting host dirs with docker in devcontainer

lawrencegripper:
This outlines how to mount directories from the `devcontainer` into docker containers started.


gregvanl:
Thanks for writing this up!
This is improved in 1.41 with `remoteEnv` and `containerEnv` properties in `devcontainers.json`. We'll update the code block.

PR #3286, opened 2019-12-08
Update advanced remote missing apt update

ichbestimmtnicht:
No install without apt update possible. Image ubuntu:bionic 
Added
```bash
    && apt-get update \
```
into the "adding non-root user" example.


Chuxel:
Yeah that's a good suggestion given the snippet is in isolation. Merging.

PR #3289, opened 2019-12-10
Line22 changes

Mohit-Nain:
images can be build using the step by step commands, specify is not the right word to use


gregvanl:
I think the intent of the sentence is to let the user know that the steps are specified in the Dockerfile.

PR #3298, opened 2019-12-12
fix: update folding marker docs

kthy:
* Added Perl5 (cf. microsoft/vscode#85642)
* Fixed language sorting to alpha
* Added both Bat region styles

### DIFF
--- a/docs/editor/codebasics.md
+++ b/docs/editor/codebasics.md
@@ -255,19 +255,20 @@ Regions can also be defined by markers defined by each language. The following l
 
 Language|Start region|End region
 --------|------------|----------
+Bat| `::#region` or `REM #region`|`::#endregion` or `REM #endregion`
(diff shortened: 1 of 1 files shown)


kthy:
The change for Bat is related to microsoft/vscode#36911

gregvanl:
Hi @kthy @aeschli Given this is for new Perl5 folding supporting coming in the next release, we can either move this PR to the vscode-docs vnext branch or just leave this PR open until 1.42 is shipped early next year. Either approach is fine.

gregvanl:
Manually applied changes to vscode-docs vnext with https://github.com/microsoft/vscode-docs/commit/ec3e9d07724510eaf0c3b7200b09c9c22d3be7c3. This will be published when vnext is merged to master when VS Code 1.42 is released.

PR #3304, opened 2019-12-13
Typo correction in line 28

Mohit-Nain:


gregvanl:
Separate here is a direction for the user. A shortened version of "You separate multiple paths with a semicolon."

PR #3312, opened 2019-12-17
fix: Update gallerycdn -> gallery

nschonni:
Troubleshooting update install issues and found the URL being called was docsmsft.gallery.vsassets.io


gregvanl:
@octref Do you have context on this URI? Does this change look correct?

gregvanl:
Hi @nschonni I checked with the engineering team and both URI versions are used. Can you change your PR to add gallery and also keep gallerycdn?

gregvanl:
Thanks, I'll make sure both URI are mentioned.

PR #3322, opened 2019-12-29
Update remote-extensions.md

ShaneOsborne:
Incorrect link for VS Code extension, was https://vso-docs/vscode should be https://aka.ms/vso-docs/vscode


gregvanl:
Thanks!

PR #3324, opened 2019-12-30
Remove mentions of postinstall

matklad:
Looks like it is no longer required to get intellisense.

Moreover, as specified the postinstall does not work, as there is no vscode dependency.

I am not completely sure this is the right thing to do, but I think it is?


api/get-started/extension-anatomy.md
```diff
@@ -86,7 +85,6 @@ Each VS Code extension must have a `package.json` as its [Extension Manifest](/a
     "vscode:prepublish": "npm run compile",
     "compile": "tsc -p ./",
     "watch": "tsc -watch -p ./",
```
octref:
```suggestion
    "watch": "tsc -watch -p ./"
```

octref:
Thanks!

PR #3342, opened 2020-01-08
Custom Data blog post

octref:


blogs/2019/11/26/custom-data-format.md
```diff
@@ -0,0 +1,94 @@
...
+}
+```
+
+Then configure `html.customData` to point to the data file:
+
+```json
+{
+  "html.customData": ["./html.html-data.json"]
+}
+```
```
kieferrm: Please make more clear that this is the end user example. You only refer to the extension authors in the next section, but we miss the example, where an extension does nothing else than providing custom data. Then the transition from the end user example to the extension example could be something like: "Since not all users want to configure custom data themselves, you can also provide an extensions ..."

blogs/2019/11/26/custom-data-format.md
```diff
@@ -0,0 +1,94 @@
...
+```ts
+import { getLanguageService } from 'vscode-html-languageservice'
+
+getLanguageService({
+  customDataProviders: [...]
+})
+```
+
+By loading different data sets, one can easily create an extension that supports a superset or a subset of HTML/CSS. For example, the [MAVO](https://github.com/octref/vscode-mavo) extension load all HTML entities with MAVO-specific global attributes to get auto completion and hover information for those attributes. It is also possible to write extensions for HTML4, XHTML or SVG by restricitng the entities being loaded.
+
```
kieferrm: The above section is best thought of as a behind the scenes topic with a reference to "You can also use this yourself.."

blogs/2019/11/26/custom-data-format.md
```diff
@@ -0,0 +1,94 @@
...
+Finally, we curate a [dataset in the Custom Data Format](https://github.com/microsoft/vscode-custom-data/tree/master/web-data) from various sources:
+
+- W3C specifications of HTML, CSS and WAI-ARIA
+- Mozilla Developer Network
+- Mozilla's [mdn-data](https://github.com/mdn/data) and [mdn-browser-compat-data](https://github.com/mdn/browser-compat-data) packages
+- Chrome's ranking of CSS properties usage
+
+The data is used in VS Code's HTML/CSS language support and published to the [vscode-web-custom-data](https://www.npmjs.com/package/vscode-web-custom-data) NPM package. You may find the data useful for implementing language servers for languages that compile down to HTML/CSS, such as Pug or Stylus.
+
```
kieferrm: This data set is included by default. So, a good way of positioning this whole section would be to show that we use the mechanism ourselves and that it's also usable if you implement your own language server.

blogs/2019/11/26/custom-data-format.md
```diff
@@ -0,0 +1,94 @@
...
+## Looking forward
+
+VS Code is a text editor built with web technologies, and we are commited to providing a good editing experience for the web. Through the Custom Data Format, we stay close to the latest HTML and CSS languages and offer a simple interface for users and extension authors to customize their HTML and CSS editing experience.
```
kieferrm: `built with web technologies` -> I propose to strike. There is no added value in this statement at this point.

kieferrm:
@octref The last revision looks good to me. Thanks for making the changes!  There are only a couple of things left to do:
- Some language tweaks are needed, please work with @gregvanl.
- The **Looking Forward** section does not contain any information about what's coming. It's more a closing/summarizing statement. Either change the heading accordingly or consider removing the section.

PR #3348, opened 2020-01-14
Fix non-root link

stuartleeks:
This link navigates to the id: https://code.visualstudio.com/docs/remote/containers-advanced#_adding-a-nonroot-user-to-your-dev-container
This link does not: https://code.visualstudio.com/docs/remote/containers-advanced#_adding-a-non-root-user-to-your-dev-container


gregvanl:
Thanks!

PR #3350, opened 2020-01-17
Added clarification regarding the location of a new cpp file.

pkeir:
The current tutorial documentation [here](https://code.visualstudio.com/docs/cpp/config-msvc#_add-a-source-code-file) will produce a new file `Untitled-1`. The user must then realise that to name it as required, they must save the file. Alas, the default save location in the popup box is the .vscode subdirectory, and the code will not compile with the configuration outlined in the previous tutorial sections.


reconix23:
thanks for this update.

gregvanl:
Thanks!

PR #3351, opened 2020-01-17
Fixes in section "Avoiding extension reinstalls on container rebuild"

FrancescoIlario:
The Dockerfile snippet in the section "Changed Avoiding extension reinstalls on container rebuild" was creating the ".vscode" directory as root and giving to the local user the rights only on the subdirectory "extensions".
When vscode is then run in the container it tries to create the some directories (e.g. ".vscode/bin"), but it fails because the local user has no rights on the ".vscode" directory


Chuxel:
Good suggestion!  Thanks.

PR #3353, opened 2020-01-19
Correct jupyter-support-py.md reference to IPython

rdipietro:
Jupyter was previously known as IPython Notebook, not IPython. (IPython is the kernel that underlies both old and new notebooks.) See, for example, https://ipython.org/notebook.html


gregvanl:
Thanks for the clarification!

PR #3355, opened 2020-01-20
replace text with argument

spinelsun:
in the "Avoiding extension reinstalls on container rebuild" section's snippet I changed user-name-goes-here to the $USERNAME argument in the command chown -R


gregvanl:
Hi @spinelsun Thank you for submitting a PR. Your change was handled by an earlier PR and commit https://github.com/microsoft/vscode-docs/commit/57c6e6dcc393e6876dd4ea9ec68f84f35342fbc6.

PR #3359, opened 2020-01-22
Flask update

ucheNkadiCode:


gregvanl:
Labelling as 'don't merge' while Uche double-checks the tutorial flow.

ucheNkadiCode:
@gregvanl I have now double and triple checked. After our discussion, you're right that it makes the most sense to just navigate into the app folder to run. So I just went ahead and updated the clarity of how to get to the and that is all. Thank you!

gregvanl:
@ucheNkadiCode I've manually applied your final changes with https://github.com/microsoft/vscode-docs/commit/129479eed08c26e69801926839abc7370dca20f9 to avoid the unnecessary commit history. Thanks for tracking this down!

PR #3367, opened 2020-01-29
docs: fix variable name in .env example

janosh:


gregvanl:
Thanks! Might be a day or so until this is published.

PR #3376, opened 2020-01-31
v1.42 container extension updates

Chuxel:
Covers:
1. Forward ports
2. Improved edit loop
3. Pre-built images
4. Updates to the "attached" workflow
5. Dotfiles (including https://github.com/microsoft/vscode-remote-release/issues/2230#event-2992618760)
6. Add to devcontainer.json menu item for extensions
7. Private repo support for open repo in container
8. https://github.com/microsoft/vscode-remote-release/issues/2201

At this point I think we really do need to break out the entire "In Depth" and "Devcontainer Reference" section into another doc - dotfiles feels like the last straw to me.  However, that's a separate PR.


docs/remote/containers.md
```diff
@@ -264,18 +278,16 @@ Editing your container configuration is easy. Since rebuilding a container will
...
-   1. **Remote-Containers: Reopen Folder Locally**, which will open a new local window.
-   2. In this local window: Edit the contents of the `.devcontainer` folder as required.
-   3. Try it again: Go back to the container window, **Developer: Reload Window** from the Command Palette (`kbstyle(F1)`)
-   4. Repeat as needed.
-5. If the build was successful, but you want to make more changes:
-      1. Edit the contents of the `.devcontainer` folder as required when connected to the container.
-      2. **Remote-Containers: Rebuild Container**.
-      3. On failure: Follow the same workflow above.
+4. If you see an error, click on **Open Folder Locally** in the dialog that appears.
+5. A copy of the **build log will appear** in the new, local window so you can investigate the problem. Edit the contents of the `.devcontainer` folder as required.
```
chrmarti: Detail: It's not a "new" window anymore, the window is reused. That should make handling the window(s) easier. It also required a way to get at the old log. In addition to opening the log automatically (which might not always be the right thing to do...), there is also the command **Remote-Containers: Open Log File...**.

docs/remote/containers.md
```diff
@@ -384,21 +404,29 @@ A value of `"ui"` instead of `"workspace"` will force the extension to run on th
...
+```json
+"forwardPorts": [3000, 3001]
+```
+
+Simply reload / reopen the window and the setting will be applied when VS Code connects to the container.
+
 ### Temporarily forwarding a port
 
-Sometimes when developing you may need to access a port in your container that you didn't add to `devcontainer.json`, your Dockerfile, or Docker Compose file. If you want to **temporarily forward** a new port for the duration of the session, run the **Remote-Containers: Forward Port from Container...** command from the Command Palette (`kbstyle(F1)`).
+If you need to access a port that you didn't add to `devcontainer.json` or publish in your Docker Compose fle, you can to **temporarily forward** a new port for the duration of the session, by running the **Remote-Containers: Forward Port from Container...** command from the Command Palette (`kbstyle(F1)`).
```
chrmarti: These can be restored by enabling a core setting:
image

docs/remote/containers.md
```diff
@@ -543,6 +571,29 @@ From the Containers Explorer you can right-click on a running container and stop
 
 If you want to clean out images or mass-delete containers, see [Cleaning out unused containers and images](/docs/remote/troubleshooting.md#cleaning-out-unused-containers-and-images) for different options.
 
+## Personalizing with dotfile repositories
+
+Dotfiles are files whose filename begins with a dot (`.`) and typically contain configuration information for various applications. Since development containers can cover a wide range of application types, it can be useful to store these files somewhere so that you can easily copy them into a container once it is up and running.
+
+A common way to do this is to store these dotfiles in a GitHub repository and then use a utility to clone and apply them. The Remote - Containers extension has bulit-in support for using these with your own containers. If you are new to the idea, take a look at [the different dotfiles bootstrap repositories](https://dotfiles.github.io/) that exist.
+
+However, note that **your local `.gitconfig` file is automatically copied into the container**, so if you are using a tool that will error if a file already exists, you may need to modify the install script to remove the file first.
```
chrmarti: That should be fixed. We now install the dotfiles first and then check if there is a remote `.gitconfig` and only copy the local one if there is none.

docs/remote/containers.md
```diff
@@ -543,6 +571,29 @@ From the Containers Explorer you can right-click on a running container and stop
...
+
+To use it, update your user `settings.json` file as follows:
+
+```json
+{
+    // Dotfile repository
+    "remote.containers.dotfiles.repository": "https://github.com/your-github-id/your-dotfiles-repo.git",
```
chrmarti: `your-github-id/your-dotfiles-repo` also works. The other two settings are shown with their default value. Would it make sense to include a screenshot of the form settings editor instead? That's what the user usually sees.

PR #3379, opened 2020-02-02
Add some plurals to command reference

brettcannon:
There also might be other places where "all" and plurality is called for (i.e. other commands that return arrays), but without knowing whether that means multiple providers are called or a single provider with multiple values is returned I didn't want to be presumptuous and get it wrong.


gregvanl:
Thanks

gregvanl:
These are auto-generated from product comments, but I can port the corrections back to the vscode repo.

brettcannon:
I opened https://github.com/microsoft/vscode/pull/89959 to fix this upstream.

Issue #3399, opened 2020-02-11
For Docker SSH scenarios, using an SSH tunnel is no longer necessary


bwateratmsft:
Beginning in version 0.9.0 of the Docker extension we added proper SSH support, so it is [no longer necessary to use an SSH tunnel to connect](https://github.com/microsoft/vscode-docs/blob/master/docs/remote/troubleshooting.md#using-an-ssh-tunnel-to-connect-to-a-remote-docker-host). Instead, users can directly set the `DOCKER_HOST` environment variable or `docker.host` VSCode setting to the appropriate `ssh://user@host` URL. It is still necessary to set up agent authentication. More information on our wiki: https://github.com/microsoft/vscode-docker/wiki/SSH#directly-via-ssh

gregvanl:
@bwateratmsft It looks like this is already covered in the section https://code.visualstudio.com/docs/containers/ssh#_directly-via-ssh which was ported from the wiki.
Are there more updates required?

bwateratmsft:
Ah, I wasn't aware that'd gotten moved into the docs (bear with me, been on parental leave 😄). Anyway, could we link to that SSH setup doc from here: https://github.com/microsoft/vscode-docs/blob/master/docs/remote/troubleshooting.md#using-an-ssh-tunnel-to-connect-to-a-remote-docker-host

And also, probably downplay the tunnel approach, as it isn't needed anymore and isn't the "preferred" way?

bwateratmsft:
@Chuxel What do you think about reducing or perhaps even removing that section about SSH tunneling? Or at least linking back to our [doc](https://code.visualstudio.com/docs/containers/ssh)? Using a tunnel isn't strictly necessary anymore and somewhat less "preferred" compared to directly configuring `DOCKER_HOST` (or `docker.host`) to an `ssh://...` value.

ghogen:
@bwateratmsft @Chuxel @gregvanl The doc says "However, you may run into situations where this [simple SSH method] does not work in your environment due to SSH configuration complexity or other limitations. In this case, an SSH tunnel can be used as a fallback."

I was about to add a link back to the SSH [doc](https://code.visualstudio.com/docs/containers/ssh) but as for removing the SSH tunneling section, what about those situations discussed here where configuration is more complex? Are there real scenarios for tunneling?

bwateratmsft:
I would say yes. Given the near infinite ways to connect to a remote machine I'm sure there are users who must use an SSH tunnel.

[...]

PR #3400, opened 2020-02-11
Add link and change H3 heading

ghogen:
This is for issue #3399

### DIFF
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -611,9 +611,9 @@ As a workaround, use your Docker ID to sign in to Docker rather than your email.
 
 There is [known issue with Docker for Mac](https://github.com/docker/for-mac/issues/1759) that can drive high CPU spikes. In particular, high CPU usage occurring when watching files and building. If you see high CPU usage for `com.docker.hyperkit` in Activity Monitor while very little is going on in your dev container, you are likely hitting this issue. Follow the [Docker issue](https://github.com/docker/for-mac/issues/1759) for updates and fixes.
 
-### Using an SSH tunnel to connect to a remote Docker host
+### Using SSH to connect to a remote Docker host
 
-The [Developing inside a container on a remote Docker Machine or SSH host](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) article covers how to setup VS Code when working with a remote Docker host. This is often as simple as setting the `docker.host` property in `settings.json` or the `DOCKER_HOST` environment variable to a `ssh://` or `tcp://` URI.
+The [Developing inside a container on a remote Docker Machine or SSH host](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) article covers how to setup VS Code when working with a remote Docker host. This is often as simple as setting the `docker.host` property in `settings.json` or the `DOCKER_HOST` environment variable to a `ssh://` or `tcp://` URI.  For information on how to set this up, see [Connect to remote Docker over SSH](/docs/containers/ssh).
 
 However, you may run into situations where this does not work in your environment due to SSH configuration complexity or other limitations. In this case, an SSH tunnel can be used as a fallback.
 


ghogen:
@gregvanl I think this is ready to go.

gregvanl:
Let's have @Chuxel take a look as well.

docs/remote/troubleshooting.md
```diff
@@ -611,9 +611,9 @@ As a workaround, use your Docker ID to sign in to Docker rather than your email.
 
 There is [known issue with Docker for Mac](https://github.com/docker/for-mac/issues/1759) that can drive high CPU spikes. In particular, high CPU usage occurring when watching files and building. If you see high CPU usage for `com.docker.hyperkit` in Activity Monitor while very little is going on in your dev container, you are likely hitting this issue. Follow the [Docker issue](https://github.com/docker/for-mac/issues/1759) for updates and fixes.
 
-### Using an SSH tunnel to connect to a remote Docker host
```
Chuxel: This will break a link from the other article, so we should update the anchor there as well.

docs/remote/troubleshooting.md
```diff
@@ -611,9 +611,9 @@ As a workaround, use your Docker ID to sign in to Docker rather than your email.
...
+### Using SSH to connect to a remote Docker host
 
-The [Developing inside a container on a remote Docker Machine or SSH host](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) article covers how to setup VS Code when working with a remote Docker host. This is often as simple as setting the `docker.host` property in `settings.json` or the `DOCKER_HOST` environment variable to a `ssh://` or `tcp://` URI.
+The [Developing inside a container on a remote Docker Machine or SSH host](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) article covers how to setup VS Code when working with a remote Docker host. This is often as simple as setting the `docker.host` property in `settings.json` or the `DOCKER_HOST` environment variable to a `ssh://` or `tcp://` URI.  For information on how to set this up, see [Connect to remote Docker over SSH](/docs/containers/ssh).
```
Chuxel: To be clear, this section is under the Remote - Containers extension, not Remote - SSH.

If we're going to unify the two documentation sets like this, we should also look at the content under this section: https://code.visualstudio.com/docs/remote/containers-advanced#_developing-inside-a-container-on-a-remote-docker-host

The `/docs/containers/ssh` section is still very much written from the perspective of the Docker extension. You can't use Remote - SSH with Remote - Containers together, so setup needs to be described as covered in the advanced article before I think we should add a link like this and potentially cut other content.  Otherwise we'll really confuse people.

Chuxel:
My two cents on this is that we need to reconcile the Remote - Containers content on using remote hosts with this section before we have any cross linking of content. The article in question doesn't even mention the existence of the extension and instead refers just to Remote - SSH.  This is because it's written from the frame of reference of the Docker extension.

ghogen:
We need the Container Tools team to look at Chuck's comments and work with him to resolve this.

docs/remote/troubleshooting.md
bwateratmsft: The concern I have is that the original text here--about using the `docker.host` setting--has nothing to do with either Remote - Containers nor Remote - SSH. That is about configuring the Docker extension to use SSH to talk to a remote Docker daemon. My preference is not that users would use a tunnel to accomplish that.

bwateratmsft: I think also that the text you linked should be updated as well, since it also pertains to the Docker extension, not any of the Remote extensions.

Chuxel: `docker.host` is respected by Remote - Containers. So, that content is the same.

bwateratmsft: Interesting; what is the effect of using that setting? Isn't it orthogonal to attaching VSCode to a container?

Chuxel: No - it enables it.  It tells VS Code Remote - Containers to use a remote Docker Host instead of the local one.  The Docker extension does nothing more than hand off the attach to Remote - Containers.

Remote - Containers then has several other non-attach workflows that can be used against a remote endpoint.  Only about 30% of traffic for the extension is attach.  The containers extension will provision sandbox containers where you can open local source code or a remote repo and more.

Chuxel: The goal behind it is you actually develop inside the container -- even from the ground up.  So it's more than just container based applications (though those are supported too).

bwateratmsft: Ah, the lightbulb is finally coming on for me! I think I understand now. Sorry for being so slow to arrive at it.

Does the Remote - Containers extension work with `ssh://...` values for `docker.host` / `DOCKER_HOST`? If so, do you have a preference on using `ssh://...` vs. `tcp://...` + a tunnel?

Chuxel: The extension uses the Docker CLI and `dockerode` like the Docker extension, so while it supports `ssh://`, it suffers from the same `dockerode` created limitations as the Docker extension... for example the SSH key must be in a running SSH agent rather than the `~/.ssh` folder.

The primary article on configuring Remote - Containers recommends the use of the `ssh://` protocol first ([see here](https://code.visualstudio.com/docs/remote/containers-advanced#_developing-inside-a-container-on-a-remote-docker-host)) and lists TCP (w/o a tunnel) or docker-machine as alternatives.  The section above is in troubleshooting to cover scenarios where the ssh protocol is not sufficient - it's 100% an exception "how to", not the recommendation.  It's specifically in a different spot (troubleshooting) because of it.

My hope is that we can drop the mention of the tunnel entirely over time - it's just that right now, there's issues we've heard about from customers with `ssh://` support.

That said, I'm all for consolidating these instructions between the two doc sets, we'd just want to accommodate the Remote - Containers extension as well.

bwateratmsft: That makes sense. In that case I'm fine with making no changes here, at least as regards #3399. Revisiting and unifying docs is a separate, bigger task I think.

Chuxel: Yep, that was my thinking as well

karolz-ms: @ghogen @Chuxel I just wanted to point out that https://code.visualstudio.com/docs/containers/ssh does not show up in the table of contents and in fact, was not meant to be included in the initial set of updated docs for containers ☺️ But hey, it is there and we'll take it. 

@ghogen do you want me to open an issue to have Docker ext SSH topic added to the TOC?

PR #3404, opened 2020-02-13
Use a more solid getNodeModule for keytar

mtirionMSFT:
The previous code for getNodeModule works fine in a development environment, but will fail loading the Node module in production. The function should be changed as indicated. That code is coming from the vscode-azureaccount repo.

### DIFF
--- a/api/advanced-topics/remote-extensions.md
+++ b/api/advanced-topics/remote-extensions.md
@@ -238,22 +238,28 @@ Visual Studio Code does not provide a secret persistence mechanism itself, but m
 For example:
 
 ```typescript
-import * as vscode from 'vscode';
+import { env } from 'vscode';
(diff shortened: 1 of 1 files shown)


Chuxel:
The webpack build should take care of all of this for you.  What issue are you hitting?

mtirionMSFT:
Webpack doesn’t solve it. See the code in vscode-azureaccount. The problem is that the current code only works in development mode (in vscode), but when you package to vsix it cannot locate modules in the given path, while they are really there. This modification fixes that problem.

Martin

Martin Tirion  |  senior software engineer |  CSE | Agile Response Team | Spatial Computing | Microsoft
Tel +31 (6) 1250 6153  |  Twitter @mtirion  |  Blog mtirion.nl<http://mtirion.nl/>

PR #3406, opened 2020-02-14
Blog post: Optimizing CI build times

joaomoreno:
This is a shared blog post between @ethanis and I, we want to get it out in February.

@gregvanl @fiveisprime Could you give it a quick review and help us get the blog post out there? Thanks!

@gregvanl How can we get a `MetaSocialImage` for this? Maybe from @misolori?

### DIFF
--- a/blogs/2020/02/14/optimizing-ci.md
+++ b/blogs/2020/02/14/optimizing-ci.md
@@ -3,7 +3,7 @@ Order: 53
 TOCTitle: Improving CI Build Times
 PageTitle: How Visual Studio Code leaverages Azure Pipelines Artifact Caching Tasks to improve CI
 MetaDescription: How Visual Studio Code leaverages Azure Pipelines Artifact Caching Tasks to improve CI
-MetaSocialImage: /assets/blogs/2020/02/14/TODO
+MetaSocialImage: /assets/blogs/2020/02/14/hero.png
 Date: 2020-02-14
 ShortDescription: How Visual Studio Code leaverages Azure Pipelines Artifact Caching Tasks to improve CI
 Author: Ethan Dennis, João Moreno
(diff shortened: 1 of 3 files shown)


misolori:
@joaomoreno how's this for the social image?

image

blogs/2020/02/14/optimizing-ci.md
```diff
@@ -0,0 +1,117 @@
...
+
+| Platform     | Before | After | Time Savings |
+|--------------|--------|-------|--------------|
+| macOS        | 68s     | 34s    | 50%          |
+
+In total, we were thrilled to see a combined ~50% reduction in VS Code’s CI build times! The best news is that you can draw inspiration from [our build definitions](https://github.com/microsoft/vscode/tree/master/build/azure-pipelines) to realize build time improvements of your own.
+
+Happy Caching,
+
+Ethan Dennis, Azure DevOps Senior Software Engineer
```
ethanis: My product group should be listed as `Developer Services`

fiveisprime:
LGTM

blogs/2020/02/14/optimizing-ci.md
joaomoreno:
```suggestion
Ethan Dennis, Developer Services Senior Software Engineer
```

gregvanl:
@joaomoreno @ethanis Readers may be curious why the Linux builds didn't see as large an improvement (Linux Alpine is slower). Do you have any explanation you could add? Maybe just a short sentence, even if you don't know why.

blogs/2020/02/14/optimizing-ci.md
```diff
@@ -0,0 +1,117 @@
...
+
+When we last talked about our CI engineering our target platforms spanned Windows, macOS, and Linux. As of today, VS Code targets a much more diverse set of platforms, such as ARM64 and Alpine Linux for its remote server components. In total, we have eight different targets that all share common build steps. This post outlines how we leveraged the caching tasks to reduce CI duplication and further improve our build times.
+
+## Room for Improvement
+
+So, what exactly were the common steps across all build jobs? Each build target has a job that follows a similar set of steps. At a very high level, each job must:
+
+1. Restore Node packages
+2. Lint the code
+3. Compile the code
```
egamma: Suggest make it more clear that in VS Code we compile from TypeScript to JavaScript and therefore would describe this step as: Compile the code from TypeScript to JavaScript.

blogs/2020/02/14/optimizing-ci.md
```diff
@@ -0,0 +1,117 @@
...
+4. Run unit tests
+5. Run integration tests
+
+Our caching tasks were an obvious choice to speed up the “Restore Node packages” step. E.g. why run an expensive `npm install` step, when you could cache the results from a previous run, given that the `package-lock.json` file rarely changes? Since we’ve discussed caching packages previously, what makes this post interesting is how we applied caching to the other steps.
+
+Since linting and compilation are platform-independent, those steps could easily be run by a single build agent that would share its results with other platform-dependent agents. We created a Linux build agent whose sole responsibility was exactly this: restore packages, lint, and compile the code. All we had to do was to share the results with other agents.
+
+## Cache Everything
+
+In order to share cache results across build agents, we needed platform independent caches, which were initially unsupported by the caching tasks. We added a `platformIndependent` parameter to the task for creating and restoring platform-independent caches. For example:
```
egamma: A general impression I had after reading the blog was that it was not fully clear what is VS Code specific and what is a general pipeline improvement that everyone could leverage.

To make this more clear I suggest the following change:

>We added a `platformIndependent` parameter to the task for creating and restoring platform-independent caches

Azure Pipline task now supports an addition parameter `platformIndependent`. This parameter is leveraged in the VS Code build as follows:

blogs/2020/02/14/optimizing-ci.md
```diff
@@ -0,0 +1,117 @@
...
+    targetfolder: target
+    vstsFeed: $(ArtifactFeed)
+    platformIndependent: true
+```
+
+When caching node modules, it’s logical to use `package-lock.json` files as a cache key: when this file changes, we must invalidate the cache. When caching compilation outputs, the entire source code must act as a cache key. To simplify things, we decided to use the HEAD commit as the cache key—a new commit will inevitably create a new cache entry. This works fine for our purposes since a single build, despite running across build agents, always runs over a single commit.
+
+Another missing feature was the ability to create multiple caches per build job. We now found ourselves juggling two caches (node modules, compilation) with no way to address each cache individually. The caching tasks output an environment variable called `CacheRestored` that can be used to optimistically skip build tasks. This environment variable works great in builds that interact with a single cache, but not so great with multiple caches—leaving us to wonder which cache `CacheRestored` is referring to. Once again, we worked on a way to add an optional `alias` parameter to the task. For example:
```
egamma: Again make it clear that this is not VS Code specific and that this is a pipelines extension

>Once again, we worked on a way to add an optional `alias` parameter to the task.

blogs/2020/02/14/optimizing-ci.md
```diff
@@ -0,0 +1,117 @@
...
+  condition: ne(variables['CacheRestored-Packages'], 'true')
+```
+
+Here, an alias of `Packages` is appended to the environment variable output, allowing us to cache NPM packages and compilation output in a single build job. We were finally deduplicating a lot of the CI work that can now be executed only once and shared across platform-specific agents.
+
+There was still room for one final optimization, given a specific use case: build resubmissions. We sometimes must retrigger builds on previously built commits since tests might be flaky or some agents might randomly fail. Ideally, the shared agent would not restore or recompile the common code but simply defer to the platform-dependent agents to perform their work. The issue we noticed was that the compilation cache packages were massive and restoring them would take around 8 minutes—all for nothing since the shared agent would simply yield control if that cache existed. We addressed this issue by implementing a `dryRun` feature to allow us to check for a cache package’s existence without restoring it—effectively cutting 8 minutes off our build resubmissions. Using the `dryRun` parameter looks like this:
```
egamma: Make 'we' more specific

>We sometimes must retrigger builds on previously built commits since tests might be flaky or some agents might randomly fail.

In VS Code we...

joaomoreno:
@misolori Thanks for the hero, added it.

@gregvanl @egamma  Thanks for all the feedback, I've incorporated said changes.

joaomoreno:
@gregvanl Might need to change the folder date, right?

gregvanl:
Yep, I'm doing an edit pass now and can update the date as well.

PR #3418, opened 2020-02-21
Documented the password attribute of the promptString input variable …

jlittorin:
…type.

@alexr00, could you please have a look at this?


alexr00:
Thanks for updating the docs!

PR #3419, opened 2020-02-21
Update c-cpp-properties-schema-reference.md

Alex-EEE:


docs/cpp/c-cpp-properties-schema-reference.md
```diff
@@ -62,7 +62,7 @@ MetaDescription: Schema reference for C++ project settings in Visual Studio Code
   A friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that will be autoselected on those platforms. The status bar in VS Code will show you which configuration is active. You can also click on the label in the status bar to change the active configuration.
 
 - `compilerPath` (optional)
-  The full path to the compiler you use to build your project, for example `/usr/bin/gcc`, to enable more accurate IntelliSense. The extension will query the compiler to determine the system include paths and default defines to use for IntelliSense.
+  The full path to the compiler you use to build your project, for example `/usr/bin/gcc`, to enable more accurate IntelliSense. The extension will query the compiler to determine the system include paths and default defines to use for IntelliSense.  Putting "compilerPath": "" will skip this step: Useful if your compiler doesn't support this query, as VS Code will default back to any compiler it can find ( like Visual C ) if you give it a bad folder, exe, or even if you just leave this field out
```
michelleangela: Surround "compilerPath":"" with code format markdown so it becomes:
 
`"compilerPath": ""`

michelleangela: Rewording suggestions:

Setting `"compilerPath": ""` will skip querying a compiler, which is useful if a specified compiler doesn't support the arguments that are used to query it, as the extension will default back to any compiler it can find (like Visual C) if the input value is invalid or the property `compilerPath` is removed from c_cpp_properties.json (i.e. it is not set).

gregvanl:
Thanks!

PR #3415, opened 2020-02-19
remove Docker troubleshooting

ghogen:

PR #3436, opened 2020-02-28
Remove troubleshooting topic

ghogen:
The troubleshooting content will be updated in the Wiki, not needed here.  No redirection necessary.


ghogen:
@gregvanl As discussed and agreed to in email, this topic needs to be removed. There are no links to it, and no redirection is required.

gregvanl:
@ghogen Looks like a dup of https://github.com/microsoft/vscode-docs/pull/3415. I'll merge this and just close the other.

PR #3438, opened 2020-03-01
Update tslint-eslint-migration.md

wafs:
If you're using npx you don't need to install anything globally, it'll download it on demand if it's not located on the system, which is helpful since we don't want this package hanging around once it's done.


gregvanl:
Thanks!

PR #3443, opened 2020-03-04
Python Debugging Update

ucheNkadiCode:
Adding Python Debug Page for containers.


docs/containers/debug-common.md
```diff
@@ -109,8 +138,22 @@ Example `launch.json` configuration for debugging a .NET Core application:
...
+
+| Property | Description | Default |
+| --- | --- | --- |
+| `host` | The host for remote debugging. | |
+| `port` | The port for remote debugging. | `5678` |
+| `pathMappings` | Maps the project path between local machine and remote host.  | |
+| `projectType` | Type of Python app. | |
+| `justMyCode` | Debug only user-written code. | |
+| `django` | Django debugging. | `false` |
+| `jinja` | Jinja template debugging (e.g. Flask). | `false` |
```
ghogen: Do not use "e.g." or "i.e." in docs - can you rephrase?

docs/containers/debug-python.md
```diff
@@ -0,0 +1,149 @@
+---
+Area: containers
+ContentId: f9ffec31-9253-4f71-a4eb-79ea7b3a8f55
+PageTitle: Configure and troubleshoot debugging of Python apps running in a Docker container
+DateApproved: 01/29/2020
```
ghogen: update date

docs/containers/debug-python.md
```diff
@@ -0,0 +1,149 @@
...
+MetaDescription: How to configure and troubleshoot debugging of Python apps running in a Docker container, using Visual Studio Code.
+---
+
+# Debug Python within a container
+
+When adding Docker files to a Python project, tasks and launch configurations are added to enable debugging the application within a Docker container. To accommodate the various scenarios of Python projects, some apps may require additional configuration.
+
+## Configuring the Docker container entry point
+
+The Docker extension infers the entry point of the Docker container via properties of `tasks.json`. This container entry point is automatically configured when using the **Docker: Add Docker Files to Workspace...** command.
```
ghogen: Trying to rephrase to make it more clear what applies to the user - is this right?
```suggestion
You can configure the entry point of the Docker container by setting properties in `tasks.json`. VS Code automatically configures the container entry point when you first use the **Docker: Add Docker Files to Workspace...** command.
```

docs/containers/debug-python.md
```diff
@@ -0,0 +1,149 @@
...
+        "file": "hello_app/webapp.py"
+      }
+    }
+  ]
+}
+```
+
+## Automatically launching the browser to the entry page of the application
+
+When the **Docker: Python - Django** or **Docker: Python - Flask** launch configurations are selected, the Docker extension will automatically launch the browser to the main page of the app. While this feature is enabled by default, this behavior can be configured explicitly via the `dockerServerReadyAction` object in `launch.json`.
```
ghogen: Can you rephrase to make it more applicable to the user?  Something like "You can select the **Docker: Python - Django** or **Docker: Python - Flask** launch configuration to automatically launch the browser..." and then later "you can configure this behavior explicitly by setting the `dockerServerReadyAction` object in *launch.json*.  BTW, I think filenames should be in italics (single asterisk) rather than code.

docs/containers/debug-python.md
```diff
@@ -0,0 +1,149 @@
...
+          {
+            "localRoot": "${workspaceFolder}",
+            "remoteRoot": "/app"
+          }
+        ],
+        "projectType": "django"
+      },
+      "dockerServerReadyAction": {
+        "action": "openExternally",
+        "pattern": "Starting development server at (https?://\\S+|[0-9]+)",
```
ghogen: That regular expression might need a brief explanation ?

docs/containers/debug-common.md
ucheNkadiCode: updated this with (such as Flask)

docs/containers/debug-python.md
ucheNkadiCode: changed to 03/05/2020

docs/containers/debug-python.md
ucheNkadiCode: Thanks, I made the updates to be a more applicable to the user. In terms of single asterisks on the filename, for consistency I will keep it as single ticks. The reason why is this is done throughout the containers section when referring to any file.

docs/containers/debug-python.md
ucheNkadiCode: Tagging @haniamr for an of how the pattern & uriFormat interact. I'll update it with a ">**Note**: " below the json example

haniamr: The expression simply attempts to capture: "Starting development server at http://localhost:8000". It accommodates variations in the url such http or https, any host name and port.

The relation between the pattern and uriFormat is as follows:
- We capture a group from the regex (i.e. anything enclosed in (...)), so in the above example it will be the url
- If the matched group is a port number, then only 1 placeholder (i.e. %s) in the uriFormat is expected
- If the matched group is a url, then 2 placeholders are expected, one for the scheme and the other for the port

Hope that helps!

ucheNkadiCode: So would there be an error here if the log output message only said "Starting development server at 8000"? Because in this case it would only be capturing the port and not the scheme?

ucheNkadiCode: I added Hani's explanation.

PR #3445, opened 2020-03-04
rephrase awkward sentence

johnpapa:
feel seemed confusing, so I replaced with experience. if that is not the intent, please edit this PR.

thanks


gregvanl:
Agreed, that is awkward. Thanks!

PR #3452, opened 2020-03-06
Update theming.md

franzese:
These links are 404ing


gregvanl:
Hi @franzese Thank you for the PR. However, the links work correctly on the live VS Code site https://code.visualstudio.com/api/extension-capabilities/theming.
My guess is that you are seeing 404 when you try to navigate when reading the topic on GitHub. Unfortunately the /api folders of the vscode-docs repo don't support the same preprocessing of .md relative links used for the /docs folders, which would allows GitHub navigation.

PR #3455, opened 2020-03-08
Added Shortcut To Close VSCode.

NotMoni:
We should give them a shortcut for MacOS to shut down the terminal.

### DIFF
--- a/docs/setup/mac.md
+++ b/docs/setup/mac.md
@@ -25,7 +25,7 @@ You can also run VS Code from the terminal by typing 'code' after adding it to t
 
 ![macOS shell commands](images/mac/shell-command.png)
 
-- Restart the terminal for the new `$PATH` value to take effect. You'll be able to type 'code .' in any folder to start editing files in that folder.
+- Restart the terminal `Ctrl + Q` for the new `$PATH` value to take effect. You'll be able to type 'code .' in any folder to start editing files in that folder.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @NotMoni Shouldn't this be Cmd+Q on macOS? Also won't macOS users already know how to shut down and restart the terminal? I'm not sure we need to provide a keyboard shortcut in the VS Code documentation.
Maybe I'm misunderstanding given the issue title mentions closing VS Code.

NotMoni:
Restart the terminal is very vague and instead of letting them scavenge the web we can provide the shortcut. i think it would help them and save some time.  If the person knows then that's good, but for beginners, its gonna be troubling and such shortcut can help them. And that MD file is to help set up VSCode on mac, so i think it's a valid reason to add the shortcut. Yes, it's Ctrl + Q.

gregvanl:
It was Cmd+Q when I tested on my Mac and that is what I see on the Apple site https://support.apple.com/guide/imac/keyboard-shortcuts-apd194062a6d/mac

image

image

I wonder why Ctrl+Q works for you.

NotMoni:
Oh sorry, that was a typo @gregvanl. You are right it is Cmd + Q. I have changed it in the pr code.

NotMoni:
This is beneficial to someone new on macOS or can be a refresher to experienced people too.

gregvanl:
Closing. In general, we want people to follow the platform documentation rather than include instructions in the VS Code topics.

PR #3459, opened 2020-03-09
Correct window.restoreWindows default setting

zjopy:
In VS Code 1.42.1 the default setting for window.restoreWindows is 'all' and not like mentioned 'one'


gregvanl:
Thanks for catching and fixing this!

PR #3460, opened 2020-03-09
Update jupyter-support.md

DavidKutu:
In docs/python/jupyter-support.md #Select a code cell,
the commands of J and K were backwards.
It should be J for down and K for up.
This commit fixes that.


JoshuaPartlow:
Thanks @DavidKutu, looks good to me.

PR #3466, opened 2020-03-10
Add a one-line command for increasing file watch limit

tjhu:
Add a one-line command for increasing the file watch limit so we can easily copy and paste it to the terminal.


Tyriar:
The reason we avoid a one liner here is because there may already be an entry for `fs.inotify.max_user_watches`.

PR #3469, opened 2020-03-12
Updates to the remote section for v1.43

Chuxel:
This PR updates docs to include the following:

1. That Windows and macOS are available for Remote - SSH.
2. Updates to tips and tricks to cover targeting a Windows SSH host.
3. Adds the `initializeCommand `, `build.args` and `build.target` properties to the devcontainer.json reference.
4. Updates all the "Remote-X: Forward ..." commands to "Forward a port"

@roblourens - It looks like the README for the SSH extension and Remote Development extension pack also needs to be updated.  Since this wasn't there, I had some people from an internal reach out and ask if the release notes were accurate.

I also left the requirement to set `remote.SSH.useLocalServer` in place in the docs since I saw you mention you may not make this the default anymore given issues.

@chrmarti I didn't move the other properties under the new `build` property since VSO hasn't adapted to this yet and I don't want to cause confusion from the docs.  We can update whenever they pick up the change.


Chuxel:
@chrmarti 
> Are you also going to add the generic dotfiles settings and build.dockerfile and build.context properties?

Updated the dotfile properties.

For the `build.dockerfile` and `build.context` changes, I'm going to wait until VSO has adapted as well to avoid confusion given they just launched their custom Dockerfile support.

Chuxel:
@roblourens Quick ping here.  I'm seeing folks asking about Mac setup which this should answer as well.

docs/remote/ssh.md
```diff
@@ -47,42 +45,52 @@ To get started, you need to:
...
 
-If you do not have a host yet, you can [setup an SSH host on an existing machine](/docs/remote/troubleshooting.md#installing-a-supported-ssh-server) or create a [Linux VM on Azure](https://docs.microsoft.com/azure/virtual-machines/linux/quick-create-portal?toc=%2fazure%2fvirtual-machines%2flinux%2ftoc.json).
+2. **[Optional]** While password-based authentication is supported, we recommend setting up **key based authentication** for your host. See the [Tips and Tricks](/docs/remote/troubleshooting.md#configuring-key-based-authentication) article for details.
 
-> **Note:** When using experimental Windows 10/Server SSH host support, enable `remote.SSH.useLocalServer` [in VS Code settings](/docs/getstarted/settings.md).
->
-> ARMv7l / ARMv8l `glibc` SSH hosts, some extensions may not work due to x86 compiled native code inside the extension.
+3. If you will be connecting to a Windows server, verify **Remote.SSH: Use Local Server** is checked in [VS Code settings](../getstarted/settings.md).
```
roblourens: I don't think I will change the default back after all, so we don't have to say this. 

I might mention disabling it as a troubleshooting step, but I can add that when I decide what I'm doing.

docs/remote/troubleshooting.md
```diff
@@ -136,6 +154,19 @@ If you are running into problems with VS Code hanging while trying to connect (a
 
 Enable the `remote.SSH.showLoginTerminal` [setting](/docs/getstarted/settings.md) in VS Code and retry. If you are prompted to input a password or token, see [Enabling alternate SSH authentication methods](#enabling-alternate-ssh-authentication-methods) for details on reducing the frequency of prompts.
 
+**Work around a bug with Windows OpenSSH server 8.x**
```
roblourens: The bug actually starts with 7.9

docs/remote/ssh.md
Chuxel: Sounds good.  Updated

docs/remote/troubleshooting.md
Chuxel: Interesting - Updated it to be more generic in terms of the specific versions.

Chuxel:
@gregvanl Assuming the PR looks good to you, I think we can merge.

PR #3471, opened 2020-03-12
Python data science tutorial 

jmew:
Using this PR to leave comments


docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+
+    ![Data Science tutorial: opening the Command Palette in VS Code](images/shared/command-palette.png)
+
+1. The **Python: Select Interpreter** command presents the list of available interpreters that VS Code was able to locate automatically (your list will vary from the one shown below; if you don’t see the desired interpreter see [Configuring Python environments](/docs/python/environments.md)). From the list, select the Anaconda environment you created, which should include the text **‘myenv’: conda**.
+
+   ![Selecting a python environment](images/data-science-tutorial/anaconda-environment.png)
+
+1. Now that an interpreter is selected, run **Terminal: Create New Integrated Terminal** from the Command Palette. This command creates a terminal and automatically activates the selected environment by running its activation script. You’ll notice that the selected environment shows on the left side of the VS Code status bar, and that the “(myenv)” indicator in the terminal indicates you’re using a virtual environment.
```
jmew: I dont think we need this step, its just an extra step that is not required

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+
+   ![New terminal with Anaconda environment activated](images/data-science-tutorial/anaconda-terminal.png)
+
+1. With the environment and VS Code setup, the final step is to create the Jupyter notebook that will be used for the tutorial. From the VS Code File Explorer, use the New File icon to create a Notebook file named **hello.ipynb**.
```
jmew: We should push for creating the new jupyter notebook using the Command Palette way rather than the new file way

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+
+   ![Running a Jupyter notebook cell](images/data-science-tutorial/jupyter-cell-01.png)
+
+1. After the cell finishes running, you can view the data that was loaded using the variable explorer and data viewer. First click on the chart icon in the notebook's upper toolbar, then the data viewer icon to the right of the *data* variable. For additional information about the data set, refer to [this document](http://biostat.mc.vanderbilt.edu/wiki/pub/Main/DataSets/titanic3info.txt) about how it was constructed.
+
+   ![Data viewer and variable explorer](images/data-science-tutorial/data-viewer.png)
+
+   You can then use the data viewer to view, sort, and filter the rows of data. After reviewing the data, it can then be helpful to graph some aspects of it to help visualize the relationships between the different variables.
+
+1. Before the data can be graphed though, you need to make sure that there aren’t any issues with it. If you look at the Titanic csv file, one thing you’ll notice is that a question mark (“?”) was used to designate cells where data wasn’t available.
```
jmew: Maybe we can add a screenshot of the data viewer and also get the user to try to use the Dataviewer filter to look for '?' in the columns

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+   from sklearn import metrics
+   predict_test = model.predict(X_test)
+   print(metrics.accuracy_score(y_test, predict_test))
+   ```
+
+   ![Running the trained model against test data](images/data-science-tutorial/jupyter-cell-05.png)
+
+   Looking at the result of the test data, you’ll see that the trained algorithm had a ~75% success rate at estimating survival.
+
+## (Optional) Use a neural network to increase accuracy
```
jmew: TODO Jeffrey: Change the code to use PyTorch

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+1. In VS Code, open the **hello_ds** folder and the Jupyter notebook (**hello.ipynb**), by going to **File** > **Open Folder**.
+1. Within your Jupyter notebook begin by importing the [pandas](https://pandas.pydata.org/) and [numpy](https://numpy.org/) libraries, two common libraries used for manipulating data, and load the Titanic data into a pandas [DataFrame](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html). To do so, copy the below code into the first cell of the notebook. For additional guidance about working with Jupyter notebooks in VS Code, see the documentation [here](https://code.visualstudio.com/docs/python/jupyter-support).
+
+   ```python
+   import pandas as pd
+   import numpy as np
+   data = pd.read_csv('data.csv')
+   ```
+
+1. Now, run the cell using the Run cell icon.
```
jmew: Or using the "SHIFT + ENTER" shorcut

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+
+   ```python
+   data.corr().abs()[["survived"]]
+   ```
+
+   ![Determining the correlation between input variables and survival](images/data-science-tutorial/jupyter-cell-03.png)
+
+1. Looking at the correlation results, you’ll notice that some variables like gender have a fairly high correlation to survival, while others like relatives (sibsp = siblings or spouse, parch = parents or children) seem to have little correlation.
+
+   Let’s hypothesize that *sibsp* and *parch* are related in how they affect survivability, and group them into a new column called “relatives” to see whether the combination of them has a higher correlation to survivability. To do this, you will check if for a given passenger, the number of *sibsp* and *parch* is greater than 0 and, if so, yuo can then say that they had a relative on board.
```
jmew: typo, should be "you" instead of "yuo" on the last line

docs/python/data-science-tutorial.md
```diff
@@ -0,0 +1,250 @@
...
+   This problem can be corrected by replacing the question mark with a missing value that pandas is able to understand. Add the following code to the next cell in your notebook to replace the question marks in the age and fare columns with the [numpy NaN](https://docs.scipy.org/doc/numpy/reference/constants.html?highlight=nan#numpy.nan) value. Notice that we also need to update the column's data type after replacing the values.
+
+   ```python
+   data.replace('?', np.nan, inplace= True)
+   data = data.astype({"age": np.float64, "fare": np.float64})
+   ```
+
+   > **Note**: If you ever need to see the data type that has been used for a column, you can use the [DataFrame dtypes](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.dtypes.html#pandas.DataFrame.dtypes) attribute.
+
+1. Now that the data is in good shape, you can use [seaborn](http://seaborn.pydata.org/) and [matplotlib](https://matplotlib.org) to view how certain columns of the dataset relate to survivability. Add the following code to the next cell in your notebook and run it to see the generated plots.
```
jmew: Should mention if you need to make a new cell, either use the new cell button at the bottom left of the cell, or use the "B" hotkey in command mode (when the bar is blue after you press ESC)

docs/python/data-science-tutorial.md
JoshuaPartlow: Ok removed.

docs/python/data-science-tutorial.md
JoshuaPartlow: Updated to use the command palette and toolbar save icon. kept a note mentioning you could create using new file in explorer.

docs/python/data-science-tutorial.md
JoshuaPartlow: Per conversation, keeping as Keras/Tensorflow for now.

docs/python/data-science-tutorial.md
JoshuaPartlow: Added.

docs/python/data-science-tutorial.md
JoshuaPartlow: Good catch. Fixed.

docs/python/data-science-tutorial.md
```diff
@@ -121,9 +123,9 @@ This tutorial uses the [Titanic dataset](http://biostat.mc.vanderbilt.edu/wiki/p
...
+1. Looking at the correlation results, you'll notice that some variables like gender have a fairly high correlation to survival, while others like relatives (sibsp = siblings or spouse, parch = parents or children) seem to have little correlation.
 
-   Let’s hypothesize that *sibsp* and *parch* are related in how they affect survivability, and group them into a new column called “relatives” to see whether the combination of them has a higher correlation to survivability. To do this, you will check if for a given passenger, the number of *sibsp* and *parch* is greater than 0 and, if so, yuo can then say that they had a relative on board.
+   Let's hypothesize that *sibsp* and *parch* are related in how they affect survivability, and group them into a new column called "relatives" to see whether the combination of them has a higher correlation to survivability. To do this, you will check if for a given passenger, the number of *sibsp* and *parch* is greater than 0 and, if so, yuo can then say that they had a relative on board.
```
jmew: last sentance, theres a typo with the word "you", it says "yuo". Other than that, LGTM

docs/python/data-science-tutorial.md
JoshuaPartlow: Agree with you, but leaving out for now. I want to verify how our name usage guidelines apply to something like this first and don't want to block on it

docs/python/data-science-tutorial.md
JoshuaPartlow: Note added.

docs/python/data-science-tutorial.md
JoshuaPartlow: Should be addressed now.

gregvanl:
@JoshuaPartlow Published

PR #3474, opened 2020-03-13
Update remote debugging doc.

fiveisprime:
Remote debugging is being expanded to support Python apps as well. This
change makes the doc a little more generic and specifies that both
Node.js and Python are supported.

Not quite ready for this to be merged, but I wanted to get the change ready
for the next release.

### DIFF
--- a/docs/azure/remote-debugging.md
+++ b/docs/azure/remote-debugging.md
@@ -1,15 +1,15 @@
 ---
 Order: 3
 Area: azure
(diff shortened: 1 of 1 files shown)


fiveisprime:
The app service team is testing the remote debugging stuff now so this should be able to merge in the next iteration.

fiveisprime:
I'll update this once the team finally picks this back up.

PR #3495, opened 2020-03-22
Fix scope in example for `punctuation.paren.open`

nhandl3:


nhandl3:
> The last paren at the end of the file is not part of the an expression.group since the end rule is not matched.

I think this part is not correct because in my vscode:
image

mjbvz:
Thanks!

PR #3496, opened 2020-03-22
Add `npx` alternative

ackvf:
This solution doesn't require premanent installation and uses latest packages always


gregvanl:
Hi @ackvf Thank you for the PR. To keep things simple in this hello world tutorial, we recommend using npm since users may want to run Yeoman and the VS Code generator multiple times while they are learning and experimenting.

ackvf:
I believe npx caches so if no newer version is available, it should run immediately. Anyway, the advantage of this approach is that it always tries to download latest from the servers, which goes well with user's desire to use the generator multiple times.

PR #3498, opened 2020-03-22
Remove "Blurriness on macOS Mojave" FAQ entry.

teo-tsirpanis:
This issue was reported twenty versions ago, and got fixed in Electron 6 (we are at 7). I don't think it is any relevant today.


gregvanl:
Thanks!

PR #3505, opened 2020-03-24
Fix incorrect indentation in an example

eduard-malakhov:
In `docker-compose.yml`, `version` and `services` nodes must be on the same level.


gregvanl:
Hi @eduard-malakhov Thank you for submitting a PR. 
Would you like to fix the indenting in the other code block further down in the topic?

eduard-malakhov:
Hi @gregvanl, sure, here it is. My bad, I hadn't noticed it the first time.

gregvanl:
Thanks!

PR #3506, opened 2020-03-24
Update tslint-eslint-migration.md

derkallevombau:
- 'npx tslint-to-eslint-config' is merely one line.
- From the current dbaeumer.vscode-eslint doc:
"Using the extension with VS Code's task running
The extension is linting an individual file only on typing. If you want to lint the whole workspace set eslint.lintTask.enable to true and the extension will also contribute the eslint: lint whole folder task. There is no need anymore to define a custom task in tasks.json."

### DIFF
--- a/api/advanced-topics/tslint-eslint-migration.md
+++ b/api/advanced-topics/tslint-eslint-migration.md
@@ -43,7 +43,11 @@ It is time to lint! Use this command: `eslint -c .eslintrc.js --ext .ts <mySrcFo
 "lint": "eslint -c .eslintrc.js --ext .ts <mySrcFolder>"
 ```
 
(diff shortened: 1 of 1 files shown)


jrieken:
>  lint whole folder task. There is no need anymore to define a custom task in tasks.json

:-1: for that, the `eslint.lintTask.enable` will also contribute a task but it's ad-hoc and persistent. The sample defined an npm-script for running lint checks (which you for instance could use as pre-commit hook) and defining a task for an npm-script has value

gregvanl:
Hi @derkallevombau Thank you for this PR. The first two changes look good. Can you revert the last change per @jrieken comments and then I'll merge it?

derkallevombau:
@gregvanl Done!
@jrieken Thank you for clarification!

gregvanl:
Thanks!

PR #3508, opened 2020-03-25
Embedded LSP doc

octref:
Related https://github.com/microsoft/vscode-extension-samples/pull/264

### DIFF
--- a/api/language-extensions/embedded-languages.md
+++ b/api/language-extensions/embedded-languages.md
@@ -21,7 +21,7 @@ Today, we see an increasing amount of embedded languages, such as:
 
 This guide focuses on implementing language features for embedded languages. If you are interested in providing syntax highlighting for embedded languages, you can find information on the [Syntax Highlight Guide](/api/language-extensions/syntax-highlight-guide#embedded-languages).
 
-This guide includes two samples that illustrates two approaches to building such a language server: **Language Services** and **Request Forwarding**. We'll take a look at both samples and conclude with each approach's pros and cons.
+This guide includes two samples that illustrates two approaches to build such a language server: **Language Services** and **Request Forwarding**. We'll take a look at both samples and conclude with each approach's pros and cons.
 
 Code for both samples can be found at:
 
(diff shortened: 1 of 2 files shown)


aeschli:
I like it!

Both approaches share some common tasks that we should explain first, before going into the library/forward request discussion:

- the document has a primary language that knows the overall syntax. You always need a parser for the overall syntax. The syntax also defines characters encoding, not just for the main content, but also for the embedded content. 
- in order to handle requests to content in an embedded language, a common technique is to create a sub document for that embedded language. That document is used as input to the library or for forwarding.
Some tricks (use whitespace) can help you simplifying the position mapping

I'd like us to state that having a full language service that includes also the support for embedded language is what we recommend:
- server is self contained and can be used in other editors/IDE too
- embedded content can be 
- full control over the user experience of features of the embedded region. 

Alternatively, the forwarding approach might be an option if
- embedded content is independent from the main content (no context needed)

octref:
> the document has a primary language that knows the overall syntax. You always need a parser for the overall syntax.

I thought "needing a parser for the primary language" is self-evident, just as in the language server guide we don't even mention "you need a parser".

> The syntax also defines characters encoding, not just for the main content, but also for the embedded content

I think that's a minor detail that can be explained later in the doc. What issue would you encounter if you don't consider character encoding?

> in order to handle requests to content in an embedded language, a common technique is to create a sub document for that embedded language. That document is used as input to the library or for forwarding.
> Some tricks (use whitespace) can help you simplifying the position mapping

I'm not sure if saying that upfront makes sense to people. In the language service approach, I don't think you actually care too much about "creating sub document". Our HTML/CSS language services have `TextDocument` in their API, but I can also see the API only taking a language ID and content.

> I'd like us to state that having a full language service that includes also the support for embedded language is what we recommend:

> server is self contained and can be used in other editors/IDE too
full control over the user experience of features of the embedded region.
Alternatively, the forwarding approach might be an option if

> embedded content is independent from the main content (no context needed)

All good points. I'll make the edits.

PR #3509, opened 2020-03-26
change #%% to # %%

scarlett2018:
#%% does not work, the "Run Cell, etc. commands" only shows for # %%


JoshuaPartlow:
I was actually able to get the #%% syntax to work, but you are correct that what's added with a new cell is # %%. Thanks for opening a PR on the documentation, I'll merge it in.

PR #3512, opened 2020-03-27
Update userdefinedsnippets.md

natakazakova:
I have a problem with assigning a shortcut for a user snippet. So here is my suggestion.


gregvanl:
Thanks!

PR #3516, opened 2020-03-30
Update volume target path

brpratt:
The "Avoiding extension reinstalls on container rebuild" section provides directions on how to create volume mounts to hold the remote VS Code server and extensions. In one of the steps, the volume mount target for VS Code Insiders is shown without the home prefix. Anyone using VS Code Insiders who blindly copies this configuration will end up mounting the volume to the wrong path.

This PR updates the volume mount to include the home directory in the target path.


gregvanl:
Thanks!

PR #3520, opened 2020-04-01
Remove outdated note about IntelliSense for ES7

zjopy:
This note about IntelliSense not implementing ES7 seems to be a outdated. Newer features like `String.prototype.trimStart` and `trimEnd` from ES2019 already work.

Note: My assumption here is that older features must be complete if later ones are. Please decline this PR if actually some ES7 features are not yet present even though ES2019 is already supported.


mjbvz:
Thanks!

PR #3527, opened 2020-04-02
Adding March update release localization contributors.

janbaltus:


janbaltus:
Please note the missing German contributors is not mistake, but no submit can be found for this language.

PR #3540, opened 2020-04-07
Update reactjs-tutorial.md

Gilssung:
Using create-react-app globally via npm install -g create-react-app let users to use outdated create-react-app. Using outdated create-react-app makes error that cannot provide a template.
(https://github.com/facebook/create-react-app)


gregvanl:
Thanks!

PR #3541, opened 2020-04-07
Update typescript-tutorial.md

vipulgupta2048:
Add emphasis on just running the command `tsc` and not the whole command `tsc helloworld.ts` 
This was something I was doing wrong and probably would help someone else too down the line. You can close the PR if you find this trivial and not needed. 

Also, `just` is not a nice word to have in docs but couldn't find a replacement.


gregvanl:
Good clarification, no need for such a long sentence. Thanks!

vipulgupta2048:
Happy to help out! thanks @gregvanl

PR #3542, opened 2020-04-07
Use aspnetcorerazor language.id in example instead of razor

JoeRobich:
The Razor tooling's language id is "aspnetcorerazor" (https://github.com/dotnet/aspnetcore-tooling/blob/b3ac44798c16fff5b95dbcfe62dea84aa9a1bd72/src/Razor/src/Microsoft.AspNetCore.Razor.VSCode.Extension/package.json#L135)

### DIFF
--- a/docs/editor/emmet.md
+++ b/docs/editor/emmet.md
@@ -62,7 +62,7 @@ For example:
 "emmet.includeLanguages": {
     "javascript": "javascriptreact",
     "vue-html": "html",
-    "razor": "html",
+    "aspnetcorerazor": "html",
     "plaintext": "pug"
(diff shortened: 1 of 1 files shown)


gregvanl:
This example is using the VS Code built-in Razor language support, which defines the `razor` language id.

image

I don't think we want to update it for another extension.

JoeRobich:
Ah, I didn't realize that was in the box. I assumed it was an old example from before the Microsoft C# extension was split from VS Code. Thanks for taking a look. I can just update the C# extension readme with an example for enabling emmet in our Razor LSP.

PR #3543, opened 2020-04-08
Updates to remote section for 1.44 release

Chuxel:
This PR adds:
1. Some refactoring around quick starts given improvements to open repository in container. Specifically, it uses the "Remote-Containers: Try a sample..." command as a starting point, covers how to use an existing folder next, then covers open repository mentioning its perf benifits.
2. Attach support for Kubernetes
3. Notes on experimental WSL2 support
4. A tip on the COMPOSE_PROJECT_NAME environment variable
5. Some related tweaks to release notes like mentioning the Python pre-built images

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -78,38 +78,25 @@ We will cover using a container as your full-time development environment first.
 
 Let's start by using a sample project to try things out.
 
-1. Clone one of the sample repositories below.
-
-    ```bash
-    git clone https://github.com/Microsoft/vscode-remote-try-node
-    git clone https://github.com/Microsoft/vscode-remote-try-python
-    git clone https://github.com/Microsoft/vscode-remote-try-go
-    git clone https://github.com/Microsoft/vscode-remote-try-java
-    git clone https://github.com/Microsoft/vscode-remote-try-dotnetcore
-    git clone https://github.com/Microsoft/vscode-remote-try-php
-    git clone https://github.com/Microsoft/vscode-remote-try-rust
-    git clone https://github.com/Microsoft/vscode-remote-try-cpp
-    ```
-
-2. Start VS Code and click on the quick actions Status Bar item in the lower left corner of the window.
+1. Start VS Code, and in a new window, click on the quick actions Status Bar item in the lower left corner.
 
     ![Quick actions Status bar item](images/common/remote-dev-status-bar.png)
 
-3. Select **Remote-Containers: Open Folder in Container...** from the command list that appears, and open the root folder of the project you just cloned.
+2. Select **Remote-Containers: Try a Sample...** from the command list that appears and select sample from the list.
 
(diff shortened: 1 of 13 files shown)


Chuxel:
@roblourens I wasn't sure if there were any changes needed to the SSH docs with the Windows server related changes. They generally seemed self explanatory and the recommendations in the tips section still seemed to apply.

roblourens:
Maybe we should write something to explain the platform selection prompt that will show up on Windows and the new setting `remote.SSH.remotePlatform`, but the users shouldn't have to worry about it too much, what do you think?

docs/remote/containers.md
```diff
@@ -1146,6 +1155,10 @@ Finally, some extensions (like C#) download secondary dependencies from `downloa
 
 VS Code Server runs on a random port inside the container and VS Code itself uses `docker exec` to communicate with it over Docker's configured communication channel.
 
+### Can I use the extension with source code stored in WSL2 on Windows?
```
chrmarti, line 1158: This also works with WSL 1. I actually couldn't get WSL 1 to work without it. Maybe I'm missing something?

Chuxel: Interesting - WSL1 can't run docker in it can it?  I thought that was WSL2 only

Chuxel:
> Maybe we should write something to explain the platform selection prompt that will show up on Windows and the new setting `remote.SSH.remotePlatform`, but the users shouldn't have to worry about it too much, what do you think?

Sure, we can do that. I can add something along the lines of "If VS Code cannot automatically detect the type of host, you will be prompted to pick the target OS" with a picture. That will give us some wiggle room based on how things progress over time.

docs/remote/containers.md
```diff
@@ -150,6 +139,40 @@ Once connected, you may want to **add the `.devcontainer` folder** to the worksp
...
 
+## Quick start: Open a Git repository or GitHub PR in an isolated container volume
+
+While you can [open a locally cloned repository in a container](#quick-start-open-an-existing-folder-in-a-container), you may want to work with an isolated copy of a repository for a PR review or to investigate another branch without impacting your work.
+
+Repository Containers use isolated, local Docker volumes instead binding to the local filesystem. In addition to not polluting your file tree, local volumes have the added benefit of improved performance on Windows and macOS. (See [Advanced Configuration](/docs/remote/containers-advanced.md#improving-container-disk-performance) for information on how to use these types of volumes in other scenarios.)
+
+For example, follow these steps to open one of the "try" repositories in a Repository Container:
+
+1. Start VS Code and run **Remote-Containers: Open Repository in Container...** from the Command Palette (`kbstyle(F1)`).
```
egamma, line 150: I suggest to explain that when opening a PR into a dev container, then the GH PR extension will be installed and the PR branch checked out so that you can start to review the PR.

Chuxel: Done!

docs/remote/containers.md
chrmarti: You're right, that's WSL2 only.

PR #3549, opened 2020-04-09
Update codebasics.md

davidstraka2:
Updated the "Find In Selection" section. The possible values for `editor.find.autoFindInSelection ` are currently: `never`, `always`, and `multiline`.


gregvanl:
Thanks for making this update!

PR #3552, opened 2020-04-10
Rename Inspect TM Scopes

IllusionMH:
Command title was changes in v1.42 however docs still had old name.

This PR updates name in docs to match name in VS Code.

There is third mention of old name, however it is in v1.9 release notes and I don't think it make sense to rename it there.
https://github.com/microsoft/vscode-docs/blob/master/release-notes/v1_9.md#new-tools-for-inspecting-textmate-scopes


gregvanl:
Thanks!
And I agree, we don't need to update historical release notes.

IllusionMH:
Thanks!

PR #3560, opened 2020-04-13
Python debugging topic update

JoshuaPartlow:


docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+You would then use the following configuration to attach from the VS Code Python extension.
+
+```json
+{
+    "name": "Python: Attach",
+    "type": "python",
+    "request": "attach",
+    "host": "localhost",
+    "port": 5678,
```
int19h, line 103: We should be promoting the new syntax for this:
```json5
"connect": {
  "host": ...,
  "port": ...,
}
```
This is to allow for connection direction reversal by using `--connect` and "listen". 

Also, "host" is optional for listening, and defaults to 127.0.0.1, same as for the CLI.

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+|Flag  |Options  |Description  |
+|---------|---------|---------|
+|**--listen**     |  `[<address>:]<port>`       |   **Required**. Specifies the host address and port for the debug adapter server. This is the same address that is used in the VS Code debug configuration. By default the host address is `localhost (127.0.0.1)`.      |
+|**--wait-for-client**     |   none      | **Optional**. Specifies that the code should not run until there's a connection from the debug server. This setting allows you to debug from the first line of your code.        |
+|**--log-to**     |   `<path>`      | **Optional**. Specifies a path to an existing directory for saving logs.         |
+|**--log-to-stderr**     |    none     |  **Optional**. Enables debugpy to write logs directly to stderr.       |
+|**--pid**     |    `<pid>`     | **Optional**. Specifies a process that is already running that the debugger should attach to.        |
```
int19h, line 133: When used directly from CLI, `--pid` simply injects the debug server into the process - no actual attach happens. Rather, it's as if (best effort) the targeted process were started with `--connect` or `--listen` as used along with `--pid`. It is used under the hood to implement "attach" with "processId", but that's an implementation detail that users don't need to be concerned about.

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+The debugger can also be run from the command line. The debugger command line syntax is as follows:
+
+```bash
+python -m debugpy --listen [<address>:]<port>
+    [--wait-for-client]
+    [--configure-<name> <value>]...
+    [--log-to <path>] [--log-to-stderr]
+    <filename> | -m <module> | -c <code> | --pid <pid>
+    [<arg>]...
```
int19h, line 86: It might be worth pointing out here that `<arg>...` is the command line arguments that will get passed to the app being launched as is.

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+For information about creating and using debugging configurations see the [Initialize configurations](#initialize-configurations) and [Additional configurations](#additional-configurations) sections. Once a configuration is added, it can be selected from the dropdown list and started using the **Start Debugging** button.
+
+![Start debugging](images/debugging/debug-start-button.png)
+
+## Command line debugging
+
+The debugger can also be run from the command line. The debugger command line syntax is as follows:
+
+```bash
+python -m debugpy --listen [<address>:]<port>
```
int19h, line 81: We'll need to doc `--connect` as well. I have an issue to do this on our wiki - once that's done, I'll let you know.

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+1. Run VS Code, open the folder or workspace containing the script, and create a `launch.json` for that workspace if one doesn't exist already.
+
+1. In the script code, add the following and save the file:
+
+    ```python
+    import debugpy
+
+    # 5678 is the default attach port in the VS Code debug configurations
+    print("Waiting for debugger attach")
+    debugpy.listen(('localhost', 5678))
```
int19h, line 150: `listen()` can take just the port number alone, same as `--listen` in CLI, with the host defaulting to `127.0.0.1`. We generally want to encourage that syntax, because it minimizes the risk of the user inadvertently listening on some public interface, and exposing their machine to remote code execution.

Also, the print should be moved one line below, so as to not imply that `listen()` itself is waiting - we've had some user confusion over this before.

docs/python/debugging.md
int19h: Ditto `--configure`.

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+1. Local computer: **Only if you modified the source code on the remote computer as outlined above**, then in the source code, add a commented-out copy of the same code added on the remote computer. Adding these lines makes sure that the source code on both computers matches line by line.
+
+    ```python
+    #import debugpy
+
+    # Allow other computers to attach to debugpy at this IP address and port.
+    #debugpy.listen(('1.2.3.4', 5678))
+
+    # Pause the program until a remote debugger is attached
+    #debugpy.wait_for_attach()
```
int19h, line 250: `wait_for_client()`

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+        The IP address used in `listen` should be the remote computer's private IP address. You can then launch the program normally, causing it to pause until the debugger attaches.
+
+   1. Launch the remote process through debugpy, for example:
+
+       ```bash
+       python3 -m debugpy --listen 5678 --wait-for-client -m myproject
+       ```
+
+       This starts the package `myproject` using `python3`, with the remote computer's private IP address of `1.2.3.4` and listening on port `5678` (you can also start the remote Python process by specifying a file path instead of using `-m`, such as `./hello.py`).
```
int19h, line 239: `--listen 1.2.3.4:5678`

docs/python/debugging.md
```diff
@@ -55,11 +55,214 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+1. Local computer: **Only if you modified the source code on the remote computer as outlined above**, then in the source code, add a commented-out copy of the same code added on the remote computer. Adding these lines makes sure that the source code on both computers matches line by line.
```
int19h, line 241: I think the more usual workflow is to modify the code locally first, and then deploy it to the remote machine with those modifications. Having mismatched code between local and remote, even when line numbers match, is generally not a good idea, and we might be more stringent about it in the future (e.g. do checksums).

docs/python/debugging.md
```diff
@@ -125,25 +328,13 @@ You can also rely on a relative path from the workspace root. For example, if th
...
-    "pythonPath": "^\"\\${env:SPARK_HOME}/bin/spark-submit\""
-},
-"windows": {
-    "pythonPath": "^\"\\${env:SPARK_HOME}/bin/spark-submit.cmd\""
-},
-"linux": {
-    "pythonPath": "^\"\\${env:SPARK_HOME}/bin/spark-submit\""
-},
-```
+You can specify platform-specific paths by placing `python` within a parent object named `osx`, `windows`, or `linux`.
```
int19h, line 337: This is true for all properties in the debug configuration - I wonder if we need to mention it for this one specifically. But if we do, it should probably link to https://code.visualstudio.com/Docs/editor/debugging#_platformspecific-properties

docs/python/debugging.md
```diff
@@ -125,25 +328,13 @@ You can also rely on a relative path from the workspace root. For example, if th
 "program": "${workspaceFolder}/pokemongo_bot/event_handlers/__init__.py",
 ```
 
-### `pythonPath`
+### `python`
 
 Points to the Python interpreter to be used for debugging, which can be a folder containing a Python interpreter. The value can use variables like `${workspaceFolder}` and `${workspaceFolder}/.venv`.
```
int19h, line 333: That bit about "a folder containing a Python interpreter" is not entirely clear to me. The debugger itself definitely requires this property to be a full path to the interpreter binary itself - i.e. "C:\Python\python.exe", not just "C:\Python". But maybe VSCode is doing something else here... @karthiknadig, do you recall if there are any provisions in debug adapter factory etc for pointing at folders like that?

docs/python/debugging.md
```diff
@@ -125,25 +328,13 @@ You can also rely on a relative path from the workspace root. For example, if th
 "program": "${workspaceFolder}/pokemongo_bot/event_handlers/__init__.py",
 ```
 
-### `pythonPath`
+### `python`
```
int19h, line 331: This needs to mention that the property value can be an array instead of a simple string, which allows passing arguments to the Python interpreter, e.g.:
```json5
"python": ["/usr/bin/python3", "-O"]
```

docs/python/debugging.md
```diff
@@ -163,11 +354,11 @@ When set to `true`, breaks the debugger at the first line of the program being d
...
 
-| Value | Where output is displayed |
-|--- | --- |
-| `"internalConsole"` | VS Code debug console |
+| Value                            | Where output is displayed                                          |
+|----------------------------------|--------------------------------------------------------------------|
+| `"internalConsole"`              | VS Code debug console                                              |
 | `"integratedTerminal"` (default) | [VS Code Integrated Terminal](/docs/editor/integrated-terminal.md) |
-| `"externalTerminal"` | Separate console window |
+| `"externalTerminal"`             | Separate console window                                            |
```
int19h, line 361: This is not entirely correct. The Debug Console contains *redirected* output, which is controlled by "redirectOutput" debug property. This defaults to true if you use "internalConsole", and false otherwise, but it can be manually overridden by the user. If you use it together with either terminal option, then the output will be both in the terminal window, and in Debug Console. Conversely, if you disable it for internal console, then no output will be displayed at all.

docs/python/debugging.md
karthiknadig: No. It should be full path to the interpreter binary.

docs/python/debugging.md
karthiknadig: What @int19h said in table form:

| redirectOutput | console | Effect |
|----------------- | ----------------- | -------- |
| true (default for internalConsole) | internalConsole | Output is only displayed in the Debug Console  |
| false                 | internalConsole | Output is NOT displayed anywhere  |
| true                  | integratedTerminal or externalTerminal | Output is displayed both in Debug Console and Terminal |
| false (default for  integratedTerminal or externalTerminal) | integratedTerminal or externalTerminal | Output is only displayed in the Terminal |

int19h:
I updated our wiki to add information about `--connect` and `--configure`.
https://github.com/microsoft/debugpy/wiki/API-Reference
https://github.com/microsoft/debugpy/wiki/Command-Line-Reference

docs/python/debugging.md
```diff
@@ -26,18 +26,18 @@ To initialize debug configurations, first select the Run view in the sidebar:
 
 ![Run icon](images/debugging/debug-icon.png)
```
samsungfaces9, line 27: Up

docs/python/debugging.md
JoshuaPartlow: Will be updated with next commit.

docs/python/debugging.md
JoshuaPartlow: Will be updated in next commit.

docs/python/debugging.md
JoshuaPartlow: Updated to the following, does that look like what you wanted?

```json
{
    "name": "Python: Attach",
    "type": "python",
    "request": "attach",
    "connect": {
        "host": "localhost",
        "port": 5678
    },
    "listen": {
        "port": 5678
    }
}
```

> **Note**: Specifying host is optional for **listen**, by default 127.0.0.1 is used.

docs/python/debugging.md
JoshuaPartlow: Thanks, reworded for next commit.

docs/python/debugging.md
JoshuaPartlow: Updated with the following. Look ok?

    ```python
    import debugpy

    # 5678 is the default attach port in the VS Code debug configurations. Unless a host and port are specified, host defaults to 127.0.0.1
    debugpy.listen(5678)
    print("Waiting for debugger attach")
    debugpy.wait_for_client()
    breakpoint()
    ```

docs/python/debugging.md
JoshuaPartlow: Good catch. Updated in next commit.

docs/python/debugging.md
JoshuaPartlow: That makes sense as a better flow. I was just trying to maintain the content that was already there. I'll flag this as an additional improvement to make once we get everything else updated.

docs/python/debugging.md
JoshuaPartlow: Thanks, updated in next commit.

docs/python/debugging.md
JoshuaPartlow: Thanks, added the following to that section. Look ok?

If you need to pass arguments to the Python interpreter, you can use the syntax `"python": ["<path>", "<arg>",...]`.

docs/python/debugging.md
JoshuaPartlow: Ok, corrected.

docs/python/debugging.md
JoshuaPartlow: Given it's not Python specific I removed it.

docs/python/debugging.md
JoshuaPartlow: Thanks, I tried to clarify this in the info for console and redirectOutput. Look better?

docs/python/debugging.md
int19h: Looks good!

docs/python/debugging.md
```diff
@@ -78,7 +78,9 @@ For information about creating and using debugging configurations see the [Initi
 The debugger can also be run from the command line. The debugger command line syntax is as follows:
 
 ```bash
-python -m debugpy --listen [<address>:]<port>
+python -m debugpy
+    --listen | --connect
+    [<address>:]<port>
```
int19h, line 83: This should be `[<host>:]<port>` - we've already tweaked debugpy health output accordingly. 

(address would be the whole thing, host+port - I mislabeled it in help originally)

docs/python/debugging.md
```diff
@@ -99,11 +101,18 @@ You would then use the following configuration to attach from the VS Code Python
     "name": "Python: Attach",
     "type": "python",
     "request": "attach",
-    "host": "localhost",
-    "port": 5678,
+    "connect": {
+        "host": "localhost",
+        "port": 5678
+    },
+    "listen": {
```
int19h, line 108: If you start debugpy with `--listen`, the debug config should only contain `"connect"`. And vice versa, `--connect` requires `"listen"`. Using both side by side wouldn't be meaningful, and the debugger will report an error if this is attempted.

JoshuaPartlow: Ok, corrected in latest commit.

docs/python/debugging.md
JoshuaPartlow: Gotcha, updated in latest commit.

docs/python/debugging.md
```diff
@@ -55,11 +55,223 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+
+1. Open a terminal using **Terminal: Create New Integrated Terminal**, which activates the script's selected environment.
+
+1. In the terminal, install the debugpy package with `python -m pip install --upgrade debugpy`.
+
+1. In the terminal, start Python with the script, for example, `python3 myscript.py`. You should see the "Waiting for debugger attach" message that's included in the code, and the script halts at the `debugpy.wait_for_client()` call.
+
+1. Switch to the Run view, select the appropriate configuration from the debugger drop-down list, and start the debugger.
+
+1. The debugger should stop on the `breakpoint()` call, from which point you can use the debugger normally. You can, of course, set other breakpoints in the script code using the UI instead of using `breakpoint()`.
```
karthiknadig, line 172: `breakpoint()` assumes Python 3.7 or greater. Should we specify that somewhere?

JoshuaPartlow: @karthiknadig - Is there something different you'd use for versions before Python 3.7? I could add a note or a comment in the code?

karthiknadig: It is `debugpy.breakpoint()`, this works for all versions. For 3.7 or greater it behaves the same as `breakpoint()`

JoshuaPartlow: Ah, I see. I updated all the breakpoint() calls to be debugpy.breakpoint() instead. If that works for all versions that makes sense to me to be what we show.

docs/python/debugging.md
```diff
@@ -55,11 +55,223 @@ Selecting the **Node.js: Gulp task** yields the following result:
...
+1. In the script code, add the following and save the file:
+
+    ```python
+    import debugpy
+
+    # 5678 is the default attach port in the VS Code debug configurations. Unless a host and port are specified, host defaults to 127.0.0.1
+    debugpy.listen(5678)
+    print("Waiting for debugger attach")
+    debugpy.wait_for_client()
+    debugpy.breakpoint()
```
karthiknadig, line 161:
```suggestion
    debugpy.breakpoint()
    print('break on this line')
```
/cc @int19h we need the `print` right. Since we break on the line that we are about to execute?

karthiknadig: Never mind. This is what we are recommend people add to their files. It will break on the next line we will execute. This is only a issue if there are no more valid lines after `debugpy.breakpoint()`.

JoshuaPartlow: Sounds like maybe we do need it then? You're saying if the file was just the code block in this snippet it wouldn't break right, since it stops on the line after the debugpy.breakpoint() call?

karthiknadig: If it is just the above code block then we will need the print line.

docs/python/debugging.md
int19h: On the other hand, we do want the users to use standard Python facilities when available.

PR #3565, opened 2020-04-15
Typo / Incorrect suffix

scottmwyant:
Using the `.color-theme.json` suffix does not enable hover support and code completions.  The files generated by 'yo code' use a `-color-theme.json` suffix, which works.  Example `dark-color-theme.json`.


aeschli:
Thanks @scottmwyant !

PR #3571, opened 2020-04-16
Fix Jetty for Java link

rzgry:
Link under the Jetty section is for VSCode TomCat

Fix capitalization of Red hat -> Red Hat


gregvanl:
Thanks!

PR #3579, opened 2020-04-19
Remove mentions of 32-bit from the Linux setup instructions

aaronfranke:
VS Code no longer ships 32-bit binaries for Linux, so these instructions were a bit out-of-date.


gregvanl:
@Tyriar Can you have a look at this PR? Thanks

PR #3580, opened 2020-04-20
Update mac.md

misterspeedy:
Double-clicking on the downloaded archive will run the installer rather than opening the archive (on Catalina at least - not sure about earlier macOS versions. Altered instructions to click the 'spyglass' icon instead.


gregvanl:
Thanks, nice clarification. I'm going to rename the `spyglass` to `magnifying glass`.

PR #3589, opened 2020-04-21
Fix property name in config example

xtheandrew:
Fixed the 'forcedInclude' property name on the "C++ / Customizing default settings" page.
Also removed the trailing comma from the configuration example.


gregvanl:
Thanks!

PR #3608, opened 2020-04-24
Update tutorial for new VS Code Python extension

pinguin999:
The new Python extension doesn't need the reload anymore. Makes the workflow a lot smoother.


JoshuaPartlow:
Thanks for catching that update and submitting a PR on the docs.

PR #3613, opened 2020-04-24
Docker Compose article work

ghogen:


ghogen:
@gregvanl Can we build this to stage now and have the product team review on stage?

gregvanl:
@ghogen Should be on the staging server in about 5 minutes.

PR #3618, opened 2020-04-27
Fix two semantic highlighting guide links

parkernilson:
The links point to 'semantic-highlighting-guide' (which does not exist), but they should point to 'semantic-highlight-guide'. This commit fixes two instances of this broken link.


gregvanl:
Thanks!

PR #3625, opened 2020-04-28
Use Docker context for remote machines

karolz-ms:
Docker context is a more convenient way to point VS Code and Docker CLI to a remote Docker engine.

Resolves https://github.com/microsoft/vscode-docker/issues/1797


docs/containers/ssh.md
```diff
@@ -25,10 +25,14 @@ We recommend using the Visual Studio Code [Remote-SSH extension](/docs/container
...
-    * You can use the `DOCKER_HOST` environment variable, or
+    ```shell
+    docker context create my-remote-docker-machine --docker "host=ssh://username@host:port"
+    ```
 
-    * There's a setting `docker.host` in VS Code, which has the same effect, but allows for user or workspace settings instead of machine settings.
+    If you don't include user name, the command will use your current local user name, which might be wrong. If you omit the port, it defaults to 22.
 
-1. It is recommended to change the refresh rate to something longer with the `docker.explorerRefreshInterval` setting. The connection over SSH is slow, and it can result in trying to refresh again before the previous refresh even finished. We recommend at least 3000 ms.
+1. Issue **Docker Context: Use** Visual Studio Code command to activate the Docker context pointing to the remote machine. This command causes both VS Code and Docker CLI to use the remote machine context.
```
ghogen, line 36:
```suggestion
1. Use the **Command Palette** (`kb(workbench.action.showCommands)`) to issue the **Docker Context: Use** command to activate the Docker context pointing to the remote machine. This command causes both VS Code and Docker CLI to use the remote machine context.
```

PR #3633, opened 2020-04-30
Adding note about Visual Studio Code package

miniwolf:
Since 2019.2 it is required to install the vscode package. This should be added to projects by default, but having this information here seems valuable.


gregvanl:
Thanks!

PR #3635, opened 2020-04-30
Updates to remote section for v1.45

Chuxel:
Includes:

1. Updates to several sections thanks to new variables
2. A separate reference section for variables now that there are more of them
3. Refinements to the containers documentation on WSL2 support
4. An advanced section in the WSL docs on Remote - Containers and WSL2
5. A note on Podman support and its limitations - I didn't move it into "supported" since you have to symlink currently and the limitations and risks.
6. Mentioned that container definitions are now auto-sorted


docs/remote/containers.md
```diff
@@ -1058,8 +1084,19 @@ If you've already built the container and connected to it, be sure to run **Remo
...
+| `remoteUser` | string | Overrides the user that VS Code runs as in the container (along with sub-processes like terminals, tasks, or debugging). Defaults to the user the container as a whole is running as (often `root`). |
+
+### Variables in attached container config files
+
+Variables can be referenced in certain string values in attached configuration files in the following format: **${variableName}**. The following is a list of available variables you can use.
+
+| Variable | Properties | Description |
+|----------|---------|----------------------|
+| `${localEnv:VAR_NAME}` | Any | Value of an environment variable (in this case,`VAR_NAME`) on your local machine. | All |
+| `${containerEnv:VAR_NAME}` | `remoteEnv` | Value of an existing environment variable inside the container (in this case, `VAR_NAME`) once it is up and running. For example: `"remoteEnv": { "PATH": "${containerEnv:PATH}:/some/other/path" }` |
```
chrmarti, line 1097: Actually only `containerEnv` works at the moment. Filed https://github.com/microsoft/vscode-remote-release/issues/2935.

docs/remote/containers.md
```diff
@@ -1070,42 +1107,31 @@ If you've already built the container and connected to it, be sure to run **Remo
...
-    ```
-
-    See [Resolving Dockerfile build failures](/docs/remote/troubleshooting.md#resolving-dockerfile-build-failures-for-images-using-debian-8) for a workaround.
 
 See the Docker troubleshooting guide for [Windows](https://docs.docker.com/docker-for-windows/troubleshoot) or [Mac](https://docs.docker.com/docker-for-mac/troubleshoot), consult [Docker Support Resources](https://success.docker.com/article/best-support-resources) for more information.
 
 ### Docker Extension limitations
 
-By default, the Docker extension will run remotely. While this is a sensible default for SSH and WSL 2, it means the extension will not show local containers when VS Code is connected to a container.
+By default, the Docker extension will run remotely. While this is a sensible default for SSH, it means the extension will not show local containers when VS Code is connected to a container.
```
chrmarti, line 1125: It runs locally for me. @egamma Was there a change that allows it to run locally when not installed remotely?

egamma: The Docker extension has the UI kind set to `["workspace", "ui"]` see https://github.com/microsoft/vscode-docker/blob/24b2554622fb9f70ee414b18c2e6583a51ad357b/package.json#L10.

This means if you have installed the Docker extension locally then it will run locally.
@sandy081 fixed this issue https://github.com/microsoft/vscode/issues/83675 and this some changes for May that allows you to install a `["workspace", "ui"]` extension that you have installed locally remotely.

Bottom line this paragraph is not correct. @sandy081 what is the easiest way to install the Docker extension remotely in the April version.

sandy081: > what is the easiest way to install the Docker extension remotely in the April version.

Unfortunately there is no easiest way to install Docker extension remotely if it is already installed locally other than uninstall and install. This will remove the Docker extension locally which is unwanted. I would suggest to mention the issue in the Documentation and say that this is fixed in the next release by providing an action in the extensions view to install the extension in remote.

docs/remote/containers.md
```diff
@@ -23,7 +23,7 @@ This lets VS Code provide a **local-quality development experience** — includi
 
 **Local:**
 
-* **Windows:** [Docker Desktop](https://www.docker.com/products/docker-desktop) 2.0+ on Windows 10 Pro/Enterprise. (Docker Toolbox is not supported. Windows container images are not supported.)
+* **Windows:** [Docker Desktop](https://www.docker.com/products/docker-desktop) 2.0+ on Windows 10 Pro/Enterprise. Windows 10 Home (2004+) requires Docker Desktop 2.2+ and the [WSL2 back-end](https://aka.ms/vscode-remote/containers/docker-wsl2). (Docker Toolbox is not supported. Windows container images are not supported.)
```
chrmarti, line 26: I had issues with Docker Desktop 2.2 and the WSL2 backend and ended up using 2.3 Edge. We have word that this will be released as stable sometime next week, so maybe say 2.3+ here for now.

docs/remote/containers.md
Chuxel: Yeah at one point it installed remotely by default (type `workspace`) - this paragraph is actually old.  The same thing is in the other extensions reference guides.

I didn't know we got in a position where you couldn't click and say "install remotely" anymore.

docs/remote/containers.md
Chuxel: Ah good to know.  Will make that fix.

docs/remote/containers.md
Chuxel: Yeah sounds good. I'd seen some chatter about that too.

PR #3637, opened 2020-05-01
Highlight inc. sync gotcha in server tutorial

jauco:
re #3611


dbaeumer:
@jauco actually there is now support for incremental text buffer change in the libraries so we need to address this in a different way. Do you want to look into this. It is fine if you say no.

gregvanl:
Closing old PR as the libraries have changed. Another PR would be required to update the tutorial.

PR #3640, opened 2020-05-01
correct example file name

jedgalbraith:
The new file created by the `tsc helloworld.ts` command is `helloworld.js` rather than `HelloWorld.js`.


gregvanl:
Thanks!

PR #3658, opened 2020-05-09
supportsMultipleEditorsPerDocument should be true

hediet:
`supportsMultipleEditorsPerDocument` should be set to true if the extension supports multiple editor instances for the same document, not `false` as the documentation stated.


mjbvz:
Thanks!

PR #3662, opened 2020-05-11
Added CELA-approved note about MSVC license

jureid:
"You can use the C++ Toolset from Visual Studio Build Tools along with Visual Studio Code to compile, build, and verify any C++ codebases as long as you also have a valid Visual Studio license (either Community, Pro or Enterprise) that you are actively using to develop these C++ codebases."


jureid:
@TylerMSFT proposed change to clarify ambiguity around VS Build Tools license agreement.

PR #3665, opened 2020-05-13
Docker Compose with multiple Compose files

ravipal:
Updated the compose document to show how user can compose up with multiple compose files.

### DIFF
--- a/docs/containers/docker-compose.md
+++ b/docs/containers/docker-compose.md
@@ -194,13 +194,13 @@ volumes:
 ```
 
 ## Docker Compose with multiple Compose files
-Workspaces can have multiple docker-compose files to handle different environments like development, test and production. The content of the configuration can be split into multiple files. For example, a base compose file that defines the common information for all environments and override files that defines environment specific information. When these files are passes as input to `docker-compose` command, it will combine these files into single configuration. By default the **Docker: Compose Up** command will pass a single file as input to compose command. But the compose command can be customized to pass in multiple files using [command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization). This can be also be achieved using [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) to invokes the `docker-compose` command with desired parameters.
+Workspaces can have multiple docker-compose files to handle different environments like development, test and production. The content of the configuration can be split into multiple files. For example, a base compose file that defines the common information for all environments and override files that defines environment specific information. When these files are passed as input to `docker-compose` command, it will combine these files into a single configuration. By default the **Docker: Compose Up** command will pass a single file as input to the compose command. But the compose command can be customized to pass in multiple files using [command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization). This can also be achieved using [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) to invokes the `docker-compose` command with desired parameters.
 
 Note: If your workspace has `docker-compose.yml` and `docker-compose.override.yml` and no other compose files, then the `docker-compose` command will be invoked with no input files and it will implicitly use these files. In this case, no customization is needed.
 ### Command Customization
-[Command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization) provides various ways to customize the compose up based on your requirement. The following are few sample command customization for compose up command.
+[Command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization) provides various ways to customize the compose up based on your requirement. The following are few sample command customization for the compose up command.
 #### Base file and an override file
-Let's assume your workspace has a base compose file (`docker-compose.yml`) and an override file for each environments (`docker-compose.dev.yml`, `docker-compose.test.yml` and `docker-compose.prod.yml`) and you always compose up with base file and an override file. In this case the compose up command can be customized like below. When the compose up command is invoked the ${configuration} will be replaced by the selected file.
+Let's assume your workspace has a base compose file (`docker-compose.yml`) and an override file for each environment (`docker-compose.dev.yml`, `docker-compose.test.yml` and `docker-compose.prod.yml`) and you always compose up with the base file and an override file. In this case the compose up command can be customized like below. When the compose up command is invoked the ${configuration} will be replaced by the selected file.
 ```json
 "docker.commands.composeUp": [
     {
@@ -210,7 +210,7 @@ Let's assume your workspace has a base compose file (`docker-compose.yml`) and a
 ]
 ```
 #### Template matching
-Let's assume you have different set of input files for each environments. You could define multiple templates with regular expression match, and the selected file name will be matched against this `match` property and corresponding template will be used.
+Let's assume you have  different set of input files for each environment. You could define multiple templates with regular expression match, and the selected file name will be matched against this `match` property and the corresponding template will be used.
 ```json
 "docker.commands.composeUp": [
     {
@@ -232,16 +232,16 @@ Let's assume you have different set of input files for each environments. You co
 ```
 
 #### Pick a template at runtime
-When invoking compose up command with the following command customization, will give an option to choose a template at runtime and the `${configurationFile}` will be replaced with the selected file.
+When invoking the compose up command with the following command customization, you will give an option to choose a template at runtime and the `${configurationFile}` will be replaced with the selected file.
 ```json
 "docker.commands.composeUp": [
     {
         "label": "dev",
-        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.test.yml ${configurationFile} up -d --build"
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.dev.yml ${configurationFile} up -d --build"
     },
     {
         "label": "test",
-        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.dev.yml ${configurationFile} up -d --build"
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.test.yml ${configurationFile} up -d --build"
     },
     {
         "label": "prod",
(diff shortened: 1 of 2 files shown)


docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
     - /host-folder-path:/container-folder-path
 ```
 
+## Docker Compose with multiple Compose files
+Workspaces can have multiple docker-compose files to handle different environments like development, test and production. The content of the configuration can be split into multiple files. For example, a base compose file that defines the common information for all environments and override files that defines environment specific information. When these files are passes as input to `docker-compose` command, it will combine these files into single configuration. By default the **Docker: Compose Up** command will pass a single file as input to compose command. But the compose command can be customized to pass in multiple files using [command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization). This can be also be achieved using [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) to invokes the `docker-compose` command with desired parameters.
```
BigMorty, line 197: passed

BigMorty, line 197: "passes" should be "passed".
"into single" should be "into a single"
"to compose command" should be "to the compose command"
"This can be also be achieved" should be "this can also be acheived"

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+
+Note: If your workspace has `docker-compose.yml` and `docker-compose.override.yml` and no other compose files, then the `docker-compose` command will be invoked with no input files and it will implicitly use these files. In this case, no customization is needed.
+### Command Customization
+[Command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization) provides various ways to customize the compose up based on your requirement. The following are few sample command customization for compose up command.
```
BigMorty, line 201: "command customization for compose up command" should be "command customizations for the compose up command"

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+#### Base file and an override file
+Let's assume your workspace has a base compose file (`docker-compose.yml`) and an override file for each environments (`docker-compose.dev.yml`, `docker-compose.test.yml` and `docker-compose.prod.yml`) and you always compose up with base file and an override file. In this case the compose up command can be customized like below. When the compose up command is invoked the ${configuration} will be replaced by the selected file.
```
BigMorty, line 203: "for each environments" should be "for each environment"
"with base file" should be "with the base file"

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+```json
+"docker.commands.composeUp": [
+    {
+        "label": "override",
+        "template": "docker-compose -f docker-compose.yml ${configurationFile}  up -d --build",
+    }
+]
+```
+#### Template matching
+Let's assume you have different set of input files for each environments. You could define multiple templates with regular expression match, and the selected file name will be matched against this `match` property and corresponding template will be used.
```
BigMorty, line 213: "have different" should be "have a different"
"each environments" should be "each environment"
"and corresponding" should be "and the corresponding"

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+    {
+        "label": "prod-match",
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.release.yml -f docker-compose.prod.yml up -d --build",
+        "match": "prod"
+    }
+]
+```
+
+#### Pick a template at runtime
+When invoking compose up command with the following command customization, will give an option to choose a template at runtime and the `${configurationFile}` will be replaced with the selected file.
```
BigMorty, line 235: "invoking compose up" should be "invoking the compose up"
"will give an option" should be "you will be given an option"

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+    },
+    {
+        "label": "prod",
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.prod.yml ${configurationFile} up -d --build"
+    },
+],
+```
+
+### Custom Tasks
+You could define a task like the following to invoke a `docker-compose` command. Please refer [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) for more detail on this.
```
BigMorty, line 254: I would start this out with... “Rather than use command customization you can also define a task…”

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
     - /host-folder-path:/container-folder-path
 ```
 
+## Docker Compose with multiple Compose files
+Workspaces can have multiple docker-compose files to handle different environments like development, test and production. The content of the configuration can be split into multiple files. For example, a base compose file that defines the common information for all environments and override files that defines environment specific information. When these files are passed as input to `docker-compose` command, it will combine these files into a single configuration. By default the **Docker: Compose Up** command will pass a single file as input to the compose command. But the compose command can be customized to pass in multiple files using [command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization). This can also be achieved using [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) to invokes the `docker-compose` command with desired parameters.
```
ghogen, line 197:
```suggestion
Workspaces can have multiple docker-compose files to handle different environments like development, test and production. The content of the configuration can be split into multiple files. For example, a base compose file that defines the common information for all environments and override files that defines environment-specific information. When these files are passed as input to the `docker-compose` command, it combines these files into a single configuration. By default, the **Docker: Compose Up** command passes a single file as input to the compose command, but you can customize the `compose up` command to pass in multiple files using [command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization). Or, you can use a [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) to invoke the `docker-compose` command with the desired parameters.
```

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+
+Note: If your workspace has `docker-compose.yml` and `docker-compose.override.yml` and no other compose files, then the `docker-compose` command will be invoked with no input files and it will implicitly use these files. In this case, no customization is needed.
```
ghogen, line 199:
```suggestion
Note: If your workspace has `docker-compose.yml` and `docker-compose.override.yml` and no other compose files, then the `docker-compose` command is invoked with no input files and it implicitly uses these files. In this case, no customization is needed.
```

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+### Command Customization
+[Command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization) provides various ways to customize the compose up based on your requirement. The following are few sample command customization for the compose up command.
```
ghogen, line 201:
```suggestion
[Command customization](https://code.visualstudio.com/docs/containers/reference#_command-customization) provides various ways to customize the `compose up` command based on your requirements. The following are few sample command customization for the `compose up` command.
```

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+#### Base file and an override file
+Let's assume your workspace has a base compose file (`docker-compose.yml`) and an override file for each environment (`docker-compose.dev.yml`, `docker-compose.test.yml` and `docker-compose.prod.yml`) and you always compose up with the base file and an override file. In this case the compose up command can be customized like below. When the compose up command is invoked the ${configuration} will be replaced by the selected file.
```
ghogen, line 203:
```suggestion
Let's assume your workspace has a base compose file (`docker-compose.yml`) and an override file for each environment (`docker-compose.dev.yml`, `docker-compose.test.yml` and `docker-compose.prod.yml`) and you always compose up with the base file and an override file. In this case the `compose up` command can be customized like below. When the compose up command is invoked the ${configuration} will be replaced by the selected file.
```

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+    {
+        "label": "prod-match",
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.release.yml -f docker-compose.prod.yml up -d --build",
+        "match": "prod"
+    }
+]
+```
+
+#### Pick a template at runtime
+When invoking the compose up command with the following command customization, you will give an option to choose a template at runtime and the `${configurationFile}` will be replaced with the selected file.
```
ghogen, line 235:
```suggestion
By invoking the `compose up` command with the following command customization, you can give an option to choose a template at runtime and the `${configurationFile}` will be replaced with the selected file.
```

ghogen: If that's right, I'm not sure who is giving the option to choose a template at runtime - the user?

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+    },
+    {
+        "label": "prod",
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.common.prod.yml ${configurationFile} up -d --build"
+    },
+],
+```
+
+### Custom Tasks
+Rather than use command customization you can also define a task like the following to invoke a `docker-compose` command. Please refer [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) for more detail on this.
```
ghogen, line 254:
```suggestion
Rather than use command customization, you can also define a task like the following to invoke a `docker-compose` command. Please refer [custom task](https://code.visualstudio.com/docs/editor/tasks#_custom-tasks) for more detail on this.
```

docs/containers/docker-compose.md
```diff
@@ -193,6 +193,76 @@ volumes:
...
+    },
+    {
+        "label": "prod-match",
+        "template": "docker-compose -f docker-compose.yml -f docker-compose.release.yml -f docker-compose.prod.yml up -d --build",
+        "match": "prod"
+    }
+]
+```
+
+#### Pick a template at runtime
```
karolz-ms, line 234: The meaning of "runtime" might be a bit unclear here--usually it is reserved for "application run time", not "VS Code run time". Consider something like this instead

### Pick a template when the command is invoked
If you omit the `match` property from command templates, you will be asked which template to use each time `compose up` command is invoked. For example:

karolz-ms: or "... VS Code will ask you which template to use when you invoke `compose up` command" (to avoid passive voice)

ravipal: Thanks! this is much cleaner.

PR #3668, opened 2020-05-14
Fix broken setup container link

rzgry:
"Setting up a folder to run in a container" link has an extra `-` which is breaking the link in the docs website. See: https://code.visualstudio.com/docs/remote/containers#_devcontainerjson-reference


gregvanl:
Thanks!

PR #3669, opened 2020-05-14
Clarify how "current workspace" is set

Cpcrook:
After spending part of a day troubleshooting why one of my devs was having the actual folder containing `.devcontainer` mounted, while everyone else was having the repository root folder mounted within their dev containers, we discovered that is because the "current workspace" changes to the repository root only if git is present on the user's `PATH` environment variable.  This is apparently the default value for the `workspaceMount` setting.  To be clear, however, the `${localWorkspaceFolder}` variable value does not change.

My dev (on a Windows host) was using git embedded in his terminal of choice, meaning vscode was unaware of the presence of a git.exe or the surrounding repo.

I think this is appropriate behavior (and we rely on it), but it needs to be documented properly so folks aren't going on a wild goose chase.

Let me know if you have any questions or how I can better explain this.

### DIFF
--- a/docs/remote/containers-advanced.md
+++ b/docs/remote/containers-advanced.md
@@ -173,7 +173,9 @@ Finally, if you've already built the container and connected to it, run **Remote
 
 ## Changing the default source code mount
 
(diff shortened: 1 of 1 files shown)


chrmarti:
LGTM, @Chuxel passing on to you for review and merging.

Chuxel:
LGTM! @gregvanl?

PR #3537, opened 2020-04-06
Updated mac.md for command palette shortcut

anotherlab:
Updated mac.md to show the correct keyboard MacOS shortcut to open the Command Palette.  Origin document had **F1** as the keyboard shortcut. The shortcut should have been **⇧⌘P**.

PR #3672, opened 2020-05-15
Fixed the command palette command

vegasbrianc:
The incorrect Command palette command was being used

### DIFF
--- a/docs/setup/mac.md
+++ b/docs/setup/mac.md
@@ -22,7 +22,7 @@ MetaDescription: Get Visual Studio Code up and running on Mac (macOS).
 You can also run VS Code from the terminal by typing 'code' after adding it to the path:
 
 - Launch VS Code.
-- Open the **Command Palette** (`kbstyle(F1)`) and type 'shell command' to find the **Shell Command: Install 'code' command in PATH** command.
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @vegasbrianc Thank you for your PR. Both keyboard shortcuts work for opening the Command Palette and we sometimes deliberately show F1 in documentation for discoverability.
See https://github.com/microsoft/vscode-docs/pull/3537

vegasbrianc:
Hi @gregvanl Thanks for responding so quickly to the PR. I think this needs a bit more clarification from a macOS user perspective as not all Mac's use Function keys the same. To use Function keys it is required to press `fn` + Function key unless the user has specifically enabled Function keys directly in `System Preferences -> Keyboard -> Use F1, F2, etc. as standard function keys`

Most Mac users, if they hit `F1` on it's own, will dim the screen ;)

In the other section Getting Started -> User interface it only reference command keys. Not sure if this will be a bit easier to understand for Mac users. - https://code.visualstudio.com/docs/getstarted/userinterface#_command-palette 

Hope this helps

gregvanl:
@vegasbrianc You've convinced me. Especially since a similar PR came in last month. Thanks

PR #3673, opened 2020-05-15
Update portable.md to include helpful details.

A-damW:
I was installing for Linux in portable mode, took me a few extra moments to notice there was no actual .zip download for Linux.
Added detail to indicate Linux download is a .tar.gz file.


gregvanl:
Thanks! Good clarification.

PR #3675, opened 2020-05-15
Fix grammar mistake

kahveci:
There is a grammar mistake at the conda environment definition:"....is a virtual environment**s** that's..."


gregvanl:
Thanks!

PR #3679, opened 2020-05-17
Update config-mingw.md

SpringHgui:
The default generated `label` setting may not be consistent with `preLaunchTask` setting


gregvanl:
Thanks! Nice mention in case tasks.json and launch.json are in sync.

PR #3682, opened 2020-05-19
Use npm init for scaffolding

felipecassiors:
This prevents the user to have 2 packages installed globally, which probably won't use anymore.


felipecassiors:
We can also use just `npm init yo code`, which seems to be better yet.

gregvanl:
Hi @felipecassiors Thank you for the PR. To keep things simple in this hello world tutorial, we recommend using npm since users may want to run Yeoman and the VS Code generator multiple times while they are learning and experimenting.

PR #3683, opened 2020-05-19
update example of reading a configuration value

mrlacey:
Change to something that works.
Assuming the example was how it used to be done but was no longer working.
I found what does work by looking at the [configuration sample](https://github.com/microsoft/vscode-extension-samples/blob/master/configuration-sample/src/extension.ts#L6).

### DIFF
--- a/api/references/contribution-points.md
+++ b/api/references/contribution-points.md
@@ -64,7 +64,7 @@ Contribute configuration keys that will be exposed to the user. The user will be
 
 ![configuration extension point example](images/contribution-points/configuration.png)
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @mrlacey The current example does work but perhaps could be clearer. You can use getConfiguration(), either without a string to get a configuration object including all settings for the session or you can scope it to just a setting namespace such as 'typescript' in the example. Your example will also work but you can do the same with:

```ts
var config = vscode.workspace.getConfiguration('myExtension`);
var value = config.get('keyName');
```

PR #3685, opened 2020-05-20
Show Remote Data is now Show Synced Data

JustinGrote:
This was changed in latest insiders and should be updated to reflect to avoid confusion.


gregvanl:
Thanks!

PR #3711, opened 2020-05-26
not clear on when initializeCommand runs

cicorias:
When reading this first, it seems as if `initializeCommand` would run only before container image being built. But it clearly runs both when first build of the image AND when re-attaching to an existing stopped container.  Maybe to others its clear, but to me it wasn't.

I'd actually would like a step that is "just before an image build" :)


Chuxel:
Since "attach" has special meaning in VS Code, perhaps the best thing to do would be to say when a "container is created or started"?

cicorias:
sure that does sound better as well. I adjusted the PR/File.

PR #3723, opened 2020-05-27
Fix for language-server-extension-guide.md

NelsonDaniel:
Update to the document so the package.json file contents mentioned reflect the structure of the sample provided at https://github.com/microsoft/vscode-extension-samples/blob/master/lsp-sample/client/package.json.


gregvanl:
Thanks!

Issue #3336, opened 2020-01-04
Consider using inclusive language in docs


KyleMit:
From [Words To Avoid in Educational Writing](https://css-tricks.com/words-avoid-educational-writing/), the following words can be difficult for folks who are just getting started and have a steep learning curve to implementing suggestions which may seem trivial to advanced users:

* Simply
* Just
* Obviously
* Basically
* Of course
* Clearly
* Easy

For example, here's some language choices in the [September 2018 Release Notes](https://code.visualstudio.com/updates/v1_28#_project-level-snippets):

> * ***Simply*** use a string array as the `prefix` property.
> * Returning the value `undefined` ***simply*** prevents the debug session from starting
> * ***Simply*** use the **Preferences: Configure User Snippets** command or create `*.code-snippets` file
> * Search for keybindings ***just*** by typing keystrokes
> * Now can you rename the referenced file ***just*** by renaming the import:
> * You can accept that current suggestion ***just*** by typing `;`

Very little is lost by removing those words altogether:

> * <del>Simply</del> Use a string array as the `prefix` property.
> * Returning the value `undefined` <del>simply</del> prevents the debug session from starting
> * <del>Simply</del> Use the **Preferences: Configure User Snippets** command or create `*.code-snippets` file
> * Search for keybindings <del>just</del> by typing keystrokes
> * Now can you rename the referenced file <del>just</del> by renaming the import:
> * You can accept that current suggestion <del>just</del> by typing `;`

Setting up language linting or including this in editorial reviews is definitely non-trivial, but just a suggestion to focus on inclusive and welcoming language.

gregvanl:
@KyleMit A great reminder to "omit needless words". 
We do try to remove words like "simply" and "just" from the VS Code document but also balance that goal with keeping a conversational style. Most of the VS Code documentation is written by the engineering team and they genuinely want the docs to be inclusive and welcoming. Words like "simply" and "just" are in some cases an attempt to assure the reader that the processes or steps aren't that cumbersome.
We use a linter, Acrolinx, which does flag words like "simply", but is run manually rather than as part of a CI build due to false positives.
PRs are always welcome if you'd like to contribute to the docs.

PR #3738, opened 2020-05-31
Made language easier for beginners

1502shivam-singh:


gregvanl:
Hi @1502shivam-singh Thank you for submitting a PR. We try to avoid using the word "simply" because what is simple or easy for one person might be challenging for another.
You can review https://github.com/microsoft/vscode-docs/issues/3336 for suggestions on inclusive writing.

PR #3748, opened 2020-06-04
Update step to change theme

HCloward:
I was updating my theme today and noticed the Preference option is not under File but Code. I was able to find it, but might reduce confusion and frustration to update the step.

### DIFF
--- a/docs/getstarted/themes.md
+++ b/docs/getstarted/themes.md
@@ -15,7 +15,7 @@ Color themes let you modify the colors in Visual Studio Code's user interface to
 
 ## Selecting the Color Theme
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @HCloward Thank you for submitting a PR. You are correct that the Preferences menu item is under Code for macOS. It is under the File menu for Windows and Linux and so we include both, with the macOS location in parentheses in step 1.

HCloward:
> Hi @HCloward Thank you for submitting a PR. You are correct that the Preferences menu item is under Code for macOS. It is under the File menu for Windows and Linux and so we include both, with the macOS location in parentheses in step 1.

oops, missed that - thanks!

PR #3751, opened 2020-06-06
Update vnext to be even with master

Chuxel:
Started doing doc updates and realized this hadn't been done yet.


gregvanl:
Thanks!

PR #3767, opened 2020-06-09
More updates

ItalyPaleAle:
Adding link at the bottom to the blog by the Go team (let me know if this is the best way to proceed)

Also, @asw101 has some more feedback coming in shortly

### DIFF
--- a/blogs/2020/06/09/go-extension.md
+++ b/blogs/2020/06/09/go-extension.md
@@ -42,4 +42,4 @@ Happy Coding!
 
 The VS Code Team
 
-PS: You can read the announcement from the Go team [here](https://blog.golang.org/vscode-go)
+P.S. You can also read the announcement from the Go team on [The Go Blog](https://blog.golang.org/vscode-go).
--- a/blogs/2020/06/09/go-extension.md
+++ b/blogs/2020/06/09/go-extension.md
@@ -41,3 +41,5 @@ From all of us in the Visual Studio Code and Go teams, we are excited for the fu
 Happy Coding!
 
 The VS Code Team
+
+PS: You can read the announcement from the Go team [here](https://blog.golang.org/vscode-go)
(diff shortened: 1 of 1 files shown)


blogs/2020/06/09/go-extension.md
```diff
@@ -41,3 +41,5 @@ From all of us in the Visual Studio Code and Go teams, we are excited for the fu
 Happy Coding!
 
 The VS Code Team
+
+PS: You can read the announcement from the Go team [here](https://blog.golang.org/vscode-go)
```
gregvanl, line 45: How about:
"P.S. You can also read the announcement from the Go team on [The Go Blog](https://blog.golang.org/vscode-go)."

Using [here] isn't good for screen readers, since it just say "here link" and people won't know where they are jumping off to.

ItalyPaleAle: Thanks for this, I didn't know that and I learnt something new and important :)

ItalyPaleAle:
Merged changes from @asw101 . Signed off from me now

blogs/2020/06/09/go-extension.md
```diff
@@ -10,9 +10,9 @@ Author: Alessandro Segala
 
 June 9, 2020 by The VS Code Team, [@code](https://twitter.com/code)
 
-Since the very beginning, we have worked with developer communities to build Visual Studio Code as a powerful, extensible editor. Five years into this journey, individuals and organizations in our community have helped us create an editor that is really for all developers, using any language.
+From the beginning, we have worked with developer communities to build Visual Studio Code as a powerful, extensible editor. Five years into this journey, individuals and organizations in our community have helped us create an editor that is really for all developers, using any language. VS Code is now the primary editor for many of our 11 million users.
```
gregvanl, line 13: How about: 
"VS Code is now the primary editor for many of the 11 million Go users."

gregvanl: Actually that is covered in the next paragraph. Do we even need this sentence?

ItalyPaleAle: No, because 11 million is VS Code users all-up, not just Go!

gregvanl: I see of the 11 million VS Code users, many use VS Code as their primary editor.

PR #3768, opened 2020-06-09
go-extension: use "Go Team at Google" instead of "Golang"

hyangah:
"golang" is the publisher ID and our display name should be "Go Team at Google".

I intended to update the screenshot of the extension that includes the publisher
name but it looks like that should be uploaded in a different location.


hyangah:
https://go-review.googlesource.com/c/blog/+/237259 is the change on our side. Thanks!

gregvanl:
Hi @hyangah I'll take a new screen shot.

gregvanl:
@hyangah Updated the image and published.

PR #3775, opened 2020-06-10
Additional guidance for container port assignment

karolz-ms:
We got some doc feedback that might indicate people have trouble going through the tutorial because of random container port assignment.

Also includes minor change reflecting UI exposed by latest Docker extension release

### DIFF
--- a/docs/containers/quickstart-aspnet-core.md
+++ b/docs/containers/quickstart-aspnet-core.md
@@ -38,9 +38,10 @@ In this guide you will learn how to:
 
    ![Add Dockerfile to a .NET project](images/quickstarts/aspnetcore-add-dotnet.png)
 
-1. Use **ASP.NET Core** when prompted for application platform.
+1. Use **.NET: ASP.NET Core** when prompted for application platform.
 1. Choose **Windows** or **Linux** when prompted to choose the operating system.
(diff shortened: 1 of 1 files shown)


docs/containers/quickstart-aspnet-core.md
```diff
@@ -103,6 +104,21 @@ You can use the Docker extension to author Docker files. The extension provides
     ]
    ```
 
+   > By default Docker will assign a randomly chosen **host port** to a port exposed by a container (the **container port**). In our application the exposed (container) port is 5000. When you issue **Run** command for an image, VS Code will try to use the same port number for the host port and container port. This makes it easy to remember which port to use to communicate with the container, but it won't work if the host port is already in use.
+   >
+   > If you cannot see the data from the container in your browser, make sure there are no errors reported by the `docker run` command (look at the command output in the terminal window). You can also verify which host port is using by the container by right-clicking the container in the Docker view and choosing **Inspect**. This will open a JSON document that describes the container in detail. Search for `PortBindings` element, for example:
+   >
+   > ```jsonc
```
ghogen, line 111: This looks like a typo. Should this be json ?

gregvanl: jsonc is the language identifier for JSON with Comments.

PR #3776, opened 2020-06-11
Corrected grammatical error in overview.md

enygmator:
Corrected grammatical error in overview.md. From 'is' to 'has' in line 12.


ornellaalt:
Thank you!

PR #3777, opened 2020-06-11
Update link to TS ESLint in 1.46 release notes

IllusionMH:
Currently link navigates to eslint.org that has only few mentions of TS.

Looking at the sentence context [Migrate from TSLint to ESLint](https://code.visualstudio.com/api/advanced-topics/tslint-eslint-migration) in VS Code docs looks more appropriate.

@mjbvz could you confirm that this was your intention or provide better link please.


mjbvz:
Yep, thanks!

PR #3782, opened 2020-06-12
Keybindings: Explain rules, then giv examples, rather than the other …

jhomme:
…way around.


gregvanl:
Thanks. Good changes.

PR #3790, opened 2020-06-18
Remove python.jediEnabled

fireattack:
in favor of python.languageServer, to match with June 16 2020 change Based on https://marketplace.visualstudio.com/items/ms-python.python/changelog


JoshuaPartlow:
Hi @fireattack - Thank you for opening a PR. We'll review it shortly and get back to you if any clarification is needed.

PR #3794, opened 2020-06-19
Update settings-sync.md

ProgramComputer:
This comment is the simplest solution. No need to install seahorse and go through the many steps.


gregvanl:
Thanks!

Issue #3784, opened 2020-06-15
Provide a notebook SDK


JacksonKearl:
Create documentation for creating notebook extensions, with helpful code samples and examples of how to create providers which perform common actions.

See webview/custom editor docs for examples.

rebornix:
Some inputs you might be interested in

* Background / Design
  * rendering https://github.com/microsoft/vscode/issues/88243

PR #3807, opened 2020-06-23
Notebook docs

JacksonKearl:
Ref #3784


api/extension-guides/notebook.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+## Parts of a Notebook
+
+A notebook consists of a sequence of cells and their outputs. The cells of a notebook can be either *markdown cells* or *code cells*, and are rendered within the core of VS Code. The outputs can be of a variety of formats. Some output formats, such as plain text, JSON, images,and HTML are rendered by VS Code core. Others, such as application-specific data or interactive applets, are rendered by extensions.
+
+Cells in a notebook are read and written to disk by a `NotebookContentProvider`, which handles reading data from disk and converting it into a description of cells, as well as saving modifications to the notebook back to disk. The *code cells* of a notebook can be exectuted by a `NotebookKernel`, which provides for taking a cell and from it producing output in a variety of formats ranging from plain text to formatted documents or interactive applets. Application-specific output formats and interactive applet outputs are rendered by an `NotebookOutputRenderer`.
```
rebornix, line 15: We may want to mention it once that the content can be read-from/write-to from all kinds of storage, in addition to disk.

api/extension-guides/notebook.md
```diff
@@ -0,0 +1,303 @@
...
+		)
+	);
+}
+
+class SampleProvider implements vscode.NotebookContentProvider {
+	async openNotebook(uri: vscode.Uri): Promise<vscode.NotebookData> {
+		const content = JSON.parse((await vscode.workspace.fs.readFile(uri)).toString());
+		return {
+			languages: [],
+			metadata: content.metadata,
```
rebornix, line 71:
```
metadata: {
  custom: content.metadata
}
```

api/extension-guides/notebook.md
```diff
@@ -0,0 +1,303 @@
...
+### Kernel
+[`NotebookKernel` API Reference](/api/references/vscode-api#NotebookKernel)
+
+A `NotebookKernel` is responsible for taking a *code cell* and from it producing some output or set of outputs.
+
+A kernel can either be directly associated with a content provider by setting the `NotebookContentProvider#kernel` property, or registered globally by envoking the `vscode.registerNotebookKernel` function with ...TODO: Figure out how this function is supposed to be called :)...
+
+Samples:
+- [GitHub Issues Notebook](https://github.com/microsoft/vscode-github-issue-notebooks/blob/master/src/notebookProvider.ts): Kernel to execute queries for GitHib Issues
+- [HTTP Request Notebook](): Kernel to issue HTTP requests (TODO: Publish my extension? Where?)
```
rebornix, line 119: Or maybe we file PR to the popular file based Rest Client https://marketplace.visualstudio.com/items?itemName=humao.rest-client

api/extension-guides/notebook.md
```diff
@@ -0,0 +1,303 @@
...
+- text/html
+- image/svg+xml
+- text/markdown
+- image/png
+- image/jpeg
+- text/plain
+- text/x-javascript
+
+To render an alternative mimetype, a `NotebookOutputRenderer` must be registered for that mimetype.
+
```
rebornix, line 200: We may also want to talk about multi-mimetype output and how extensions and users can control the display order of multi-mimetype output.

api/extension-guides/notebook.md
```diff
@@ -0,0 +1,303 @@
...
+Error outputs are helpful for displaying runtime errors in a consistant and understandable manner. They contain `ename` and `evalue` fields for displaying the error type and message, respectively, as well as `traceback` field which takes a list of strings which get displayed like a callstack:
+```ts
+{
+	outputKind: vscode.CellOutputKind.Error,
+	ename: 'Error Name',
+	evalue: 'Error Value',
+	traceback: ['stack frame 1', 'stack frame 2', 'stack frame 3', 'stack frame 4']
+}
+```
+![Cell with error output showing error name and message, as well as a stack trace](images/notebook/error-output.png)
```
rebornix, line 167: maybe we can show errors which have ansi colors.

api/extension-guides/notebook.md
JacksonKearl: Using "filesystem" as a more general form.

api/extension-guides/notebook.md
JacksonKearl: Thats a good idea, I put up a issue to discuss https://github.com/Huachao/vscode-restclient/issues/628

api/extension-guides/notebook.md
JacksonKearl: I gave an example above with 
```ts
{
	outputKind: vscode.CellOutputKind.Rich,
	data: {
		'text/html': '<b>Hello</b> World',
		'application/json': { hello: 'world' },
		'application/hello-world': 'my-hello-world-data-interchange-format',
	}
}
```
is that what you mean by multi-mimetype output? 

And how can extensions and users can control the display order of multi-mimetype output?

api/extension-guides/notebook.md
JacksonKearl: noted

api/extension-guides/notebook.md
rebornix: Firstly, we have builtin display orders

https://github.com/microsoft/vscode/blob/d8ef78152ce82d2e821ae2ef85ca08a54f89fa84/src/vs/workbench/contrib/notebook/common/notebookCommon.ts#L33-L53

and both notebook content providers and users can control that:

roblourens:
Minor point but it sounds like maybe we should be using `mediaType` or `contentType` instead of `mimeType` which might be an outdated term.

roblourens:
In the actual API I mean

JacksonKearl:
@roblourens that makes sense to me. Mayne open an issues in vscode?

api/extension-guides/notebook.md
```diff
@@ -360,4 +360,11 @@ Using the above renderer with a kernel that produces output in the `application/
 
 Samples:
  - [Notebook Renderer Starter](https://github.com/microsoft/notebook-renderer-starter): Starter code for dynamic renderers with support for debugging in the output context, webpack, hot reloading, and css modules.
- - [nteract Renderer](https://github.com/microsoft/notebook-extension-samples/tree/master/notebook-renderer): Example code for rendering [nteract](https://nteract.io/) outputs
\ No newline at end of file
+ - [nteract Renderer](https://github.com/microsoft/notebook-extension-samples/tree/master/notebook-renderer): Example code for rendering [nteract](https://nteract.io/) outputs
+
+## Supporting Debugging
+For some kernels, such as those that implement a programming language, it can be desirable to allow debugging a cell's execution. To accomplish this, a notebook kernel can implement a [debug adapter](https://microsoft.github.io/debug-adapter-protocol/), either by directly implementing the protocol, or providing an interface between an existing notebook debugger and the protocol.
```
roblourens, line 366: I think any kernel would be tied to a programming language. A provider might not be (markdown notebook etc) but I think any kernel would execute something right?

JacksonKearl: I have some that execute HTTP requests..

roblourens: Oh good point

JacksonKearl:
cc @gregvanl or @ornellaalt :)

rebornix:
> Minor point but it sounds like maybe we should be using mediaType or contentType instead of mimeType which might be an outdated term.

@roblourens @JacksonKearl I'm wondering if `mimeType` is a correct/better term here, which represents `type/subType` but without parameters (e.g., `text/html; charset=UTF-8`). Also `mimeType` is used in [Jupyter notebook](https://jupyter.readthedocs.io/en/latest/reference/mimetype.html) and VS Code code base ([Language Defintion](https://github.com/microsoft/vscode/blob/7a5bca9fafbbb23de289179be95e9c4832540c02/extensions/css/package.json#L25), [Emmet](https://github.com/microsoft/vscode/blob/7a5bca9fafbbb23de289179be95e9c4832540c02/extensions/css/package.json#L25), [Git](https://github.com/microsoft/vscode/blob/92164dd93d7d3f96dcc44ba6af994a62e5def354/extensions/git/src/api/git.d.ts#L155))

roblourens:
Ok I guess we should stick with it if that's what is typically used

Issue #3808, opened 2020-06-23
Add Editor Actions to the User Interface Docs


KyleMit:
Here's an example of the portion of the interface I'm referring to:

![editor-actions UI](https://user-images.githubusercontent.com/4307307/85452705-27ac0580-b569-11ea-99ca-e8a38f608bd9.png)

It's labelled as **Editor Actions** according to the Dev Tools window:

![editor actions dev tools](https://user-images.githubusercontent.com/4307307/85453522-f2ec7e00-b569-11ea-9f40-17f4d22bde30.png)

In the API reference, it's referenced via the [`contributes.menus`](https://code.visualstudio.com/api/references/contribution-points#contributes.menus) as `editor/title`

> The editor title menu bar 

However, I can't find it officially referred to in the [User Interface Docs](https://code.visualstudio.com/docs/getstarted/userinterface) with similar UI elements like breadcrumbs, status bar, etc.

Can it be documented there?

gregvanl:
Hi @KyleMit Thank you for creating an issue and PR. I'm not sure editor title bar actions need to be explicitly documented as title bar actions are common across most of the VS Code UI and readily discoverable by users.

KyleMit:
@gregvanl , that's a good point on inherent discoverability.  My particular use case was actually one of taxonomy.  Because I have lots of installed extensions, my editor actions became very cluttered. I was trying to find a way to disable them one by one within each extension.  Without knowing how to refer to that part of the UI, it was difficult to find the correct search terms to identify if the issue was already entered for other VS Code extensions.  

Without a clear, common, and consistent language term, it makes it hard to search and refer to that particular set of features.

PR #3809, opened 2020-06-23
#3808 add Editor Actions to UI page

KyleMit:
Adding for issue I opened for clarity on the Editor Actions portion of the UI - but copy change from maintainers certainly welcome


gregvanl:
Hi @KyleMit I'm going to close this PR as I don't think we need to explicitly document title bar actions and the new section, while correct and well-written, doesn't describe anything a user wouldn't readily discover.

PR #3812, opened 2020-06-24
Change apt key install directory

dlech:
On Debian-based systems, it is [recommended](https://www.debian.org/doc/packaging-manuals/fhs/fhs-3.0.html#ftn.idm236087550400) to not install local files outside of `/usr/local/` or `/etc/`, so we should not be manually installing files `/usr/share/`. Instead, we can use the existing `/etc/apt/trusted.gpg.d/` directory for apt archive keys.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -49,8 +49,8 @@ The repository and key can also be installed manually with the following script:
 
 ```bash
 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > packages.microsoft.gpg
-sudo install -o root -g root -m 644 packages.microsoft.gpg /usr/share/keyrings/
-sudo sh -c 'echo "deb [arch=amd64 signed-by=/usr/share/keyrings/packages.microsoft.gpg] https://packages.microsoft.com/repos/vscode stable main" > /etc/apt/sources.list.d/vscode.list'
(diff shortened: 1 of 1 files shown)


dlech:
According to the [sources.list](https://manpages.debian.org/buster/apt/sources.list.5.en.html) man page:

> *•Signed-By (signed-by)* is an option to require a repository to pass apt-secure(8) verification with a certain set of keys rather than all trusted keys apt has configured. It is specified as a list of absolute paths to keyring files (have to be accessible and readable for the _apt system user, so ensure everyone has read-permissions on the file) and fingerprints of keys to select from these keyrings. If no keyring files are specified the default is the trusted.gpg keyring and all keyrings in the trusted.gpg.d/ directory (see apt-key fingerprint). If no fingerprint is specified all keys in the keyrings are selected. A fingerprint will accept also all signatures by a subkey of this key, if this isn't desired an exclamation mark (!) can be appended to the fingerprint to disable this behaviour. The option defaults to the value of the option with the same name if set in the previously acquired Release file of this repository (only fingerprints can be specified there through). Otherwise all keys in the trusted keyrings are considered valid signers for this repository.

So if the key is put into `/etc/apt/trusted.gpg.d/`, we don't actually need to provide the `signed-by` option. It is the directory where `apt` expects keys to be by default.

PR #3817, opened 2020-06-24
-y flag to apt-get install

DrewDevereux:
`apt-get install` will fail (without a meaningful error message) in a Docker `RUN` command if it seeks `[Y|n]` input from the command line. We need to tell it not to ask.


Chuxel:
Good catch!!

PR #3823, opened 2020-06-26
Add context to contributes and deactivate

### LINKED ISSUE #3800: Extensions Documentation: Get Started
On https://code.visualstudio.com/api/get-started/your-first-extension, users are prompted to add to their extension by trying to "[Contribute](https://code.visualstudio.com/api/references/contribution-points) another command that displays current time in an information message." 
 
**Proposed change:** 
- Give users a bit more context into "Contribute"
     - As a first-time extension creator, the [contribution points article](https://code.visualstudio.com/api/references/contribution-points) was a long doc to get thrown into, and while it has great detail, I wasn't sure what to do. A sentence or two explaining what "contributes" means, or maybe an example showing the current time command they could contribute, might help.
     - Something like this explanation from [Extension Anatomy](https://code.visualstudio.com/api/get-started/extension-anatomy): 

> Contribution Points: static declarations that you make in the package.json Extension Manifest to extend VS Code.
 
On https://code.visualstudio.com/api/get-started/extension-anatomy, users are introduced to the concept of the `deactivate` method. However, its body is left blank, and throughout other samples, it seems to be left blank or excluded.
 
**Proposed change/question:**
- Are there any best practices or necessary code that should be included in deactivate? It could be great to see an explanation with a couple sentences, or a quick example of how to use it correctly in this doc. 
- I had found additional information about the method much later in the [References](https://code.visualstudio.com/api/references/activation-events) section: 

> Note: An extension must export an activate() function from its main module and it will be invoked only once by VS Code when any of the specified activation events is emitted. Also, an extension should export a deactivate() function from its main module to perform cleanup tasks on VS Code shutdown. Extension must return a Promise from deactivate() if the cleanup process is asynchronous. An extension may return undefined from deactivate() if the cleanup runs synchronously.

Perhaps we could include something like this explanation. Another possible location would be a comment in the source / template code generated for the extension itself.

bamurtaugh:
**Changes:** 
- Provide a bit more context about Contribution Points in https://code.visualstudio.com/api/get-started/your-first-extension
- Provide a bit more context about `deactivate` in https://code.visualstudio.com/api/get-started/extension-anatomy

Fixes #3800

### DIFF
--- a/api/get-started/extension-anatomy.md
+++ b/api/get-started/extension-anatomy.md
@@ -97,7 +97,7 @@ Each VS Code extension must have a `package.json` as its [Extension Manifest](/a
 
 ## Extension Entry File
 
-The extension entry file exports two functions, `activate` and `deactivate`. `activate` is executed when your registered **Activation Event** happens. `deactivate` gives you a chance to clean up before your extension becomes deactivated. For many extensions, explicit cleanup may not be required, and the `deactivate` method can be removed. But if an extension needs to perform an operation when VS Code is shutting down or the extension is disabled or uninstalled, this is the method to do so.
+The extension entry file exports two functions, `activate` and `deactivate`. `activate` is executed when your registered **Activation Event** happens. `deactivate` gives you a chance to clean up before your extension becomes deactivated. For many extensions, explicit cleanup may not be required, and the `deactivate` method can be removed. However, if an extension needs to perform an operation when VS Code is shutting down or the extension is disabled or uninstalled, this is the method to do so.
 
 The [`vscode`](https://www.npmjs.com/package/vscode) module contains a script located at `node ./node_modules/vscode/bin/install`. The script pulls the VS Code API definition file depending on the `engines.vscode` field in `package.json`. After running the script, you would get IntelliSense, jump to definition and other TypeScript language features in your code.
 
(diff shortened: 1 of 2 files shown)


api/get-started/extension-anatomy.md
```diff
@@ -97,7 +97,7 @@ Each VS Code extension must have a `package.json` as its [Extension Manifest](/a
 
 ## Extension Entry File
 
-The extension entry file exports two functions, `activate` and `deactivate`. `activate` is executed when your registered **Activation Event** happens. `deactivate` gives you a chance to clean up before your extension becomes deactivated.
+The extension entry file exports two functions, `activate` and `deactivate`. `activate` is executed when your registered **Activation Event** happens. `deactivate` gives you a chance to clean up before your extension becomes deactivated. For many extensions, explicit cleanup may not be required, and the `deactivate` method can be removed. But if an extension needs to perform an operation when VS Code is shutting down or the extension is disabled or uninstalled, this is the method to do so.
```
ornellaalt, line 100: Change "But" in last sentence to "However, "

bamurtaugh: Thanks for the feedback! Made the change.

ornellaalt:
Thank you!

PR #3825, opened 2020-06-26
update link to new location for cmake tools docs

TylerMSFT:


TylerMSFT:
#sign-off

gregvanl:
There is a reference to the old extension name in https://code.visualstudio.com/docs/cpp/c-cpp-properties-schema-reference#_configuration-properties

Do you want to update that as well with this PR.

PR #3832, opened 2020-07-01

devthepenguin:
### Fix `code` is not recognized as an internal command or keyword issue

Additional description for the initial step to modify the environment variable to find command `code` and correctly set up the development environment.

### DIFF
--- a/api/get-started/your-first-extension.md
+++ b/api/get-started/your-first-extension.md
@@ -14,6 +14,11 @@ In this topic, we'll teach you the fundamental concepts for building extensions.
 ```bash
 npm install -g yo generator-code
 ```
+Allow the terminal to recognize `code` command by modifying the environment variable:
+
+```
+setx Path "%Path%;C:\Users\USERNAME\AppData\Local\Programs\Microsoft VS Code\bin"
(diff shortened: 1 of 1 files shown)


api/get-started/your-first-extension.md
```diff
@@ -14,6 +14,11 @@ In this topic, we'll teach you the fundamental concepts for building extensions.
 ```bash
 npm install -g yo generator-code
 ```
+Allow the terminal to recognize `code` command by modifying the environment variable:
+
+```
+setx Path "%Path%;C:\Users\USERNAME\AppData\Local\Programs\Microsoft VS Code\bin"
```
devthepenguin, line 20: Add `code` to Path environment variable for the VS Code terminal to identify it for all future references

gregvanl:
Hi @devthepenguin Thank you for submitting a PR. We'd prefer that users review the setup instructions for their platform, rather than include those directions in the tutorial. https://code.visualstudio.com/docs/setup/mac.
Also setting the path for `code` shouldn't be required for Windows and Linux and may be confusing for users on those platforms.

PR #3842, opened 2020-07-05
wrong command

IAMRogerXi:
installing matplotlib in windows should be the following.

py -m pip install matplotlib


JoshuaPartlow:
Hi @IAMRogerXI - Thanks for creating a pull request on the docs. Looking at the change, the command with `python` should work on Windows, it just doesn't allow specifying a specific version. Is that the reason you're suggesting using the launcher instead?

JoshuaPartlow:
Closing since 4 days have passed. If you can clarify the above, please feel free to reactivate.

PR #3844, opened 2020-07-06
[cpp] Include extra info for Linux to Attach

WardenGnaw:
This PR adds the additional info to attach to a process in Linux with
GDB.

This information comes from
https://askubuntu.com/questions/41629/after-upgrade-gdb-wont-attach-to-process

@pieandcakes

### DIFF
--- a/docs/cpp/cpp-debug.md
+++ b/docs/cpp/cpp-debug.md
@@ -138,7 +138,13 @@ Windows:
 
 Linux:
 
-* GDB needs elevated permissions to attach to a process. When using *attach to process*, you need to provide your password before the debugging session can begin.
+* You may see an error saying: `ptrace: Operation not permitted`. This is due to GDB needing elevated permissions in order to attach to a process. This can be solved using the solutions below:
+    1. When using *attach to process*, you need to provide your password before the debugging session can begin.
+    1. To disable this error temporarily, use the following command:
+
+        `echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope`
+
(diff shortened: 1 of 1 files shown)


docs/cpp/cpp-debug.md
```diff
@@ -138,7 +138,13 @@ Windows:
...
 Linux:
 
-* GDB needs elevated permissions to attach to a process. When using *attach to process*, you need to provide your password before the debugging session can begin.
+* You may see an error saying: `ptrace: Operation not permitted`. This is due to GDB needing elevated permissions in order to attach to a process. This can be solved using the solutions below:
+    1. When using *attach to process*, you need to provide your password before the debugging session can begin.
+    1. To disable this error temporarily, use the following command:
+
+        `echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope`
+
+    1. To remove the error permanently, edit `/etc/sysctl.d/10-ptrace.conf` and modify `kernel.yama.ptrace_scope` value of `1` to `0`.
```
pieandcakes, line 147: this only works on Ubuntu. I don't think the sysctl.d exists in other distros. @andrewcrawley  as he knows more.

dannyalan:
Great job, thanks for updating this

docs/cpp/cpp-debug.md
andrewcrawley: Sysctl is a Linux kernel API, so it should be broadly applicable.  The specific path for the config files may vary, but I verified that (in addition to Ubuntu) Debian, CentOS and Arch all use this directory, so I'd expect these instructions to work on most distros.  However, the inclusion of a default config in "10-ptrace.conf" seems Ubuntu specific, so it might be better to just list the expected file contents and tell users to create or update as appropriate.

PR #3845, opened 2020-07-06
Docker Desktop not supported on VMs

karolz-ms:
We learned from Docker that installing Docker Desktop on virtual machines is not a supported configuration (although, technically, not a violation of the license agreement). As a result @BigMorty and I agreed with Docker product management that our documentation should recommend using Linux VMs and Docker Engine product if a VM is used for Docker development.

### DIFF
--- a/docs/containers/choosing-dev-environment.md
+++ b/docs/containers/choosing-dev-environment.md
@@ -76,6 +76,8 @@ Windows Subsystem for Linux represents a great choice for container-based servic
 
 The recommended way to enable container development with a remote machine is to do [a full Docker installation](https://docs.docker.com/install/) on the machine, including Docker daemon.
 
+>**NOTE**: Docker Desktop product is supported only on physical Windows and macOS machines, not virtual machines. If you want to use a virtual machine as a remote development environment, we recommend to use a Linux VM with [Docker Engine](https://docs.docker.com/engine/).
+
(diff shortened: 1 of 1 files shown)


karolz-ms:
> Are Hyper-V, Parallels,and Oracle VirtualBox all available for Linux?

Hyper-V is a Windows technology, Parallels is macOS. VirtualBox works on all 3 platforms. All can host Linux VMs of course.

BigMorty:
Gotcha!

docs/containers/choosing-dev-environment.md
```diff
@@ -76,6 +76,8 @@ Windows Subsystem for Linux represents a great choice for container-based servic
 
 The recommended way to enable container development with a remote machine is to do [a full Docker installation](https://docs.docker.com/install/) on the machine, including Docker daemon.
 
+>**NOTE**: Docker Desktop product is supported only on physical Windows and macOS machines, not virtual machines. If you want to use a virtual machine as a remote development environment, we recommend to use a Linux VM with [Docker Engine](https://docs.docker.com/engine/).
```
ghogen, line 79:
```suggestion
>**Note**: Docker Desktop product is supported only on physical Windows and macOS machines, not virtual machines. If you want to use a virtual machine as a remote development environment, we recommend to use a Linux VM with [Docker Engine](https://docs.docker.com/engine/).
```

PR #3857, opened 2020-07-09
Update Pipeline publishing steps to explicitly map secret variable

JeffreyCA:
The current instructions for publishing using Azure Pipelines will cause the publishing step to fail, with an error like:
```
TF400813: Resource not available for anonymous access. Client authentication required.
```

Since `VSCE_PAT` is set as a secret variable, it cannot be immediately referenced and used as an environment variable (according to the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#secret-variables)). One has to explicitly map the secret variable using the `env` block.

I've updated the YAML snippet to set the mapping for the `VSCE_PAT` environment variable.

### DIFF
--- a/api/working-with-extensions/continuous-integration.md
+++ b/api/working-with-extensions/continuous-integration.md
@@ -95,9 +95,11 @@ You can enable the build to run continuously when pushing to a branch and even o
 
 ```yaml
 trigger:
-  tags:
+  branches:
     include:
     - master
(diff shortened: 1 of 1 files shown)


joaomoreno:
Thanks! 🍻

PR #3858, opened 2020-07-09
Updates to flow of containers tutorial

bamurtaugh:
Rather than requiring users to clone one of the "try" repos and have preexisiting Git knowledge, updating containers tutorial to use the built-in "Try a sample" command. This flow feels smooth and avoids having to reopen a folder in a container or deal with a volume outright. 

Also updated the text and image related to the version of Node (updated from 10 -> 12).

Pulling the ["Try a dev container"](https://code.visualstudio.com/docs/remote/containers#_quick-start-try-a-dev-container) quick start out of the Container doc and here instead could also enable us to further shorten the overall documentation.

While there are 2 other scenarios (bind mounts and attaching to running containers), since we're planning to refactor the container docs to separate the scenarios a bit more, it seems like this intro tutorial can choose to focus on just the 1 path for users.


docs/remote/containers-tutorial.md
```diff
@@ -136,17 +128,25 @@ The `devcontainer.json` is basically a config file that determines how your dev
...
     "settings": {
         "terminal.integrated.shell.linux": "/bin/bash"
     },
+
+    // Add the IDs of extensions you want installed when the container is created in the array below.
+    "extensions": [
+        "dbaeumer.vscode-eslint"
+    ],
+    
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
```
Chuxel, line 142: Looks like an extra space or two got here.

bamurtaugh: You were too quick for me! I ended up removing the comments as you suggested originally since we have the table below the devcontainer.json definition explaining the various properties. I also updated that table to match the properties in the newer version of devcontainer.json we're now including. 

Let me know what you think of that update, or if you think the original version with comments and the other table was better.

Chuxel: That works too!

PR #3866, opened 2020-07-14
Fixed Using SSH keys example on Linux

ATGardner:
I am not %100 sure of this, but the example of adding the ssh-agent to my profile file didn't work as expected. And after I googled around, I found an almost identical example at https://code.visualstudio.com/docs/remote/troubleshooting#_setting-up-the-ssh-agent
The only difference was that it enters the condition if `$SSH_AUTH_SOCK` is empty, while the one here checks it is *not* empty (as far as I can tell, at least). Making this change got it to work for me.


gregvanl:
Thanks!

PR #3867, opened 2020-07-14
Add link to vscode remote - containers extension

guenp:
When reading these docs, I thought it would be nice to add a convenient link to the actual extension so users can just download it right away.


gregvanl:
Hi @guenp Thank you for submitting a PR. 
We'd like to leave the first extension link to be in the Installation section so that readers will have Docker configured and running before installing the Remove - Containers extension.

PR #3870, opened 2020-07-15
Refactor containers.md into several topics

gregvanl:
Containers.md content pulled out into:
Attach to a Running Container - docs/remote/attach-container.md - in TOC
Create a Development Container - docs/remote/create-dev-container.md - in TOC
devcontainer.json reference - docs/remote/devcontainerjson-reference.md - not in TOC

I also streamlined the Containers.md Getting Started content to make it less scary. Let me know if you think we need to keep in some of the caveats about Linux and Docker or if most users won't hit these.

I couldn't help myself but I took out some of the `-ing` in H2 headers. If you think many people have created heading bookmarks, I can revert those updates.

### DIFF
--- a/docs/remote/attach-container.md
+++ b/docs/remote/attach-container.md
@@ -1,7 +1,7 @@
-g---
+---
 Order: 9
 Area: remote
-TOCTitle: Attach to a Running Container
+TOCTitle: Attach to Container
 PageTitle: Attach to a running container using Visual Studio Code Remote Development
 ContentId: ed14ef07-f44c-4675-b95b-cb5faffc7abb
 MetaDescription: Attach to a running container using Visual Studio Code Remote Development
@@ -9,11 +9,11 @@ DateApproved: 7/9/2020
 ---
 # Attach to a running container
 
-While using Visual Studio Code to spin up a new container can be useful in many situations, it may not match your workflow and you may prefer to "attach" VS Code to an already running Docker container - regardless of how it was started. Once attached, you can install extensions, edit, and debug like you can when you open a folder in a container using [devcontainer.json](/docs/remote/devcontainerjson-reference.md).
+Visual Studio Code can create and start containers for you but that may not match your workflow and you may prefer to "attach" VS Code to an already running Docker container - regardless of how it was started. Once attached, you can install extensions, edit, and debug like you can when you open a folder in a container using [devcontainer.json](/docs/remote/devcontainerjson-reference.md).
 
 ## Attach to a Docker container
 
-To attach to a Docker container, either select **Remote-Containers: Attach to Running Container...** from the Command Palette (`kbstyle(F1)`) or use the **Remote Explorer** in the Activity Bar and select the **Connect to Container** inline action on the container you want to connect to.
+To attach to a Docker container, either select **Remote-Containers: Attach to Running Container...** from the Command Palette (`kbstyle(F1)`) or use the **Remote Explorer** in the Activity Bar and from the **Containers** view, select the **Attach to Container** inline action on the container you want to connect to.
 
 ![Containers Explorer screenshot](images/containers/containers-attach.png)
 
--- a/docs/remote/attach-container.md
+++ b/docs/remote/attach-container.md
@@ -1,0 +1,79 @@
+g---
+Order: 9
+Area: remote
+TOCTitle: Attach to a Running Container
+PageTitle: Attach to a running container using Visual Studio Code Remote Development
+ContentId: ed14ef07-f44c-4675-b95b-cb5faffc7abb
+MetaDescription: Attach to a running container using Visual Studio Code Remote Development
+DateApproved: 7/9/2020
+---
+# Attach to a running container
+
+While using Visual Studio Code to spin up a new container can be useful in many situations, it may not match your workflow and you may prefer to "attach" VS Code to an already running Docker container - regardless of how it was started. Once attached, you can install extensions, edit, and debug like you can when you open a folder in a container using [devcontainer.json](/docs/remote/devcontainerjson-reference.md).
+
+## Attach to a Docker container
+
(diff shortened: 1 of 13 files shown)


docs/remote/containers.md
```diff
@@ -73,34 +69,18 @@ To get started, follow these steps:
 
 **Next Steps:** The Remote - Containers extension supports two primary operating models:
```
bamurtaugh, line 70: I was looking for "Next Steps" at the bottom of the page and wasn't anticipating to find them here. While it does seem to fit into the flow here, should we also add it somewhere to the bottom of the doc?

docs/remote/attach-container.md
```diff
@@ -0,0 +1,79 @@
...
+---
+# Attach to a running container
+
+While using Visual Studio Code to spin up a new container can be useful in many situations, it may not match your workflow and you may prefer to "attach" VS Code to an already running Docker container - regardless of how it was started. Once attached, you can install extensions, edit, and debug like you can when you open a folder in a container using [devcontainer.json](/docs/remote/devcontainerjson-reference.md).
+
+## Attach to a Docker container
+
+To attach to a Docker container, either select **Remote-Containers: Attach to Running Container...** from the Command Palette (`kbstyle(F1)`) or use the **Remote Explorer** in the Activity Bar and select the **Connect to Container** inline action on the container you want to connect to.
+
+![Containers Explorer screenshot](images/containers/containers-attach.png)
```
bamurtaugh, line 18: We could update this image since the Remote Explorer now includes a drop down to choose between WSL, containers, or SSH, and the VS Code debug icon is different as well.

docs/remote/attach-container.md
```diff
@@ -0,0 +1,79 @@
+g---
```
Chuxel, line 1: There's an extra "g" here that I assume will cause trouble once it lands in the docs site

docs/remote/attach-container.md
```diff
@@ -0,0 +1,79 @@
...
+Area: remote
+TOCTitle: Attach to a Running Container
+PageTitle: Attach to a running container using Visual Studio Code Remote Development
+ContentId: ed14ef07-f44c-4675-b95b-cb5faffc7abb
+MetaDescription: Attach to a running container using Visual Studio Code Remote Development
+DateApproved: 7/9/2020
+---
+# Attach to a running container
+
+While using Visual Studio Code to spin up a new container can be useful in many situations, it may not match your workflow and you may prefer to "attach" VS Code to an already running Docker container - regardless of how it was started. Once attached, you can install extensions, edit, and debug like you can when you open a folder in a container using [devcontainer.json](/docs/remote/devcontainerjson-reference.md).
```
Chuxel, line 12: "spin up a new container" could be a reference to the other containers article since if you jump right here you may not know what that means

docs/remote/create-dev-container.md
```diff
@@ -0,0 +1,547 @@
...
+```text
+📁 github.com
+    📁 microsoft
+        📁 vscode-dev-containers
+           📁 .devcontainer
+```
+
+Once in place, the configuration will be automatically picked up when using any of the remote containers commands. Once in the container, you can also select **Remote-Containers: Open Container Configuration File** from the Command Palette (`kbstyle(F1)`) to open the related `devcontainer.json` file and make further edits.
+
+## Configure a sandbox for multiple projects or folders
```
Chuxel, line 118: This section could also move to advanced containers instead of here. Its somewhat like the multi-container article there in that it is a variation on a theme established in the main containers article.

docs/remote/containers.md
```diff
@@ -192,225 +172,7 @@ Note that if the container fails to come up due to something like a Docker build
...
-
-    ![Open python folder in the container](images/containers/open-folder-python.png)
-
-6. In a moment, VS Code will open the folder inside the same container. In the future, you can use the **Remote Explorer** in the Activity Bar to open this sub-folder in the container directly.
-
-    ![Container explorer with multiple folders under python container](images/containers/containers-explorer-python.png)
-
-> **Tip:** Instead of mounting the local filesystem, you can use a similar flow to set up a container with an isolated, more performant volume that you clone your source code into. See the [Advanced Containers](/docs/remote/containers-advanced.md#use-a-named-volume-for-your-entire-source-tree) article for details.
-
-## Creating a devcontainer.json file
```
Chuxel, line 244: Reading through this, we really don't talk about devcontainer.json in this article until i starts to be referenced in later sections on extensions, etc. It feels a bit disjointed given its a central concept on first pass.

I wonder if this introduction could stay to help bridge the gap.  The "for example" part of this could then point to the other article.

docs/remote/containers.md
```diff
@@ -1067,86 +450,6 @@ See the [Advanced Container Configuration](/docs/remote/containers-advanced.md)
 * [Developing inside a container on a remote Docker Machine or SSH host](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host)
 * [Reducing Dockerfile build warnings](/docs/remote/containers-advanced.md#reducing-dockerfile-build-warnings)
 
-## devcontainer.json reference
```
Chuxel, line 1070: This is a common jump point - so I wonder if we should keep the heading so it shows up on the right nav but then just point to the dedicated topic. This is also one I'd expect to see people directly linking to - unless we can meta redirect.

For example, it is referenced in the GitHub docs: https://docs.github.com/en/github/developing-online-with-codespaces/configuring-codespaces-for-your-project

gregvanl: Yep, agreed. I haven't been able to sort out a redirect and I found a link in the docs.microsoft.com Codespaces docs as well.  I also think more visibility (right nav) will be helpful. I'm reconsidering having the devcontainer.json reference be hidden and probably good to have it visible in the TOC.

PR #3865, opened 2020-07-14
Update command.md -  Custom When Clause Context for Extension Developers

warrenbuckley:
This will replace PR #3852 - cc @gregvanl

PR #3872, opened 2020-07-16
Fix up my code typo from my PR #3865

warrenbuckley:
CC @gregvanl went to use my own documentation today & then notice the typo in the code, note the **`** that is not needed in the string

Fixes typo introduced from my PR #3865


gregvanl:
Thanks, I missed that as well.

PR #3873, opened 2020-07-17
Make it clear where to type the word Deployment

JohnRusk:
Since without this clarification, it may be assumed that the instructions are talking about something on the the command palette.  Or it may be attempted in a file with an extension that does not trigger the autocompletion.


gregvanl:
Thanks!

PR #3901, opened 2020-07-22
Docker extension: improve node quickstart

ghogen:


docs/containers/quickstart-node.md
```diff
@@ -35,16 +35,28 @@ In this guide you will learn how to:
...
+The extension creates `Dockerfile` and `.dockerignore` files. If you elected to include Docker Compose files, `docker-compose.yml` and `docker-compose.debug.yml` will be generated as well. Finally, the extension will create a set of **VS Code tasks** in `.vscode/tasks.json` for building and running the container (in both debug- and release-configurations) and a **launch debug configuration** in `.vscode/launch.json` for debugging the service within the container.
+
+## Add an environment variable to the image
+
+The Docker extension helps you author Dockerfiles by using [IntelliSense](../editor/intellisense.md) to provide auto-completions and contextual help. To see this feature in action, add an environment variable to your service image by following these steps:
+
+1. Open the `Dockerfile` file.
+1. Use `ENV` instruction to add an environment variable to the service container image. The instruction should be placed in the `base` stage of the `Dockerfile` (the first stage in the file).
```
karolz-ms, line 52: The part about using the base stage of the Dockerfile applies to .NET, where we start from that stage to run a special version of the container for debugging, file-mapping debugger and application bits into the container.

All this wizardry is not necessary for Node which has a debugger (they call it "inspector") built-in into the runtime. Please double-check, but I do not think our extension scaffolds multi-stage Dockerfile for node. If my memory is right, none of the information referring to the base stage in the Dockerfile would apply to node, i.e. to this quickstart

docs/containers/quickstart-node.md
```diff
@@ -89,18 +101,29 @@ When the Docker extension adds files to the application, it also adds a **VS Cod
...
-    - The debugger hits the breakpoint in `index.js`.
+    - The debugger stops at the breakpoint in `index.js`.
 
     > Note that, because the debugger attaches *after* the application starts, the breakpoint may missed the first time around; you might have to refresh the browser to see the debugger break on the second try.
     >
     > You can configure the application to wait for the debugger to attach before starting execution by setting the [inspectMode](/docs/containers/reference.md#node-object-properties-docker-run-task) property to `break` in the `docker-run: debug` task in `tasks.json` under the `node` object.
 
+## Use the Docker view
+
+The Docker view provides an interactive experience to examine and manage Docker assets such as containers, images and so on. To see an example:
```
karolz-ms, line 118: Docker view is referenced in the "Run the service container" paragraph above. This paragraph sounds a bit as if we were assuming the user is encountering Docker view for the first time. Thus it feels a bit out of place

ghogen:
Good to merge now.

PR #3905, opened 2020-07-25
Fix wrong link

DAXaholic:
The list items 'Opening existing source code in a container' and 'Working with a GitHub repository' both refer to the same place in the documentation where it is described how to open an existing folder in a container.
However, there is a dedicated section about how to open a Git repository or GitHub PR which is probably more appropriate for the last list item.


gregvanl:
Thanks! Looks like a copy and paste error.

PR #3909, opened 2020-07-28
Add VS Code to PATH in .zshrc (Catalina default shell)

mcsage:
Add VS Code to PATH in .zshrc (Catalina default shell)

### DIFF
--- a/docs/setup/mac.md
+++ b/docs/setup/mac.md
@@ -43,6 +43,15 @@ Start a new terminal to pick up your `.bash_profile` changes.
 
 **Note**: The leading slash `\` is required to prevent `$PATH` from expanding during the concatenation. Remove the leading slash if you want to run the export command directly in a terminal.
 
+**Note**: Since `zsh` became the default shell in macOS Catalina, run the following commands to add VS Code to your path:
+
+```zsh
+cat << EOF >> ~/.zshrc
+# Add Visual Studio Code (code)
+export PATH="/Applications/Visual Studio Code.app/Contents/Resources/app/bin:$PATH"
+EOF
+```
+
 ## Touch Bar support
 
 Out of the box VS Code adds actions to navigate in editor history as well as the full Debug tool bar to control the debugger on your Touch Bar:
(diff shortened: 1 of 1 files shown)


docs/setup/mac.md
```diff
@@ -43,6 +43,15 @@ Start a new terminal to pick up your `.bash_profile` changes.
...
 **Note**: The leading slash `\` is required to prevent `$PATH` from expanding during the concatenation. Remove the leading slash if you want to run the export command directly in a terminal.
 
+**Note**: Since `zsh` became the default shell in macOS Catalina, run the following commands to add VS Code to your path:
+
+```zsh
+cat << EOF >> ~/.zshrc
+# Add Visual Studio Code (code)
+export PATH="/Applications/Visual Studio Code.app/Contents/Resources/app/bin:$PATH"
+EOF
+```
```
Tyriar, line 53: Instead of duplicating everything the note could call out to replace `.bash_profile` with `.zshrc`, `.profile`, etc. Also is there a reason you don't use `.zprofile`?

mcsage: The reason not adding this to `.zprofile` is that it is sufficient to have it in an interactive shell instead of sourcing it in a login shell. I suggest changing `.bash_profile` to `.bashrc` and adding a note this could replace with `.zshrc` if that's the default shell.

Tyriar: MacOS terminals by default run in a login shell though, not on actual login? I'm pretty sure I've had to investigate issues with scripts in `.bashrc` not being run in the terminal because it wasn't being sourced from `.bash_profile`, it seems safer to leave it as is with the profile.

Tyriar:
Thanks!

mcsage:
Hi @Tyriar, I still recommend using `.zshrc` because also other popular project, (like docker, kubernetes, oh-my-zsh) using `.zshrc` in their docs to add alias or extend PATH variable.

Tyriar:
@mcsage changed it back since you feel strongly and I don't use my mac much these days 🙂

PR #3911, opened 2020-07-28
remove mention to (deprecated) tslint extension

dimitropoulos:
since eslint is the successor to tslint, it might be best to just remove this from the docs to keep things current with the times

### DIFF
--- a/docs/editor/extension-gallery.md
+++ b/docs/editor/extension-gallery.md
@@ -238,7 +238,7 @@ An example `extensions.json` could be:
 }
 ```
 
-which recommends two linter extensions: ESLint, and the Chrome debugger extension.
+which recommends a linter extension, ESLint, and the Chrome debugger extension.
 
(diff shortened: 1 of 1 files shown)


docs/editor/extension-gallery.md
```diff
@@ -232,14 +232,13 @@ An example `extensions.json` could be:
...
     "recommendations": [
-        "ms-vscode.vscode-typescript-tslint-plugin",
         "dbaeumer.vscode-eslint",
         "msjsdiag.debugger-for-chrome"
     ]
 }
 ```
 
-which recommends two linter extensions, TSLint and ESLint, as well as the Chrome debugger extension.
+which recommends two linter extensions: ESLint, and the Chrome debugger extension.
```
gregvanl, line 241: "which recommends a linter extension, ESLint, and the Chrome debugger extension."

ornellaalt:
Thank you!

PR #3927, opened 2020-08-07
Fix port to be number

bsivanov:
`port` in debug configurations should be number and not string, otherwise 'Incorrect type. Expected "number".' error is shown.


gregvanl:
Thanks!

PR #3944, opened 2020-08-14
Update ssh.md with more clarity on settings overrides

studgeek:
Make it clearer that Remote specifically overrides User settings and that Workspace overrides both.


gregvanl:
Thanks, nice clarification

Issue #3889, opened 2020-07-21
Updates to Working with GitHub page


ornellaalt:
On https://code.visualstudio.com/docs/editor/github:
- Explain how to use SSH keys, or link out to reference
- Add details about authenticating to GitHub & different methods
- Add information about manually providing authentication token if authenticating with an existing repository doesn't work automatically

PR #3945, opened 2020-08-14
Update github.md

ornellaalt:
Fixes https://github.com/microsoft/vscode-docs/issues/3889

### DIFF
--- a/docs/editor/github.md
+++ b/docs/editor/github.md
@@ -19,7 +19,7 @@ Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.
 
 ![Extension Sign In](images/github/extension-signin.png)
 
-If you are not redirected to VS Code, you can add your authorization token manually. In the browser window, you will receive your authorization token. Copy the token, and switch back to VS Code. Select **Signing in to github.com...** in the Status bar, paste the token, and hit **Enter**.
+If you are not redirected to VS Code, you can add your authorization token manually. In the browser window, you will receive your authorization token. Copy the token, and switch back to VS Code. Select **Signing in to github.com...** in the Status bar, paste the token, and hit (`kbstyle(Enter)`).
 
 ## Setting up a repository
 
--- a/docs/editor/github.md
+++ b/docs/editor/github.md
@@ -19,6 +19,8 @@ Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.
 
 ![Extension Sign In](images/github/extension-signin.png)
 
+Note that there are several ways to authenticate to GitHub, including using your username and password with two-factor authentication (2FA), a personal access token, or an SSH key. See [About authentication to GitHub](https://docs.github.com/en/github/authenticating-to-github/about-authentication-to-github) for more information and details about each option. 
+
 ## Setting up a repository
 
 ### Cloning a repository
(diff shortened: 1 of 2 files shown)


docs/editor/github.md
```diff
@@ -19,6 +19,8 @@ Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.
 
 ![Extension Sign In](images/github/extension-signin.png)
 
+Note that there are several ways to authenticate to GitHub, including using your username and password with two-factor authentication (2FA), a personal access token, or an SSH key. See [About authentication to GitHub](https://docs.github.com/en/github/authenticating-to-github/about-authentication-to-github) for more information and details about each option.
```
RMacfarlane, line 22: I'm not sure if this is useful to include here. While this is true, when logging in through VSCode, we always open the browser and have the user login there.

docs/editor/github.md
```diff
@@ -35,7 +37,7 @@ When you do something that requires GitHub authentication, you'll see a prompt t
 
 ![Authentication Prompt](images/github/auth-prompt.png)
 
-Follow the steps to sign into GitHub and return to VS Code.
+Follow the steps to sign into GitHub and return to VS Code. If authenticating with an existing repository doesn’t work automatically, you may need to manually provide a personal access token. See [Creating a personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) for more information.
```
RMacfarlane, line 40: It would also be good to note that to enter the personal access token, the user should click on the "Signing into GitHub..." status bar item in VS Code that appears when they start the sign in process

RMacfarlane: We could also note that for the GitHub Pull Requests and Issues extension, the personal access token should be created with scopes `read:user`, `user:email`, and `repo`. https://github.com/microsoft/vscode-pull-request-github/wiki#personal-access-token-authentication

docs/editor/github.md
ornellaalt: Once they get to the browser, does the username and password always work? The feedback we received was around providing more information about SSH and other ways to authenticate. 

Is there another doc where it makes sense to include this reference link?

RMacfarlane: It sometimes fails on Linux, which isn't able to reopen VSCode afterwards to complete the process. But it should allow 2FA.

Maybe this would make sense to move below to the change about authentication not working automatically.

docs/editor/github.md
```diff
@@ -19,6 +19,8 @@ Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.
 
 ![Extension Sign In](images/github/extension-signin.png)
```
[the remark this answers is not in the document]
ornellaalt: Fixed!

docs/editor/github.md
```diff
@@ -15,10 +15,12 @@ If you're new to source control and want to start there, you can learn about VS
 
 ## Getting started with GitHub Pull Requests and Issues
 
-Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) extension, you'll need to sign in. Follow the prompts to authenticate with GitHub and return to VS Code.
+Once you've installed the [GitHub Pull Requests and Issues](https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-pull-request-github) extension, you'll need to sign in. Follow the prompts to authenticate with GitHub in the browser and return to VS Code.
 
 ![Extension Sign In](images/github/extension-signin.png)
 
+If you are not redirected to VS Code, you can add your authorization token manually. In the browser window, you will receive your authorization token. Copy the token, and switch back to VS Code. Select **Signing in to github.com...** in the Status bar, paste the token, and hit **Enter**.
```
gregvanl, line 22: Use `kbstyle(Enter)` for keys rather than bold.

PR #3963, opened 2020-08-25
Remove confusion about .NET Core Docker Config Deprication

MarcusOy:
In VSCode, creating a new launch config within launch.json with the template name Docker: .NET Core Attach (Preview) will create a config listed in the .NET Core section. I believe the quote saying "previous (Preview)" is deprecated is ambiguous, since the documentation doesn't say that the current version of the config available as a launch config template is still labeled as a preview. This let me to believe that the config I had added within my launch.json was the deprecated implementation.

I propose this simple change to clear up my original confusion.
See comment: https://github.com/microsoft/vscode-docker/issues/2248#issuecomment-680043505

### DIFF
--- a/docs/containers/debug-common.md
+++ b/docs/containers/debug-common.md
@@ -78,7 +78,7 @@ Example `launch.json` configuration for debugging a Python application:
 
 More information about debugging .NET Core applications within Docker containers can be found in [Debug .NET Core within Docker containers](/docs/containers/debug-netcore.md).
 
-> The previous (Preview) .NET Core Docker debugging support is being deprecated. You can still find documentation on that support at [Debug .NET Core - Deprecated](https://github.com/microsoft/vscode-docker/wiki/Debug-NetCore-Deprecated).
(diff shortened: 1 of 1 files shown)


ghogen:
@BigMorty @karolz-ms Does this change look good to you?

BigMorty:
Looks good to me. @bwateratmsft shouldn't we remove the "(preview)" label on the new debug config?

bwateratmsft:
> Looks good to me. @bwateratmsft shouldn't we remove the "(preview)" label on the new debug config?

@BigMorty I'm reluctant to because I feel the experience of getting the debugger into the container is still not good. If it's already downloaded on the host *and* volume mapped into the container it works reasonably well, but otherwise it must be copied in which involves annoying prompts and takes a lot of time.

I'd really like to consider producing a .NET dev-container-ish image that includes vsdbg built in. Unfortunately that would mean a pretty big bump in size...but I'd also like to see vsdbg get smaller. I think one of the focus points of .NET 5 is to make self-contained executables much smaller so that might solve that problem.

PR #3966, opened 2020-08-27
Refactor "Create a development container" doc

bamurtaugh:
Based on prior discussions, I've made an initial pass at refactoring the [Create a development container](https://code.visualstudio.com/docs/remote/create-dev-container) doc. The goal is to create more of a "happy path," especially for users less familiar with containers or Docker, and to fill in more of the middle ground from "Hello World" pre-configured container -> advanced containers.

Proposed path:
- Create devcontainer.json
- Install software in container. Try via command line, notice it doesn't persist. Try postCreateCommand, but also realize there's a more efficient way.
- Create Dockerfile
- Building container/edit configuration loop
- Docker Compose

I believe the main section I removed from the current doc is [Set up a folder to run in a container](https://code.visualstudio.com/docs/remote/create-dev-container#_set-up-a-folder-to-run-in-a-container). It felt like the content it discussed (i.e. image directly in devcontainer.json vs using a Dockerfile) was now mentioned in the updates, but I also wasn't sure where it would fit into the story at large or if it was a necessary part of this updated flow. There is other content I removed or reworded as well.

I'd really like to hear feedback about these updates, such as the path chosen, sections added/removed, if anything should be further parred down (either deleted completely or moved to the [Advanced Containers](https://code.visualstudio.com/docs/remote/containers-advanced) doc or [Tips and Tricks](https://code.visualstudio.com/docs/remote/troubleshooting)), the level of the content, any wording, and the overall flow and readability.

cc @burkeholland @Chuxel @gregvanl @chrmarti


docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
+If `devcontainer.json`'s supported workflows do not meet your needs, you can also [attach to an already running container instead](/docs/remote/attach-container.md).
 
-To get started quickly, **open the folder** you want to work with in VS Code and run the **Remote-Containers: Add Development Container Configuration Files...** command in the Command Palette (`kbstyle(F1)`).
+> **Tip:** Want to use a remote Docker host? See the [Advanced Containers article](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) for details on setup.
 
-![Select Dockerfile](images/containers/select-dockerfile.png)
+With this `devcontainer.json`, your dev container is functional, and you can connect to and start developing within it. Try it out with the **Remote-Containers: Reopen Folder in Container** command (or **Remote-Containers: Reopen in Container** if you're just working with a single app file):
```
Chuxel, line 68: I think we are just now showing "Reopen in Container", so this is a bit out of date.

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
-        "--security-opt",
-        "seccomp=unconfined"
-    ]
+    "forwardPorts": [ 3000 ],
+    "postCreateCommand": "sudo apt-get update && sudo apt-get install git",
 }
 ```
 
-While less efficient than a custom Dockerfile, you can also use the `postCreateCommand` property to **install additional software** that **may not be in your base image** or for cases where you would prefer **not to modify a deployment Dockerfile** you are reusing.
+The `postCreateCommand` is run once the container is running, so you can also use the property to run commands like `npm install` or to execute a shell script in your source tree (if you have mounted it).
```
Chuxel, line 126: In reflecting back, two tips I'm wondering if we should add here is that:

1. There's also a postStartCommand which executes every time
2. The command needs to exit or the container won't start.

I've seen bugs logged and we saw Dina hit really both of these by adding an application start to postCreateCommand.

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
 
-For example, here is a `devcontainer.json` that adds compiler tools and the [C++](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools) extension to the base Ubuntu 18.04 container image:
+## Dockerfile
+
+A Dockerfile will also live in the `.devcontainer` folder. You can replace the `image` property in `devcontainer.json` with `dockerFile`:
 
 ```json
 {
-    "image": "ubuntu:18.04",
+    "dockerFile": "Dockerfile",
```
Chuxel, line 140: I'm thinking of all sorts of things re-reading this. 😀  I wonder if we should use the new "build": { "dockerfile": "Dockerfile" } syntax here.  That's where things like "args" live.  A bit more verbose, but would set people up for success later.

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
-
-This command is run once your source code is mounted, so you can also use the property to run commands like `npm install` or to execute a shell script in your source tree.
-
-```json
-"postCreateCommand": "bash scripts/install-dev-tools.sh"
-```
+When you make changes like installing new software, changes made in the Dockerfile will persist even upon a rebuild of the dev container.
 
-By default, when VS Code starts a container, it will **override the container's default command** to be `/bin/sh -c "while sleep 1000; do :; done"`. This is done because the container will stop if the default command fails or exits. However, this may not work for certain images. If the image you are using requires the default command be run to work properly, add the following to your `devcontainer.json` file.
+In your Dockerfile, use `FROM` to designate the image, and the `RUN` [instruction](#using-an-image-or-dockerfile) to install any software. You can use `&&` to string together multiple commands.
```
Chuxel, line 150: Same missing anchor here

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
 
-```json
-"overrideCommand": false
+```Dockerfile
+FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:0-12
+RUN apt-get update && apt-get install git
```
Chuxel, line 154: We could show this as

```
RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
    && apt-get install git
```

You'll see warnings if you don't... which could confuse new people as I think about it.  Not sure if the addition of the export is more confusing or not.

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
+5. After the window reloads, a copy of the **build log will appear** in the console so you can investigate the problem. Edit the contents of the `.devcontainer` folder as required. (You can also use the **Remote-Containers: Open Log File...** command to see the log again if you close it.)
+6. Run **Remote-Containers: Rebuild and Reopen Folder in Container** and jump to step 4 if needed.
 
-```json
-"postCreateCommand": "sudo apt-get update && sudo apt-get install <package>"
-```
+If you already have a successful build, you can still edit the contents of the `.devcontainer` folder as required when connected to the container and then select **Remote-Containers: Rebuild Container** in the Command Palette (`kbstyle(F1)`) so the changes take effect.
 
-The `postCreateCommand` is run once the container is running, so you can also use the property to run commands like `npm install` or to execute a shell script in your source tree (if you have mounted it).
+You can also iterate on your container when using the **Remote-Containers: Open Repository in Container** command.
```
Chuxel, line 190: FYI - Getting rid of this difference is one of the things I'm hoping gets done in the Sept iteration - this is where I'd mentioned Erich and Christof in that other doc. Adds complexity where we really don't need / want it.

Chuxel (approved):
I like these revisions! Flows much better.  Made a few comments / suggestions inline.

Two other thoughts:
1. We may want to include the trick @chrmarti figured out to get localhost working in the Docker Compose section. I already unblocked some people in the Codespaces Beta with it, so it's really good info. vscode-dev-containers is updated, but a good call out for those building their own.

2. The Docker Compose section references the remoteUser topic, but the Dockerfile does not.  I wonder if it would be more effective to just bring that section out of Advanced and into here. It is a bit advanced, but given it's needed for Linux users, I kind of suspect it's worth having here. We still get people who don't realize they actually have to create the user in the container before referencing it for example.

Otherwise looks great!

chrmarti:
@Chuxel Would it make sense to start with the 'Add Config' command instead of starting with an empty devcontainer.json? That would get users started much quicker.

I'd avoid the `apt-get` examples in `postCreateCommand`, these are better run in the Dockerfile as you explain. We used `npm install` as an example for `postCreateCommand ` in the past, but I'd be fine with omitting `postCreateCommand` here and only have it in the reference section.

docs/remote/create-dev-container.md
bamurtaugh: Thanks for catching this! I think this will also affect our "full configuration edit loop" diagram and explanation, as well as a couple other items a bit later potentially.

docs/remote/create-dev-container.md
bamurtaugh: So the flow for working with repositories in containers would match the 6 steps above more closely rather than having its own unique 6 steps here? Do you think we should modify/remove anything for this doc as it currently stands, or wait until changes potentially happen in the September iteration?

docs/remote/create-dev-container.md
```diff
@@ -56,253 +43,162 @@ All of the predefined container configurations you can pick from come from the [
...
-  * [Docker Compose](#using-docker-compose): You are working with multiple orchestrated services that are described using a `docker-compose.yml` file.
-
-  * In each case, you may also need to [build container images and deploy to Docker or Kubernetes](/docs/remote/containers-advanced.md#using-docker-or-kubernetes-from-a-container) from inside your container.
-
-  * [Attach only](/docs/remote/attach-container.md): While not backed by `devcontainer.json`, you can attach to an already running container if none of the workflows described in this section meet your needs.
-
-This section will walk you through configuring your project for each of these situations. The [vscode-dev-containers GitHub repository](https://aka.ms/vscode-dev-containers) also contains dev container definitions to get you up and running quickly.
-
-### Using an image or Dockerfile
+* Spin up a [stand-alone "sandbox" container](#set-up-a-folder-to-run-in-a-container) to isolate your toolchain or speed up setup.
```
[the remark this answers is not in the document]
bamurtaugh: Thanks for pointing these out! Since I removed the stand-alone "sandbox" section, I removed the link here, and then went through and updated other anchor tags in the doc.

docs/remote/create-dev-container.md
bamurtaugh: Updated this description along with the diagram (and some other references) mentioned above.

docs/remote/create-dev-container.md
bamurtaugh: That's a good idea. I feel like it'd make sense to include those tips- I added some information about them here, let me know what you think.

docs/remote/create-dev-container.md
bamurtaugh: I agree updating to the new syntax would make sense to set readers up for success later! Updated this reference- I didn't see other outdated references, but if I missed any, please just let me know.

docs/remote/create-dev-container.md
bamurtaugh: I don't think this one needs a link/tag, so removed altogether.

docs/remote/create-dev-container.md
bamurtaugh: I think leading users to something where they'll run into warnings and not explaining the warnings or how to combat them would be confusing. I updated `RUN` to match your suggestion and also added a note below it that explains it was added to avoid running into warnings later on.

bamurtaugh:
Thank you both for the great feedback!

@Chuxel I believe I addressed your outstanding comments and suggestions, including adding a section about getting localhost working in Docker Compose. I'm still a bit unsure about adding the `remoteUser` topic- were you thinking to include the entire [Adding a non-root user to your dev container](https://code.visualstudio.com/docs/remote/containers-advanced#_adding-a-nonroot-user-to-your-dev-container) section, or just the [Specifying a user for VS Code](https://code.visualstudio.com/docs/remote/containers-advanced#_specifying-a-user-for-vs-code) sub-section within it that focuses on `remoteUser`?

I would also be interested in hearing your perspective on @chrmarti's idea to start with the 'Add Config' command as it might save time and reflect a more realistic user journey, although this would likely impact our current "flow" building up from a blank canvas of devcontainer.json.

@chrmarti In discussions for building up this "happy path," we talked about going from devcontainer.json -> installing software in terminal -> installing software via `postCreateCommand` -> using Dockerfile and installing software as a best practice there -> using Docker Compose, so that's where the motivation came from to start with an empty devcontainer.json (rather than 'Add Config,') and to include `apt-get` examples in `postCreateCommand.` If a hesitancy is that the `apt-get` examples might encourage non-ideal practices, perhaps instead of showcasing actual code with `apt-get`, I could just explain that it would be possible to install software using `postCreateCommand` but it'd be better to use the Dockerfile?

cc @burkeholland as well since this content aligns with discussions we've recently had, and would love to know if he has any input, too.

chrmarti:
Omitting the `apt-get` examples for `postCreateCommand` and instead explain that this is a suboptimal option sounds good.

On the flow: I think both flows work for different users. Maybe when showing how to start from scratch mention and link to the 'Add Config' section so users preferring that discover it early on.

bamurtaugh:
@chrmarti Those suggestions sound good to me as well.

I've removed the `apt-get` examples and instead kept the `postCreateCommand` explanations more conceptual. 

I added a note early on (to the end of `Path to creating a dev container`) explaining to users that they can skip the step-by-step/manual process of setting up `devcontainer.json` and a Dockerfile outlined in the doc and instead automate the process using 'Add Config,' and I link to the `Automating dev container creation` section that explains the command.

bamurtaugh:
@gregvanl Since I believe I've addressed @Chuxel's and @chrmarti's feedback, would you be able to give this a review when you get a chance? 

I believe the next steps here would be getting your input on this doc, iterating on any of your feedback, getting changes merged in, and then looking toward the nav/TOC changes we wanted to make to improve the content navigation and consumption experience for readers (i.e. finding ways to group related remote topics together, potentially matching the "API" style).

Chuxel (commented):
Did another pass:

1. One thing I noticed is a number of the image paths have slashes in the wrong direction which won't work in the docs system (or on mac or Linux in general).  I only marked a couple, but there seemed to be a number of them.
2. On the "Clone Repository in Container Volume" flow, Erich is picking up this update this sprint. Assuming we merge before the sprint is over, we can leave it as is for now.
3. Mentioned one other possible tip as well below.

Otherwise LGTM!

docs/remote/create-dev-container.md
```diff
@@ -56,253 +45,158 @@ All of the predefined container configurations you can pick from come from the [
...
+If `devcontainer.json`'s supported workflows do not meet your needs, you can also [attach to an already running container instead](/docs/remote/attach-container.md).
 
-  * [Attach only](/docs/remote/attach-container.md): While not backed by `devcontainer.json`, you can attach to an already running container if none of the workflows described in this section meet your needs.
+> **Tip:** Want to use a remote Docker host? See the [Advanced Containers article](/docs/remote/containers-advanced.md#developing-inside-a-container-on-a-remote-docker-host) for details on setup.
 
-This section will walk you through configuring your project for each of these situations. The [vscode-dev-containers GitHub repository](https://aka.ms/vscode-dev-containers) also contains dev container definitions to get you up and running quickly.
+With this `devcontainer.json`, your dev container is functional, and you can connect to and start developing within it. Try it out with the **Remote-Containers: Reopen in Container** command:
 
-### Using an image or Dockerfile
+![Quick pick with list of Remote-Containers commands](images\containers\Remote-Containers-Commands-Reopen.png)
```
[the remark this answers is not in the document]
gregvanl: Git has trouble with casing so best to use lower case for image filenames.
So change filename to remote-containers-command-reopen.png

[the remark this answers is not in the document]
bamurtaugh: Fixed the slashes and changed filename to use lowercase!

docs/remote/create-dev-container.md
```diff
@@ -56,253 +45,158 @@ All of the predefined container configurations you can pick from come from the [
...
 
-You can configure VS Code to reuse an existing image from a source like [DockerHub](https://hub.docker.com) or [Azure Container Registry](https://azure.microsoft.com/services/container-registry/) for your dev container by adding a `.devcontainer/devcontainer.json` (or `.devcontainer.json`) config file to your project. In addition, if you are not able to find an image that meets your needs, have a single container-based project, or just want to automate the installation of several additional dependencies, you can use a [Dockerfile](https://docs.docker.com/engine/reference/builder/) to generate the image instead.
+With this command, you're now developing within a Node.js and TypeScript dev container with port 3000 forwarded and the eslint extension installed. Once you're inside, notice the green remote indicator updates to show you are connected to your dev container:
 
-To get started quickly, **open the folder** you want to work with in VS Code and run the **Remote-Containers: Add Development Container Configuration Files...** command in the Command Palette (`kbstyle(F1)`).
+![VS Code instance connected to dev container](images\containers\connected-to-dev-container.png)
```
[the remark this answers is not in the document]
bamurtaugh: Thanks for catching these! Did a search on the doc, and it looks like you caught the two images that had slashes in the wrong direction, so fixed them both now.

bamurtaugh:
Thanks so much for reviewing again, @Chuxel! Fixed the image paths and added the `postCreateCommand` tips you mentioned above about the interactive shell `i` flag. 

Also addressed @gregvanl's comment about using lowercase image names.

gregvanl:
@bamurtaugh Git is a little finicky about filename casing and it looks like the image name in the PR is still using uppercase.
You may need to run `git mv -f OriginalFile.txt originalfile.txt` to force git to respect the casing change.

bamurtaugh:
@gregvanl I went ahead and renamed the file- let me know if this seems to work properly on your end as well.

PR #3969, opened 2020-08-28
SSH tutorial updates

### LINKED ISSUE #3965: Update SSH tutorial Azure flow and screen shots
The Azure portal UI has been updated and the user is offered to create an SSH key pair as part of creation.

Also Node.js debugging didn't work if you take that apt-get node default of version 8. You need version 10.

bamurtaugh:
Update Azure Portal UI screenshot to include dropdown, explain dropdown.

Since Burke and I both ran into issues with port 22 timeout (we each lost about 3 hours to this error and only figured it out because Chuck was familiar with it), I tried adding a warning note and screenshot in case other users just starting out face this problem. If you think the warning could be moved to the main SSH doc, remote tips & tricks, or just removed, I'm definitely open to feedback.

Fixes #3965


docs/remote/ssh-tutorial.md
```diff
@@ -71,9 +71,9 @@ You will then be prompted to enter a secure passphrase, but you can leave that b
 
 ## Add SSH key to your VM
 
-In the previous step, you generated an SSH key pair. Take the public key and paste it into your VM setup, by copying the entire contents of the `id_rsa.pub`. You also want to allow your VM to accept inbound SSH traffic by selecting **Allow selected ports** and choosing **SSH** from the **Select inbound ports** dropdown list.
+In the previous step, you generated an SSH key pair. Select **Use existing public key** in the drop down for **SSH public key source** so that you can use the public key you just generated. Take the public key and paste it into your VM setup, by copying the entire contents of the `id_rsa.pub` in the **SSH public key**. You also want to allow your VM to accept inbound SSH traffic by selecting **Allow selected ports** and choosing **SSH (22)** from the **Select inbound ports** dropdown list.
 
-![Add SSH public key to VM](images/ssh-tutorial/add-ssh-public-key.png)
+![Add SSH public key to VM](images/ssh-tutorial/add-ssh-public-key-updated.png)
```
gregvanl, line 76: You can just commit the new image to the old image filename. Otherwise we end up with lots of obsolete images on the website. We've got source control so the old image is always with us. :-)

docs/remote/ssh-tutorial.md
```diff
@@ -99,6 +99,12 @@ The `user` is the username you set when adding the SSH public key to your VM. Fo
 
 ![Virtual machine public IP address](images/ssh-tutorial/vm-public-ip-address.png)
 
+Before connecting in Remote - SSH, you can verify you're able to connect to your VM via a command prompt using `ssh user@hostname`.
+
+> Note: If you run into an error `ssh: connect to host <host ip> port 22: Connection timed out`, you may need to delete NRMS-Rule-106 from the Networking tab of your VM:
+
+![Virtual machine list of NRMS rules](images/ssh-tutorial/vm-nrms-rules.png)
```
gregvanl, line 106: This image doesn't clearly show NRMS-Rule-106. Perhaps it is the entry with the SSH warning triangle?
Can you add a red outline (3 pixels is the convention) to highlight on the entry to remove?
Also since this image is for the Note, you can indent 3 spaces to indent the image.

docs/remote/ssh-tutorial.md
bamurtaugh: That's a great point, it'll always be with us :) Updated the new image name to remove `-updated` and removed the obsolete image.

docs/remote/ssh-tutorial.md
bamurtaugh: I realized that I included an image from a VM where I had already deleted Rule 106, so you're absolutely right that the image didn't clearly show NRMS-Rule-106! I added a red outline around the name of the rule and the prompt to delete it. I used paint to do this, and the 3 pixel border looks a bit smaller than the other red rectangles I see in tutorials, so let me know if I should make it more bold, or if you'd like me to make any other adjustments to this image (i.e. cropping it more, highlighting some other component of it with the red rectangle).

Also, I recognize the VM I had on-hand with a Rule 106 available has a different name/IP address than Sana's VM we reference in other parts of the tutorial, but since this screenshot is to emphasize Rule 106, I figured that would be alright. If you think I should crop or erase the name and CPU, or make a new VM with a different name, please let me know.

I added the 3 space indentation as well.

gregvanl: Looks great. Thanks for redoing the image.

Issue #1363, opened 2018-01-24
Chinese (zh-Hans) translation of keyboard shortcuts keymap wanted


gregvanl:
_From @xgqfrms-GitHub on January 2, 2018 15:32_

https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf

# i18n & L10n

> zh-Hans is wanted!

https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows-zh.pdf

_Copied from original issue: Microsoft/vscode#41035_

gregvanl:
_From @auchenberg on January 24, 2018 0:43_

@gregvanl Should we move this to the docs repo?

lorderikir:
@gregvanl. Is there a template for this doc?

gregvanl:
@lorderikir You can find the original Word documents that are used to create the PDFs at https://github.com/Microsoft/vscode-docs/tree/master/docs/customization.

lorderikir:
Awesome thanks @gregvanl

gregvanl:
@lorderikir Just a heads-up, the VS Code website doesn't support localized content or currently have a way to provide a localized PDF. It would require some engineering work to support this but Chinese versions of the PDFs could be hosted in this repo initially.

lorderikir:
I've started the initial work here: https://github.com/lorderikir/vscode-docs/tree/feature/docs/zh-HANS-doc

gregvanl:
Closing old issue. No one has responded to the `help wanted` label.

PR #3974, opened 2020-09-01
Swedish keyboard shortcuts for Windows

samirgunic:
Added keyboard shortcuts cheat sheet for Windows, localized for Swedish keyboards.


samirgunic:
I basically ran into the same problem as described here:
https://github.com/microsoft/vscode-docs/issues/1363#issue-291063736

The only difference is that I'm using a Swedish keyboard, not a Chinese keyboard.

I grabbed the original DOCX file and started translating the key bindings myself, one command at a time. The result is a localized version of VS Code keyboard shortcut cheat sheet for Swedish keyboards. Now I can finally share this with my classmates.

samirgunic:
Creation of these keyboard shortcut cheat sheets for VS Code should be automated, similar to how the VS Code website automatically detects the operating system of the site visitor and displays the corresponding keyboard shortcuts for that particular system.

Not all VS Code users have a keyboard with English (language) layout. Well, actually, I think my teacher does. But that's an anomaly. It's a seldomly seen exception, and the case of VS Code cheat sheets makes it very obvious why he decided to get a Mac with English layout rather than Swedish layout. In any case, cheat sheets are great tools for learning when available, so they should be localized to whatever locale necessary, just like VS Code itself is localized to the environment it's used in.

Now that I think of it, the VS Code website is yet another resource that users cannot use for reference unless they have a English keyboard layout. Because the website is unable to detect keyboard layouts, so it's limited to English layouts by design.

PR #3978, opened 2020-09-03
coding pack docs page with latest feedback

JoshuaPartlow:


docs/python/coding-pack-python.md
```diff
@@ -0,0 +1,145 @@
...
+Along with the tools and packages necessary for Python development, the Coding Pack also configures common user settings and PowerShell. This includes Python extension settings, such as the default interpreter and language server, as well as execution policies to allow for virtual environment activation in the terminal.
+
+>**Note**: If there was an existing version of Visual Studio Code installed on your machine, your settings.json will not be overwritten and you’ll need to configure Python settings yourself.
+
+## Getting started
+
+With the Coding Pack for Python it's easy to get started developing with Python and VS Code.
+
+1. [Download](https://aka.ms/coding-pack-for-python-win) and run the Coding Pack for Python installer.
+   >**Note**: The installer only supports installation on 64-bit Windows.
```
zooba, line 39:
```suggestion
   >**Note**: The installer only supports Windows 10 64-bit. This download is 200MB, and up to 100MB will be downloaded during installation.
```

docs/python/coding-pack-python.md
```diff
@@ -0,0 +1,145 @@
...
+
+      [![Python extension on Marketplace](images/coding-pack/python-extension-marketplace-install.jpg)](https://marketplace.visualstudio.com/items?itemName=ms-python.python)
+
+### Python interpreter
+
+1. If there was an issue installing the Python interpreter, you can install Python 3.8 from the [Microsoft Store](https://www.microsoft.com/p/python-38/9mssztt1n39l?activetab=pivot:overviewtab). Along with the Python extension, you need to install a Python interpreter for development with Python. There are other options for installing the Python interpreter, such as directly from [Python.org](https://www.python.org/downloads/), and which interpreter you use is dependent on your specific needs. If you use the Python.org version, you can typically use the **Download Python** button that appears first on the page to download the latest version. Just make sure to uncheck the "Install launcher for all users" box if you don't have admin access.
```
zooba, line 69:
```suggestion
1. If there was an issue installing the Python interpreter, you can install Python 3.8 from the [Microsoft Store](https://www.microsoft.com/store/apps/9MSSZTT1N39L?cid=vscode-docs). Along with the Python extension, you need to install a Python interpreter for development with Python. There are other options for installing the Python interpreter, such as directly from [Python.org](https://www.python.org/downloads/), and which interpreter you use is dependent on your specific needs. If you use the Python.org version, you can typically use the **Download Python** button that appears first on the page to download the latest version. Just make sure to uncheck the "Install launcher for all users" box if you don't have admin access.
```

Updated the link to the Store page.

> If you use the Python.org version, you can typically use the **Download Python** button that appears first on the page to download the latest version.

Just a heads up that in about a month this button will change to Python 3.9. We might be better linking to https://www.python.org/downloads/release/python-385/

docs/python/coding-pack-python.md
```diff
@@ -0,0 +1,145 @@
...
+- Useful Python packages
+  - jupyter
+  - numpy
+  - sklearn
+  - pandas
+  - Matplotlib
+
+Along with the tools and packages necessary for Python development, the Coding Pack also configures common user settings and PowerShell. This includes Python extension settings, such as the default interpreter and language server, as well as execution policies to allow for virtual environment activation in the terminal.
+
+>**Note**: If there was an existing version of Visual Studio Code installed on your machine, your settings.json will not be overwritten and you’ll need to configure Python settings yourself.
```
zooba, line 32: Can we move this below the Getting started section? Just trying to keep the main CTA (the download link) above the fold.

docs/python/coding-pack-python.md
JoshuaPartlow: Changed in the latest push.

docs/python/coding-pack-python.md
JoshuaPartlow: Sure thing, changed in the latest push.

PR #3990, opened 2020-09-08
docs: update disabling crash reporter section

deepak1556:
With Electron 9 that is going out with `1.49` we have changed the preferred way of disabling crash reporter.

### DIFF
--- a/docs/getstarted/telemetry.md
+++ b/docs/getstarted/telemetry.md
@@ -25,22 +25,6 @@ If you use the JSON editor for your settings, add the following line:
     "telemetry.enableTelemetry": false
 ```
 
-## Disable crash reporting
-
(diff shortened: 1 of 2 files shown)


docs/getstarted/telemetry.md
```diff
@@ -25,22 +25,6 @@ If you use the JSON editor for your settings, add the following line:
     "telemetry.enableTelemetry": false
 ```
 
-## Disable crash reporting
```
gregvanl, line 28: We should keep both sections, in FAQ and the telemetry topic. 
There may be external hard-coded links to code.visualstudio.com/docs/getstarted/telemetry#_disable-crash-reporting.
I don't think there is harm in having the same incantation in two places.

deepak1556: updated, thanks!

bpasero (approved):

PR #4013, opened 2020-09-17
clarify notebook's debugger section

weinand:
Since I've updated the notebook debugging samples to the latest APIs, I noticed that they are referred in the notebook API doc. I took the liberty to clarify how notebook debugging can be implemented and how this is shown in the samples.


JacksonKearl:
Thanks for the clarification, looks good to me.

PR #4015, opened 2020-09-18
Update config-msvc.md

139629905:
The 'Configure Default Build Task' option is subordinate to 'Tasks'
image


gregvanl:
Hi @139629905 Thank you for submitting a PR. However, the text is describing the menu item not the entry in the Command Palette.

image

PR #4021, opened 2020-09-22
Change default VS Code path to use %localappdata%

BasPaap:
The documentation presumed that the users folder is on C: and used a {username} pseudo-wildcard. Using %localappdata% ensures that the path goes to the user's local appdata folder and allows them to copy paste the path from the docs.


gregvanl:
Thanks! Good refinement.

PR #4024, opened 2020-09-23
Use Dockerfile ENV key=value

StefanScherer:
This PR cleans up the docs a bit to use the Dockerfile `ENV key=value` syntax.
I haven't changed the screenshots, I think it's good enough to have the copy-pastable code examples updated.

See also https://github.com/microsoft/vscode-docker/pull/2344


ghogen (approved):

ghogen:
@JoshuaPartlow Some of these are in the python docs - can you look?

PR #4025, opened 2020-09-23
IntelliSense minor update

shymanel:
By default, `tabCompetition` is off.

https://code.visualstudio.com/docs/getstarted/settings#_default-settings

Although I still use `Tab` to confirm suggestions, even though it's disabled. 🤔


gregvanl:
Thanks!

PR #4026, opened 2020-09-23
Remove broken image from unity docs and fix link

GabLeRoux:
https://code.visualstudio.com/docs/other/unity#_unity-version-20192-or-above

image


gregvanl:
Thanks! Not sure why no one noticed this before.

PR #4027, opened 2020-09-24
Add description of envFile

andreas-bergstrom:
It would probably help plenty of people if this attribute was documented here. See for example this SO-thread:
https://stackoverflow.com/questions/29971572/how-do-i-add-environment-variables-to-launch-json-in-vscode


gregvanl:
Thanks!

PR #4036, opened 2020-09-28
Update network.md

sean-mcmanus:
Add download.visualstudio.microsoft.com to the list of hostnames to allow through a firewall. See https://github.com/microsoft/vscode-cpptools/issues/3335#issuecomment-697764255 .

### DIFF
--- a/docs/setup/network.md
+++ b/docs/setup/network.md
@@ -29,6 +29,7 @@ If you are behind a firewall that needs to allow specific domains used by VS Cod
 * `raw.githubusercontent.com`
 * `vsmarketplacebadge.apphb.com`
 * `az764295.vo.msecnd.net`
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @sean-mcmanus Do you think we should also add `download.microsoft.com`?

sean-mcmanus:
@gregvanl I don't think download.microsoft.com should be added unless we find a case where that is required. A long time ago we used to use download.microsoft.com, but now it's all in a form like https://download.visualstudio.microsoft.com/download/pr/ebeeef98-9fc6-4e24-8db3-1820882eacb4/e200839092f03dfff01d992592f480f0/bin_windows.zip. The 500 is fine since the URL doesn't serve normal browser web traffic.

PR #4043, opened 2020-10-04
Update config-clang-mac.md

jacklu20:
The label of the 'preLaunchTask' has to match the label of the build task in task.json file, otherwise, the debugger will use the default build task which would not compile the helloworld program properly due to the usage of C++11 & 17 features.

### DIFF
--- a/docs/cpp/config-clang-mac.md
+++ b/docs/cpp/config-clang-mac.md
@@ -229,7 +229,7 @@ By default, the C++ extension won't add any breakpoints to your source code and 
 
 Change the `stopAtEntry` value to `true` to cause the debugger to stop on the `main` method when you start debugging.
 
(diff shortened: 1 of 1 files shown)


docs/cpp/config-clang-mac.md
```diff
@@ -229,6 +229,8 @@ By default, the C++ extension won't add any breakpoints to your source code and
 
 Change the `stopAtEntry` value to `true` to cause the debugger to stop on the `main` method when you start debugging.
 
+Also make sure the value of `preLaunchTask` matches the `label` of the build task in task.json file.
```
TylerMSFT, line 232: Thank you for adding this.

We could also make it more direct, e.g.:

Ensure that the `preLaunchTask` value matches the `label` of the build task in the task.json file.

jacklu20 (commented):
Updated with Tyler's comment.

PR #4046, opened 2020-10-07
Improve black warning troubleshooting

default50:
- Move the warning pop up about `black` not supporting "Format Select" inside the troubleshooting table.
- Add setting for formatting the whole file at save (instead of only "modifications") since that also triggers the warning.

### DIFF
--- a/docs/python/editing.md
+++ b/docs/python/editing.md
@@ -154,17 +154,8 @@ If formatting fails, check the following possible causes:
 | The path to the python interpreter is incorrect. | Check the `pythonPath` setting. |
 | The formatter is not installed in the current environment. | Open a command prompt, navigate to the location specified in the `pythonPath` setting, and run `pip install` for the formatter.
 | The path to the formatter is incorrect. | Check the value of the appropriate `python.formatting.<formatter>Path` setting. |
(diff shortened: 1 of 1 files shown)


flavioelawi (approved):

luabud (changes requested):
Thanks for submitting this PR! 😊 

I have only two requests: 
1. I think it would be better to change "solved" to "prevented", as the warning doesn't go away if you run "Format Selection" command, it's just prevented in case formatting on paste is enabled
 2. I also believe the sentence "This will force formatting of the *whole* file at save." is true, as format on save may be disabled and in that case, the file won't be formatted on save. So that sentence can be removed.

default50:
> Thanks for submitting this PR! 😊
> 
> I have only two requests:
> 
>     1. I think it would be better to change "solved" to "prevented", as the warning doesn't go away if you run "Format Selection" command, it's just prevented in case formatting on paste is enabled
> 
>     2. I also believe the sentence "This will force formatting of the _whole_ file at save." is true, as format on save may be disabled and in that case, the file won't be formatted on save. So that sentence can be removed.

I agree with your comments, implemented them in the last commit. Thanks for your feedback!

flavioelawi (approved):

PR #4047, opened 2020-10-07
Add new git theme options

mwood77:
Updated theme-color.md with new git theme options


joaomoreno:
This is already on `vnext` branch: https://github.com/microsoft/vscode-docs/commit/6b9a0ac47c37b9431d58ccd328961eb80690b07f

Thanks anyway

PR #4059, opened 2020-10-12
Update linux.md

tbjgolden:
There were two issues with the description:

1. Users on Arch may not have found the part where it mentions that Arch does things differently (I didn't)
2. The linked solution on Manjaro (generally one of the more popular noob-friendly distros) didn't work because another config file is blocking it

The description as I've modified it would have caught both issues earlier and hopefully will help others not make the same mistakes, without getting in the way of folks on other distros

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -205,15 +205,17 @@ When you see this notification, it indicates that the VS Code file watcher is ru
 cat /proc/sys/fs/inotify/max_user_watches
 ```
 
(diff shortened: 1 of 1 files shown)


joaomoreno:
@gregvanl The changes here are logically sound, but make the answer flow strangely. Can you rewrite it nicely? Thanks!

tbjgolden:
@joaomoreno updated - better?

gregvanl:
Looks good. Thanks!

PR #4066, opened 2020-10-15
Run $USERPROFILE/.bash_profile in Git Bash

karlhorky:
Running Git Bash like the change in the pull request causes `$USERPROFILE/.bash_profile` to also be loaded, potentially loading important environment variables, aliases or other configuration.

### DIFF
--- a/docs/editor/integrated-terminal.md
+++ b/docs/editor/integrated-terminal.md
@@ -66,7 +66,12 @@ Just like on other platforms you can fine tune the exact executable used in your
 // PowerShell
 "terminal.integrated.shell.windows": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
 // Git Bash
-"terminal.integrated.shell.windows": "C:\\Program Files\\Git\\bin\\bash.exe"
+"terminal.integrated.shell.windows": "C:\\Program Files\\Git\\git-cmd.exe",
+"terminal.integrated.shellArgs.windows": [
+  "--command=usr/bin/bash.exe",
+  "-l",
+  "-i"
+]
 // Bash on Ubuntu (on Windows)
 "terminal.integrated.shell.windows": "C:\\Windows\\System32\\bash.exe"
 ```


Tyriar:
I think this is already covered in the below section explaining shell args.

Additionally the main way of launching git bash will be to select the shell in the dropdown which we currently do not support shell args for. Maybe this would be a good change to include in that when it supports shell args https://github.com/microsoft/vscode/issues/45444#issuecomment-609836354.

karlhorky:
@Tyriar ah, I think there's a misunderstanding... the problem is that the version being recommended in the docs currently is **broken** - the behavior is different than the Git Bash program as launched from the start menu (`$USERPROFILE/.bash_profile` is not loaded, and because of that, any environment variables or aliases or anything else are not run)

Tyriar:
I don't believe the user starts out with a `.bash_profile file and I worry about complicating the setup on the docs and also making it different from how it behaves as part of the Select Default Shell commands. Additionally it appears this is how Windows Terminal also sets it up with this:

```json
{
    "name": "Git Bash",
    "commandline": "C:\\Program Files\\Git\\bin\\bash.exe",
    "hidden": false
}
```

It's probably a good change, but I'd rather do it when we can hide this complexity from the user when https://github.com/microsoft/vscode/issues/45444#issuecomment-609836354 is implemented.

karlhorky:
Hm, ok, I understand not wanting to complicate things.

How about:

1. removing Git Bash from section it's currently in
2. adding another example in **Shell arguments** below with Git Bash, configured as in this PR

...as an alternative, until the shell args support comes?

Tyriar:
@karlhorky I'm curious where you got these command line args from. More specifically, how do you know git bash runs exactly this and is `--command` and `-i` really required?

karlhorky:
If I remember correctly, I think I collected these from various GitHub issues and blog posts around the Internet, after comparing multiple solutions... I can spend some time looking for a good canonical source if this would be helpful.

One thing that leads me to believe that this is the correct way to launch Git Bash is that these arguments flash up in the title bar of the program when starting the official Git Bash program from the Start Menu.

Tyriar (approved):
Great point, I think this is enough to justify bringing this change in:

image

In the long run I'd like to see this eventually work with selecting the default shell as well.

The dropdown title is the major user facing change from this:

image

Creating https://github.com/microsoft/vscode/issues/110098 to fix the lack of support for recognizing git-cmd.exe as git bash too.

karlhorky:
Sounds good, thanks! Following the new issue and the other mentioned issue now too.

PR #4074, opened 2020-10-18
Show key binding for 'Find All References'

DAXaholic:
Currently there is no key binding shown for **Find All References** because it refers to an invalid (probably old) command ID.  
By replacing it with the current ID the key binding should be shown again.

image


gregvanl:
Thanks!

PR #4085, opened 2020-10-27
add missing new keyword

yoshiakis:


api/extension-guides/virtual-documents.md
```diff
@@ -22,7 +22,7 @@ vscode.workspace.registerTextDocumentContentProvider(myScheme, myProvider);
 Calling `registerTextDocumentContentProvider` returns a disposable with which the registration can be undone. A provider must only implement the `provideTextDocumentContent`-function which is called with an uri and cancellation token.
 
 ```ts
-const myProvider = class implements vscode.TextDocumentContentProvider {
+const myProvider = new class implements vscode.TextDocumentContentProvider {
```
gregvanl, line 25: There is another codeblock on line 59 that creates this same provider. Does that also need to be updated?

yoshiakis: Good point! I added another commit for that.

gregvanl:
Thanks!

PR #4094, opened 2020-10-31
Document fontLigatures format

lf-:
I had to read the source code to figure out how this property worked. Let me know if there's anywhere else I should improve the documentation of this setting.

### DIFF
--- a/docs/getstarted/settings.md
+++ b/docs/getstarted/settings.md
@@ -292,7 +292,9 @@ Below are the Visual Studio Code default settings and their values. You can also
     // Controls the font family.
     "editor.fontFamily": "Consolas, 'Courier New', monospace",
 
-    // Configures font ligatures or font features.
+    // Configures font ligatures or font features. Can be of two formats:
+    //   - true/false: enables/disables `calt` and `liga` font features together
+    //   - string: of the same format as the CSS property `font-feature-settings`.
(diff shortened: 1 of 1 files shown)


gregvanl:
HI @lf- Thank you for tracking this down and submitting a PR. The settings section is actually auto-generated from the in-product defaultSettings.json (Preferences: Open Default Settings (JSON)) so the description fix should be made in the vscode product repo in https://github.com/microsoft/vscode/blob/master/src/vs/editor/common/config/editorOptions.ts. 
Search for class EditorFontLigatures and you can see the description to update.
I think you'll have to format the description as a single string.

lf-:
@gregvanl In that case, I'd say that the current documentation generator is buggy! Although the option should mention in its own description that there are multiple types possible and I should PR that, the documentation generator should emit the docs for both of the cases in the anyOf as defined here: https://github.com/microsoft/vscode/blob/master/src/vs/editor/common/config/editorOptions.ts#L1410

gregvanl:
Hi @lf- Agreed, that sometimes the generator does the right thing, for example with enumerated values (see editor.unusualLineTerminators), but not in the case of `anyOf`.
The quick fix would be to update the description string but one could also look into adding better support for `anyOf` in the defaultSettings.json output.

PR #4098, opened 2020-11-01
Doc: Fix vscode.executeDefinitionProvider return type

a7madgamal:
Seems the returned object is of type LocationLink (based on a debug inspection of the values)


gregvanl:
Thanks for pointing this out! This page is autogenerated and the comment is actually "an array of Location or LocationLink instances. See https://github.com/microsoft/vscode/blob/master/src/vs/workbench/api/common/extHostApiCommands.ts. There are also several other API comments that need to be updated. 
I'll take your PR and add the other updates.

Issue #3831, opened 2020-06-30
EventEmitter fire() method outdated in TreeView docs


cemore2048:
Hi vscode team 👋 Partook in the MLH VSCode hackathon this weekend, and in going through the docs for the `TreeView`, it seems like the `fire()` method in the `EventEmitter` example here is out of date. 

```
private _onDidChangeTreeData: vscode.EventEmitter<Dependency | undefined> = new vscode.EventEmitter<Dependency | undefined>();
  readonly onDidChangeTreeData: vscode.Event<Dependency | undefined> = this._onDidChangeTreeData.event;

  refresh(): void {
    this._onDidChangeTreeData.fire();
  }
```

The signature now has to take in a type 
```
fire(data: T): void;
```

Happy to submit a PR to change, just wanting to make sure it is indeed out of date.

gregvanl:
Hi @cemore2048 Thanks for pointing this out. This is actually an issue with our API generation tooling which takes the vscode.d.ts type definition and generates the document at https://code.visualstudio.com/api/references/vscode-api. 

In vscode.d.ts, you can see that the data argument is optional:

```typescript
	export class EventEmitter<T> {

		/**
		 * The event listeners can subscribe to.
		 */
		event: Event<T>;

		/**
		 * Notify all subscribers of the [event](#EventEmitter.event). Failure
		 * of one or more listener will not fail this function call.
		 *
		 * @param data The event object.
		 */
		fire(data?: T): void;

		/**
		 * Dispose this object and free resources.
		 */
		dispose(): void;
	}
```
Unfortunately the '?' is not preserved in the generated API reference. This issue is also noted in https://github.com/microsoft/vscode-docs/issues/3278.

For API questions, vscode.d.ts and IntelliSense (code suggestions and hovers) are the ultimate source of truth.

I'll close this as a duplicate of https://github.com/microsoft/vscode-docs/issues/3278

PR #4099, opened 2020-11-02
Fix TreeDataProvider.onDidChangeTreeData example

dsanders11:
I believe #3831 was closed incorrectly, as it was referring to this docs example, not the auto-generated API reference.

I think the example code is overly wordy with repeating the type three times when once will suffice, but since that's how it already was I'm leaving that and simply updating it so that the provided example actually compiles.


gregvanl:
Thanks!

PR #4100, opened 2020-11-03
Add devcontainer.json note about formatting of properties

bamurtaugh:
Based on user feedback, we wanted to add details about the formatting of certain properties in `devcontainer.json`.

While the user who ran into this quote/shell formatting issue faced it for the `runArgs` property specifically, @chrmarti pointed out that the same principle applies to `postCreateCommand`, `postStartCommand`, `postAttachCommand`, and `initializeCommand` too, so we thought adding a note to the bottom of the table could help encapsulate the point for multiple args.

@gregvanl If you have feedback over the wording and the format or placement of the note (i.e. any best practices to ensure a balance between a tip being applied to multiple properties and ensure users see it, without duplicating it unnecessarily), would love to hear it!

@chrmarti Would love to hear if you have any feedback over the wording or technical accuracy of the note.


docs/remote/devcontainerjson-reference.md
```diff
@@ -59,6 +59,12 @@ While some devcontainer.json properties apply generally, others are only used in
 
 If you've already built the container and connected to it, be sure to run **Remote-Containers: Rebuild Container** from the Command Palette (`kbstyle(F1)`) to pick up the change.
 
+> **Note:** The format of certain properties will vary depending on the involvement of a shell.
+
+> `postCreateCommand`, `postStartCommand`, `postAttachCommand`, and `initializeCommand` all have an array and a string type. An array is passed to the OS for execution without going through a shell, whereas a string goes through a shell (it needs to be parsed into command and arguments).
```
chrmarti, line 64: Maybe mention "runArgs" here too to make this paragraph complete as a high-level description. (Maybe in separate sentence since "runArgs" only has the array variant.)

docs/remote/devcontainerjson-reference.md
```diff
@@ -59,6 +59,12 @@ While some devcontainer.json properties apply generally, others are only used in
...
+
+> When using `runArgs` via a typical command line, you'll need single quotes if the shell runs into parameters with spaces. However, these single quotes aren't passed on to the executable. Thus, in your `devcontainer.json`, you'd follow the array format used by the properties mentioned above, i.e.: `"runArgs": ["--device-cgroup-rule=my rule here"]`, rather than `"runArgs": ["--device-cgroup-rule='my rule here'"]`. Its arguments will be passed as an array, and no quoting for any shell is involved.
```
chrmarti, line 66: The name "runArgs" doesn't appear on the command line, I'd rephrase "using `runArgs` via a typical command line". Or: Use an example with "post*Command" comparing the string and the array versions: `"postAttachCommand": "echo foo='bar'"` will remove the single quotes as part of the shell's parsing whereas `"postAttachCommand": ["echo", "foo='bar'"]` will keep them and write them to standard out. (You can see the output in the Dev Container log.)

docs/remote/devcontainerjson-reference.md
bamurtaugh: That's good feedback, so that it hopefully won't feel as though `runArgs` is randomly introduced later. I updated the section here, let me know if you think it helps, or I can tweak it a bit more.

docs/remote/devcontainerjson-reference.md
bamurtaugh: These are great suggestions too. I rephrased using your first suggestion and also added the `postAttachCommand` example. Let me know what you think!

docs/remote/devcontainerjson-reference.md
```diff
@@ -59,6 +59,14 @@ While some devcontainer.json properties apply generally, others are only used in
 
 If you've already built the container and connected to it, be sure to run **Remote-Containers: Rebuild Container** from the Command Palette (`kbstyle(F1)`) to pick up the change.
 
+> **Note:** The format of certain properties will vary depending on the involvement of a shell.
+
+> `postCreateCommand`, `postStartCommand`, `postAttachCommand`, and `initializeCommand` all have an array and a string type, while `runArgs` only has the array variant. An array is passed to the OS for execution without going through a shell, whereas a string goes through a shell (it needs to be parsed into command and arguments).
```
gregvanl, line 64: "only has the array type".
I was confused by the word "variant".

gregvanl:
@bamurtaugh Rather than using the Note and indent formatting, I think it would be better to use an H3, something simple like "Formatting string properties".
Another suggestion is to use standalone code blocks, rather than inline code blocks for the examples. Easier to parse and compare.

docs/remote/devcontainerjson-reference.md
bamurtaugh: Thanks for the feedback- updated.

bamurtaugh:
> @bamurtaugh Rather than using the Note and indent formatting, I think it would be better to use an H3, something simple like "Formatting string properties". There is some evidence that users skip Notes and Tips when reading an article.
> Another suggestion is to use standalone code blocks, rather than inline code blocks for the examples. Easier to parse and compare.

Thanks for the feedback @gregvanl, that definitely makes sense. I made updates to the formatting and a bit of the wording as well to follow this advice (i.e. switched to an H3, migrated from inline to standalone code blocks). Let me know what you think.

chrmarti (approved):
Looks good!

PR #4107, opened 2020-11-04
Mark 2 command switches ineffective in Portable

myfonj:
Portable mode (presence of  `/data/` folder in VSC installation directory) effectively prevents `--user-data-dir` and `--extensions-dir` switches from working.
Been told it is OK and will not be changed: https://github.com/microsoft/vscode/issues/63657
This info could potentially save a lot of WTH moments and confused fuddling to folks looking at this docs and not getting it working in their "Portable installs" when for example trying to launch a "vanilla" profile without touching their `/data/` directory.
(This edit is crude, probably would be better as a note below table, but I couldn't figure it out how to formulate and mark it.)

### DIFF
--- a/docs/editor/command-line.md
+++ b/docs/editor/command-line.md
@@ -93,8 +93,8 @@ There are several CLI options that help with reproducing errors and advanced set
 
 Argument|Description
 ------------------|-----------
-`--extensions-dir <dir>` | Set the root path for extensions.
-`--user-data-dir <dir>` | Specifies the directory that user data is kept in, useful when running as root.
+`--extensions-dir <dir>` | Set the root path for extensions. Has no effect in [Portable Mode](/docs/editor/portable.md).
+`--user-data-dir <dir>` | Specifies the directory that user data is kept in, useful when running as root. Has no effect in [Portable Mode](/docs/editor/portable.md).
 `-s, --status` | Print process usage and diagnostics information.
 `-p, --performance` | Start with the **Developer: Startup Performance** command enabled.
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks for calling this out. We might create a note below and add a mention in the Portable Mode doc as well but this is a good first start.

myfonj:
Nice, thanks!

Needless to say it would be better if this notice was not necessary and all command line switches worked predictably in both modes.  This state really feels unfortunate to me, and I would be very happy have them (or their better variants) removed in unlikely case https://github.com/microsoft/vscode/issues/63657 was revised and fixed after all.

Btw there is that _"Can I specify the settings location for VS Code […]"_ section closing this page that could mention those switches as well. Now answers is _"Not directly through the command line […]"_ what seems kinda contradictory to preceding sections.

https://github.com/Microsoft/vscode-docs/blob/master/docs/editor/command-line.md#can-i-specify-the-settings-location-for-vs-code-in-order-to-have-a-portable-version

myfonj:
Phrasing further improved and clarified with prose thanks microsoft/vscode/issues/181876 / #6309 .

---

`<necro-mumble to="void">`For the record, I am still 2000 % convinced that this behaviour, however well documented, is a bug, should have never existed in the first place, is against all the "\*nix" philosophy, since it violates all best practices and habits regarding configuration retrieval and arguments passing: **explicit command-line argument should always take precedence over any other implicit mean**. At least for this particular type of application.
Prove me wrong.`</necro-mumble>`

PR #4110, opened 2020-11-05
Delete api-event-bridge.png

favatacharles:


gregvanl:
Hi @favatacharles Thanks for noticing this obsolete image and cleaning it up.
It also looks like api-bridge.png and basic-helper.png are also no longer displayed, would you like to clean those up as well?

favatacharles:
thank you, yes i would that would be great!   
  
thank you nicole favata
  
  
  
    
  

  
  
>

gregvanl:
I'll delete the other two images.
Thanks!

PR #4115, opened 2020-11-08
Update mocha code on "testing-extension"

macabeus:
Since [Mocha 8.0.0](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#800--2020-06-10) `useColors` was removed, and now we should set the `color` as `true` on on `new Mocha`


gregvanl:
Thanks!

PR #4117, opened 2020-11-09
Update jupyter-support.md

andyrosa2:
Fix a noun/pronoun number disagreement.
a Python files-> a Python file.


gregvanl:
Thanks!

PR #4119, opened 2020-11-09
Add a clear step to open the Vs code in the "helloworld" folder  your-first-extension.md

caiobessa:
It is not clear in the documentation that we have to open the visual code in the ``helloworld`` folder. 

The command is in the end line of the scaffold and use environment variable.


gregvanl:
Hi @caiobessa We really like people to learn to use the VS Code command line to open folders.
What about leaving `code ./helloworld` in the code block and instead calling out what that command does. You could also rephrase your current addition to say something like "You can also launch VS Code and use File > Open...."

Issue #2480, opened 2019-03-14
tasks.json schema docs are missing custom task types (e.g. typescript, npm) and also other config properties


justingrant:
The task.json schema documentation (https://code.visualstudio.com/docs/editor/tasks-appendix) is incomplete and/or out of date. For example, it's missing documentation for NPM and TypeScript task types that are auto-detected by VSCode. Below are examples of two tasks that were auto-detected by VS Code 1.32.2. Other than `problemMatcher`, none of these settings are documented. 

    {
      "type": "typescript",
      "tsconfig": "api/tsconfig.json",
      "option": "watch",
      "problemMatcher": [
        "$tsc-watch"
      ]
    },
    {
      "type": "npm",
      "script": "start",
      "path": "api/",
      "problemMatcher": []
    }

In particular, I'm trying to figure out how the `option` config property works for the `typescript` task type, because I want to figure out how to to add `--noEmit` to the typescript command line used for a watch task. But more generally, all settings for all task types should be documented!

If you look at the task [auto-detection code](https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/contrib/tasks/node/taskConfiguration.ts#L1312-L1322) (excerpted below), there are at least 5 different task types that are not documented: `typescript`, `npm`, `gulp`, `jake`, and `grunt`.

```ts
if (customize.indexOf(grunt) === 0) {
  identifier = { type: 'grunt', task: customize.substring(grunt.length) };
} else if (customize.indexOf(jake) === 0) {
  identifier = { type: 'jake', task: customize.substring(jake.length) };
} else if (customize.indexOf(gulp) === 0) {
  identifier = { type: 'gulp', task: customize.substring(gulp.length) };
} else if (customize.indexOf(npm) === 0) {
  identifier = { type: 'npm', script: customize.substring(npm.length + 4) };
} else if (customize.indexOf(typescript) === 0) {
  identifier = { type: 'typescript', tsconfig: customize.substring(typescript.length + 6) };
}
```

Besides different `type` values (and the undocumented settings they use), there are other settings in the documentation that seem to be out of date compared with the source here: https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/contrib/tasks/node/taskConfiguration.ts#L28-L506
For example, the following properties are in the code but not the docs. I'm not sure if all of these correspond to user-facing config settings, but I assume at least some do:
* CommandOptions.shell.quoting
* PresentationOptions.showReuseMessage
[...]

PR #4135, opened 2020-11-18
Update tasks-appendix.md

steamwings:
Include a note that additional options for tasks.json are viewable with the Trigger Suggestions command.

[A similar change was discussed](https://github.com/Microsoft/vscode-docs/issues/2480#issuecomment-544710599) but seems to be missing.

My personal motivation for this change is that it would have helped me discover [the `group` option](https://github.com/microsoft/vscode/pull/65973) under a task's `presentation` without having to refer to [old issues](https://github.com/Microsoft/vscode/issues/47265).


gregvanl:
Thanks, good suggestion.

PR #4138, opened 2020-11-19
Changed  Python: to Jupyter: for new notebooks

manashty:
In newer versions, the following can not be found: "Python: Create New Blank Jupyter Notebook" and "Jupyter: Create New Blank Jupyter Notebook" is available instead.

### DIFF
--- a/docs/python/data-science-tutorial.md
+++ b/docs/python/data-science-tutorial.md
@@ -40,7 +40,7 @@ Visual Studio Code and the Python extension provide a great editor for data scie
 
    ![Selecting a python environment](images/data-science-tutorial/anaconda-environment.png)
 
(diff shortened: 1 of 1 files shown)


gregvanl:
@manashty Thanks for noticing this change. I think there are several other commands that now have the "Jupyter:" prefix instead of "Python:". For example, "Jupyter: Open in Notebook Editor" .
@JoshuaPartlow

manashty:
Thank you for your reply. Correct. I think some of the screenshots may need to change too.

JoshuaPartlow:
Thanks @manashty for submitting a PR on this instance. We have a PR in the works to update these, including the screenshots. You should see it corrected soon.

PR #4145, opened 2020-11-25
Fix typo in cmake-linux.md

Marcythm:
Just fixing a typo.


gregvanl:
Thanks! It might be a few days until this is published live due to the US holidays.

PR #4146, opened 2020-11-26
Fix Disconnect After Debugging setting description

tksh164:
The "Disconnect After Debugging" setting's description was a little bit different in the current version. I confirmed with Bridge to Kubernetes v1.0.120201118.

image


gregvanl:
Thanks!

PR #4149, opened 2020-11-27
Remove outdated colors

usernamehw:
I don't think people are using vscode or creating a color theme that supports 3 years old outdated colors.


gregvanl:
Thanks for the tidy up!

PR #4165, opened 2020-12-05
Include webpack step in your-first-extention.md

aussiDavid:
`$ yo code` now asks to bundle the source code with webpack.
Update the documentation to include the new step

`? Bundle the source code with webpack? No`


gregvanl:
Thanks!

PR #4178, opened 2020-12-13
Fix "Marketplace" capitalization

hadenpf:
Very minor chore, but the inconsistency bothered me.

## Changes

- Changed all instances of "MarketPlace" to "Marketplace" which is the official spelling as found everywhere else it is referenced, as well as in the Marketplace itself.


gregvanl:
Agreed. Thanks for doing this!

PR #4180, opened 2020-12-14
Add documentation about `markdownDescription` JSON schema extension

charrondev:
Currently the json language service escapes markdown in description tooltips unless it is specified as `markdownDescription` instead of `description`.

|With `description`|With `markdownDescription`|
|---|---|
|image|image|

Since this is a VSCode specific extension of JSON Schema, I figured it could use a spot in the documentation.

I did notice the other VSCode specific schema extension, `defaultSnippets`, [does have documentation](https://code.visualstudio.com/docs/languages/json#_define-snippets-in-json-schemas).

I would have had no idea how to do this if I hadn't gone digging into the language server's code.

https://github.com/microsoft/vscode-json-languageservice/blob/8a284774bf942ffd883d7f307f960a5255abfad3/src/services/jsonHover.ts#L72

```ts
markdownDescription = markdownDescription || s.schema.markdownDescription || toMarkdown(s.schema.description);

```

As you can see, it will check for a `markdownDescription` property in order to actually render markdown in descriptions. Otherwise it will pass the description contents to the `toMarkdown` function which escapes all markdown formatting.

### DIFF
--- a/docs/languages/json.md
+++ b/docs/languages/json.md
@@ -186,6 +186,28 @@ Use the property `defaultSnippets` to specify any number of snippets for the giv
 
 Note that `defaultSnippets` is not part of the JSON schema specification but a VS Code-specific schema extension.
 
+### Use rich formatting in hovers
+
+VSCode will use the standard `description` field from the [JSON Schema specification](https://json-schema.org/latest/json-schema-core.html#rfc.section.7) in order to provide information about property on hover and during autocomplete.
+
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @charrondev Thank you for your PR. Before I can merge it, you need to sign the Contributor License Agreement.

charrondev:
@gregvanl Done!

PR #4183, opened 2020-12-15
Update clone repository link

michaelpaul:
The link to see more details in **Use Clone Repository in Container Volume** section is pointing to **Open an existing folder in a container**.

I believe it should point to **Open a Git repository or GitHub PR in an isolated container volume** since we are talking about how to "Clone Repository" and not "Open an existing folder...".


gregvanl:
Agree. Thanks!

PR #4195, opened 2020-12-22
Install TS as a dev dependency

mrakgr:
I ran into into this issue and had to spend a while resolving it. The top Google search said to install TS globally, but that did not work for me.

### DIFF
--- a/api/working-with-extensions/bundling-extension.md
+++ b/api/working-with-extensions/bundling-extension.md
@@ -27,6 +27,12 @@ This will install webpack and update your extension's `package.json` file to inc
 npm i --save-dev ts-loader
 ```
 
+It might also be good to install Typescript as a dev dependency if the loader cannot find it.
+
+```bash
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @mrakgr Thank you for submitting a PR and your interest in the VS Code documentation. If someone is working on an extension written in TypeScript, they should already have TS installed globally or local to their project.
I'll update the paragraph to make it clearer that ts-loader is only needed if working on a TypeScript extension.

gregvanl:
Updated with https://github.com/microsoft/vscode-docs/commit/dc3e89d0984186b92f3f48345bf0656ad1eeebdf

mrakgr:
I did have Typescript installed globally, but it did not work for me. I have no idea at this point, maybe I should have tried  restarting the terminal?

gregvanl:
Not sure, getting terminals to inherit the right environment can be pesky

PR #4198, opened 2020-12-26
cpp.md: Apply the principle of least astonishment

skycommand:
There are several redundant and confusing elements on the live version of this page:

1. A link whose label reads "the C++ section of the VS Code website" but leads to "Using GCC with MinGW"! Please do not assume that all Windows developers (even veteran ones) know what GCC and MinGW are. And anyway, "Using GCC with MinGW" is hardly the equivalent of "the C++ section of the VS Code website."
2. A screenshot of a web browser showing VS Code website. I'm sure the developers know what a web browser is.
3. Two different sections, each listing a portion of all C++ doc articles. I'm confident that one section containing links is better.

To fix these issues, this proposed version recommends:

1. Merge the "Documentation" section into the "Tutorials" section
2. Add the missing link to "Configure IntelliSense for Cross-compilation"
3. Remove the non-intuitive link
4. Remove the redundant screenshot, which serves no real purpose

### DIFF
--- a/docs/languages/cpp.md
+++ b/docs/languages/cpp.md
@@ -154,23 +154,17 @@ This has been a very simple example to help you get started with C++ development
 
 Get started with C++ and VS Code with tutorials for your environment:
 
-- [GCC on Windows via MinGW](/docs/cpp/config-mingw.md)
-- [Microsoft C++ on Windows](/docs/cpp/config-msvc.md)
-- [GCC on Linux](/docs/cpp/config-linux.md)
+- [GCC on Windows](/docs/cpp/config-mingw.md)
 - [GCC on Windows Subsystem For Linux](/docs/cpp/config-wsl.md)
+- [GCC on Linux](/docs/cpp/config-linux.md)
+- [Microsoft C++ on Windows](/docs/cpp/config-msvc.md)
 - [Clang/LLVM on macOS](/docs/cpp/config-clang-mac.md)
 - [CMake Tools on Linux](/docs/cpp/cmake-linux.md)
-
-## Documentation
-
-You can find more documentation on using the Microsoft C/C++ extension under the [C++ section](/docs/cpp) of the VS Code website, where you'll find topics on:
-
 - [Debugging](/docs/cpp/cpp-debug.md)
 - [Editing](/docs/cpp/cpp-ide.md)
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @skycommand Thank you for submitting a PR and your interest in the VS Code documentation.
We just recently updated the docs/languages/cpp topic and some of the changes you point out are intentional.
We noticed that some users were not finding the C++ TOC and so that is the reason to add a screenshot of the web site.
We also found that many users were searching for MinGW explicitly so that is why we added "MinGW" to the tutorial link.
And thank you for pointing out that the docs/cpp link redirects to the MinGW tutorial. It is meant to select the first topic in the TOC and now goes to the second after a recent reordering. I'll fix that.
We're watching how the updated topic performs looking at ratings and verbatim and will keep your suggestions in mind. So far the updated page is doing well, CSAT up from 55% to close to 80% positive.

skycommand:
@gregvanl Hi. You seem quite serious the readability of this page. So, do you mind if I ask you a few questions?

> We noticed that some users were not finding the C++ TOC and so that is the reason to add a screenshot of the web site.

1. As long as you are supplying them with a whole in-page TOC, does that even matter?
2. I missed the TOC anyway. Have you thought of perhaps drawing attention to it via markup, e.g. a magenta box and an arrow?

> CSAT up from 55% to close to 80% positive

Why not go all the way up to 100%? Your site has C++ contents in two different places, because it doesn't support a third-level hierarchy. That sounds like one problem causing more problems.  Why not adopt the tried-and-proven layout of Microsoft Docs? It provides a hierarchical relational layout that developers have known and loved for almost three decades now.

gregvanl:
@skycommand Good suggestion to highlight the C++ section in the TOC screenshot. Right now the image isn't very eye catching.
We've been thinking about moving the docs/languages/cpp topic to be the first overview page of the C++ TOC but haven't yet given the very high pageviews where it is currently located (it is always in the top 1-3 pages per month)

skycommand:
Oh, my. The plot thickens. A website that doesn't even serve ads is worried about its page views! (Hint: A simple redirect can resolve that concern.)

This adds up to the older mysteries: Why a product that bears the Visual Studio brand name and the phrase "Copyright © 2021 Microsoft" doesn't host its help contents on docs.microsoft.com.

PR #4200, opened 2020-12-28
Fix '|' character in keybindings markdown

XyyF:


gregvanl:
Hi @XyyF Thank you for submitting a PR. This renders fine on the VS Code website at https://code.visualstudio.com/docs/getstarted/keybindings#_conditional-operators 

image

But as you have noticed the content doesn't render on GitHub or with the VS Code Markdown preview.
I'll merge your PR and see how it looks when published. There is a chance I may need to revert your change if it doesn't look correct on code.visualstudio.com.

gregvanl:
Unfortunately, your fix doesn't work with the website rendering so I need to revert the change.

image

But thanks again for noticing and taking the time to submit a PR!

PR #4206, opened 2020-12-29
Add a hint to the watcher problem.

mgsergio:
Makes it a bit easier to spot the problem with file watcher.


gregvanl:
Thanks! Good suggestion to try filtering out appropriate directories first.

PR #4208, opened 2020-12-29
Fix example configuration for docker-compose task

abaart:
The example configuration for a docker-compose task does not align with the reference outlined below it. I fixed this by adding the `dockerCompose` property to the example.
I ran in to this issue myself during configuring of a project that used docker-compose. The errors of the VSCode linter pointed me in the right direction, declaring that "up" was not an allowed property; Even though I just copied the example from the docs.


bwateratmsft:
@ghogen looks good to me. Thank you @abaart!

ghogen (approved):

PR #4212, opened 2020-12-31
Tasks: Add note of multiple errors on one line

hirooih:
I reported this issue as bug on https://github.com/microsoft/vscode/issues/111134 (open Nov, 23, 2020).
It was closed as a duplicated of https://github.com/microsoft/vscode/issues/105159 (opened Aug. 22, 2020).
And it was closed as "it is by design. We only allow one problem per location."

I still think this is a critical bug, but I have to follow your team's decision.

But this is a very important "feature" which VS code users should know.  It should be documented.

I also hope someone read this will create fix of this issue, which I could not make it.

### DIFF
--- a/docs/editor/tasks.md
+++ b/docs/editor/tasks.md
@@ -1009,3 +1009,9 @@ Alternatively, you can override a task's shell with the `options.shell` property
     },
     ...
 ```
+
+### Why does the problem matcher display only one of multiple messages on a line?
+
+Because it is by design. We only allow one problem per location.
+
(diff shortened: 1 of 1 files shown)


alexr00:
Thanks for the PR @hirooih! I think this is a valuable note to add to the documentation, so I added it in the "Defining a multiline problem matcher" section since that's where I expect folks to end up needing this info: ea5dbf9c826d3810d27616673172d8ef002016a5

PR #4213, opened 2020-12-31
Changed write/read to read/write

adriancuadrado:
It feels so awkward to read "write/read" instead of "read/write" that I had to change it.


gregvanl:
Fair enough. Thanks

PR #4215, opened 2021-01-02
Remove duplicate 'the's

noritada:
Some 'the the's are found in the document. This PR provides corrections of them.
I hope these are not intentional.
Thanks!


gregvanl:
Thanks! 
To really fix vscode-api.md, which is autogenerated from the VS Code API vscode.d.ts, you'll need submit a PR to update https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts

noritada:
@gregvanl Thank you very much for your suggestion! I will do that.

PR #4216, opened 2021-01-03
Update html.md

RedstonecraftHD:
Changed programming to scripting because HTML is NOT a Programming Language.


gregvanl:
Hi @RedstonecraftHD Thanks for submitting a PR and your interest in the VS Code documentation. I think saying "HTML programming" is fine in this context. "Scripting" is also problematic since JavaScript is the scripting part of HTML/CSS/JavaScript used on web pages.

PR #4226, opened 2021-01-07
Update advanced container doc

bamurtaugh:
Update snippet based on https://github.com/microsoft/vscode-remote-release/issues/4178 and to make the doc more consistent with our definitions in https://github.com/microsoft/vscode-dev-containers.

@chrmarti I interpreted the comment in the other thread and the Dockerfiles I've seen to mean that we can reduce this code snippet to just the 3 lines you mention, but if I'm mistaken and we still need `ENV DEBIAN_FRONTEND=dialog`, please just let me know.


chrmarti (approved):
That's correct. The env variable only exists within the shell that runs the `RUN` statement, no need to set it to dialog later.

PR #4233, opened 2021-01-13
Fix conditional operators table

rbutera:
Was not rendering the line for 'Or' correctly on GitHub due to GitHub's handling of pipe characters

### DIFF
--- a/docs/getstarted/keybindings.md
+++ b/docs/getstarted/keybindings.md
@@ -256,7 +256,7 @@ Operator | Symbol | Example
 -------- | ------ | -----
 Equality | `==` | `"editorLangId == typescript"`
 Inequality | `!=` | `"resourceExtname != .js"`
-Or | `||` | `"isLinux || isWindows"`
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @rbutera Thanks for submitting a PR and your interest in the VS Code docs. 
While the table doesn't render correctly on GitHub, it does render correctly on the VS Code website https://code.visualstudio.com/docs/getstarted/keybindings#_conditional-operators
You can also reviewhttps://github.com/microsoft/vscode-docs/pull/4200, which had to be reverted

rbutera:
@gregvanl #4200 was rejected because it broke on the VSCode website. I think that this change will render correctly on both the VSCode website and GitHub. Please reopen and review

gregvanl:
@rbutera I think you are right. I tested your change building the site locally and at it appears to render correctly.
We should also escape the next column:
` Or | <code>\|\|</code> | `"isLinux`<code>\|\|</code>`isWindows"``
I'll merge your PR and follow to add the second code block.
Thanks for figuring out the right incantation!

PR #4234, opened 2021-01-13
[docs] update settings docs with proper macOS path

Aghassi:
The path is technically correct as is, but can't be pasted into a terminal without the escape character.

My use case was I copied the path and ran `code` on the path in my terminal to open the editor with the code. It did not work without the escaped path.


gregvanl:
Hi @Aghassi Thank you for your PR and interest in the VS Code documention.
Copy and paste is a nice workflow but I think we should leave the folder name as it appears in the terminal.

PR #4251, opened 2021-01-19
official arch linux package exists

soloturn:
for arch linux based distributions like arch linux, manjaro an official package does exist nowadays.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -102,12 +102,12 @@ sudo zypper refresh
 sudo zypper install code
 ```
 
-### AUR package for Arch Linux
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @soloturn Thank you for submitting a PR and your interest in the VS Code documentation. The Open Source build you reference is not officially supported by Microsoft so we shouldn't reference it in the VS Code documenation.

soloturn:
@gregvanl i see what you mean, "official" in the sense of officially part of the arch package repository, versus arch user repository, is not something you consider on your website :) understandable. created #4281.

PR #4253, opened 2021-01-19
fix JS debug terminal link

sana-ajani:


gregvanl:
Fixed with https://medium.com/politically-speaking/why-did-trump-install-his-loyalists-at-the-pentagon-before-the-capitol-attack-a147c714cc55

PR #4259, opened 2021-01-21
Update versioncontrol.md

LesleyPhillips:
Don't use abbreviations or slang without defining where it fits in.


gregvanl:
Thanks! Good clarification

PR #4273, opened 2021-01-28
added troubleshooting for no input terminal

Jay-study-nildana:
There is a huge issue with the input entry with the newer versions of Mac. I spent at least 2 hours trying to understand why this is happening. 

Until the issue is fixed, I think, this troubleshooting must be included in the main documentation as this is where new developers, especially students would land, when trying to run a simple hello world on their new Macs.


jureid:
@sean-mcmanus, @WardenGnaw - do you know if the recent change with 1.2-insiders, which replaces the "externalConsole" setting with "console" in launch.json fixes this issue?

WardenGnaw:
Version 1.2.0-insiders2 added `console` to `cppvsdbg` which is the windows C++ debugger. 

Mac users can only use `vsdbg` and is not affected by those changes.

jureid:
@WardenGnaw got it, thanks for the reply! So then the troubleshooting steps for no input terminal are required and should be added to the docs?

WardenGnaw:
The changes are correct from the https://github.com/microsoft/vscode-cpptools/issues/5079 issue. 

This is because when the debugger is requesting a window to show, VS Code is somehow eating that request. Using a tasks, allows the user to see the request to use the terminal window under VS Code.

gregvanl:
Thanks @Jay-study-nildana and everyone who reviewed!

PR #4277, opened 2021-01-29
fix .ssh/ path

hidekuro:
In some cases, relative paths doesn't work correctly. (For example, if .zprofile is a symlink)
The solution is to use an absolute path.


gregvanl:
Thanks! Good suggestion

PR #4279, opened 2021-01-30
Update to add a precautionary step

itanex:
Step helps the tutorial user to insure that the Docker Context is supported for the **Remote-Containers: Try a Sample...** command

> ??
> If I have targeted the wrong branch for this change proposal, please advise and I will correct it.
> ??

This is in reference to [#4386](https://github.com/microsoft/vscode-remote-release/issues/4386)

### DIFF
--- a/docs/remote/containers-tutorial.md
+++ b/docs/remote/containers-tutorial.md
@@ -60,6 +60,18 @@ The Remote Status bar item can quickly show you in which context VS Code is runn
 
 ![Remote - Containers commands](images/containers-tutorial/remote-containers-commands.png)
 
+### Verify Context
+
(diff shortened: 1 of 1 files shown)


bamurtaugh:
Thanks so much for opening this, @itanex! 

Since users may not know what a Docker context is yet (and likely won't need to right away if they don't run into this error), my initial thought is perhaps we could move this section as a note after users select **Remote-Containers: Try a Sample...*, in case they run into trouble.

Perhaps at the end of the "Wait for the container to build" section, right before "Check the container?" Would that sound reasonable to you?

itanex:
@bamurtaugh I thought about that. 

I came at this tutorial following the `Containers` section of the documentation. Though this specific tutorial/doc is in a different section. 

If this doesn't change how you think this should change I welcome the suggestion and will commit my local variant that complies with that request.

gregvanl:
Thanks!

itanex:
Glad to contribute.

PR #4282, opened 2021-01-30
sysctl.conf is systemd not arch

soloturn:
sysctl.conf is here for all distributions with systemd-207 or greater. see e.g. https://archlinux.org/news/deprecation-of-etcsysctlconf/ (arch), https://manpages.debian.org/stable/procps/sysctl.conf.5.en.html (debian).

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -211,7 +211,7 @@ When you see this notification, it indicates that the VS Code file watcher is ru
 cat /proc/sys/fs/inotify/max_user_watches
 ```
 
-The limit can be increased to its maximum by editing `/etc/sysctl.conf` (except on Arch Linux, read below) and adding this line to the end of the file:
+The limit can be increased to its maximum by editing `/etc/sysctl.d/*.conf` by adding this line to the end of the file:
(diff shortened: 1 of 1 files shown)


docs/setup/linux.md
```diff
@@ -211,7 +211,7 @@ When you see this notification, it indicates that the VS Code file watcher is ru
 cat /proc/sys/fs/inotify/max_user_watches
 ```
 
-The limit can be increased to its maximum by editing `/etc/sysctl.conf` (except on Arch Linux, read below) and adding this line to the end of the file:
+The limit can be increased to its maximum by editing `/etc/sysctl.d/*.conf` by adding this line to the end of the file:
```
joaomoreno, line 214: We should probably be more descriptive here and clarify what is said on https://gist.github.com/tbjgolden/c53ca37f3bc2fab8c930183310918c8c:

- There might already be a file limiting the value, so that file should be used
- Otherwise, a new file should be created

joaomoreno (changes requested):

zxl223:
****ccz

PR #4293, opened 2021-02-02
Update Go extension documentation to reflect the recent change

hyangah:
We are now using the gopls as the backend of the most language features.

https://blog.golang.org/gopls-vscode-go

This PR removes mention of old tools and replaces the two pngs with one gif to demonstrate all the intellisense features.
Also updates the extension image.


gregvanl:
Thanks for keeping this topic up to date!

PR #4300, opened 2021-02-05
Fix link to multi root workspaces

yekver:


gregvanl:
Hi @yekver Thanks for pointing out the broken link and submitting a PR. The link is actually supposed to point to the H2 in the same topic, rather than the multi-root workspaces article. The '-' between "multi" and "root" doesn't render a good link on the website. I'll fix shortly.
Thanks again for noticing!

gregvanl:
Fixed with https://github.com/microsoft/vscode-docs/commit/3e7edf7eab1f346138ee4d5c861b63cca0672d68
Thanks!

PR #4311, opened 2021-02-09
update typescript-tutorial.md to avoid confusion on change build output

effectivewang:
…avoid confusion on rerun

According to https://www.typescriptlang.org/docs/handbook/tsconfig-json.html,  When input files are specified on the command line, tsconfig.json files are ignored.

See Issue: https://github.com/microsoft/vscode/issues/116171


gregvanl:
Thanks! Good idea to be more explicit about not adding the source file.

PR #4316, opened 2021-02-09
Update to ReadMe

Martinnilsen99:
I helped a friend on a Windows PC running through this installation, and she did not select win_32 for threads, but kept the POSIX option. Could be nice to clarify this. 
I am on a Mac so can't confirm that "win_32" is the exact name, but am pretty sure it was/is.

### DIFF
--- a/docs/languages/cpp.md
+++ b/docs/languages/cpp.md
@@ -67,7 +67,7 @@ We will install Mingw-w64 via the SourceForge website. You can use this [Mingw-w
 
    ![Choose x86_64 on the MinGW settings page](images/cpp/choose-x86-64.png)
    
-3. For **Threads** select **win_32** and then click on **next**. 
+3. For **Threads** select **win_32** and then select **Next**. 
 
 4. On the **Installation Folder** page, use the default location for the **Destination folder**. Copy the location as you will need it later.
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @Martinnilsen99 Thank you for submitting a PR and taking an interest in the VS Code documentation.
I'm not sure switching from the threads default is correct. 
Did your friend have issues when using the "posix" option? 
I believe "posix" is the more general threading model and should work correctly on Windows with the Architecture set to "x86_64".

Martinnilsen99:
Well, she was not able to run her code at all through a terminal in vscode when first installed with posix. The installation steps were followed several times. When switched to win_32, and reopening vscode - the code ran. Not sure whether the win_32 option is correct, but it did work for her. As far as I know it may have been another factor involved, and that this option should be left on default. I thought posix was standard for unix-like operating systems, but do not hold knowledge about Windows on that matter.

gregvanl:
Hi @Martinnilsen99, thanks for responding. I think it is unlikely that the threads model setting was your friend's issue.
I've got a query out to the C++ extension team but so far they think using the default should work fine.
Closing this PR but will reopen if I learn more.

PR #4322, opened 2021-02-11
It now defaults to PowerShell even on Win7

fireattack:
Tested on 1.53.1

### DIFF
--- a/docs/editor/integrated-terminal.md
+++ b/docs/editor/integrated-terminal.md
@@ -50,7 +50,7 @@ Key|Command|
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @fireattack Thank you for submitting at PR and your interest in the VS Code documentation.
I'm not sure your change is correct and when I look at the VS Code source code, I still see a check for Windows 10.
Might you have terminal.integrated.shell.windows set in your settings.json?

fireattack:
> Hi @fireattack Thank you for submitting at PR and your interest in the VS Code documentation.
> I'm not sure your change is correct and when I look at the VS Code source code, I still see a check for Windows 10.
> Might you have terminal.integrated.shell.windows set in your settings.json?

Nope, I noticed my terminal suddenly became Powershell on after updated to 1.53.1. I then created a brand new profile with portable VS Code and can confirm it by default uses PS now (with no `terminal.integrated.shell.windows` set up).

Please see the screenshot below:

image

gregvanl:
@fireattack Thanks for confirming. I've reached out to a VS Code engineer and will reply back here on whether its a "feature" or a bug. :-)

gregvanl:
You are absolutely correct. Thanks for pointing this out!

PR #4324, opened 2021-02-13
Update extensions.md

yusufsaygili:
Document this extension is not available


gregvanl:
Thanks for calling this out and fixing the topic.

PR #4328, opened 2021-02-15
Update suggested images for AZP configuration

giordano:
`macos-10.13` isn't available anymore and [should migrate](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software) to at least `macos-10.14`, but perhaps it's better to just move to the latest versions of all images, as done in the GitHub Actions example.


joaomoreno:
Thanks!

PR #4369, opened 2021-03-03
Setting File Access Lists

ucheNkadiCode:
Responded to user request for more information on how to add file permissions to access files on a host machine from within a container. I went ahead and revamped the document.

**Important**: I chose to change the title, so I have also since renamed the root file. This means that the URL will change. I need to know AS SOON AS the link is active because this will affect our dockerfile scaffolding https://github.com/microsoft/vscode-docker/issues/2747.

Question: Should the DateApproved metadata be updated to indicate when the doc was last edited? Or should we keep it as April 2020?
Thank you!


docs/containers/debug-python.md
```diff
@@ -232,4 +232,4 @@ When you select **Docker: Add Docker Files to Workspace** for Django or Flask, w
 
 Learn more about:
 
-- [Configuring a non-root user in your container](/docs/containers/python-user-rights.md)
+- [Configuring a non-root user in your container](/docs/containers/python-configure-containers.md.md)
```
bwateratmsft, line 235: Is `.md.md` right?

gregvanl:
@ucheNkadiCode We will need to add a redirect on the website for this file rename so that people who already have bookmarked the old URL will go to the new URL. I can add the redirect in the private website repo before this PR is merged.
Per the DateApproved, yes, it is good to update the date so users know the topic is getting attention and is not stale.

gregvanl:
@ucheNkadiCode @ghogen I'll do the merge once everyone is signed off so I can control the timing relative to the website redirect.

ucheNkadiCode (commented):
> @ucheNkadiCode We will need to add a redirect on the website for this file rename so that people who already have bookmarked the old URL will go to the new URL. I can add the redirect in the private website repo before this PR is merged.
> Per the DateApproved, yes, it is good to update the date so users know the topic is getting attention and is not stale.

Thank you @gregvanl! I have updated the DateApproved. I appreciate the help with the website redirect! We are gearing up to have our aka.ms links point to the new link whenever that goes live. Please either you or @ghogen send me the exact link so I don't mess it up!

docs/containers/quickstart-python.md
```diff
@@ -55,7 +55,7 @@ After verifying your app runs properly, you can now Dockerize your application.
 
     >**Tip**: You may also enter the path to a folder name as long as this folder includes a `__main__.py` file.
 
-1. If **Python: Django** or **Python: Flask** was selected, specify app port for local development. Django defaults to port 8000 while Flask defaults to port 5000; however, any unused port will work. We recommend selecting port 1024 or above to mitigate security concerns from [running as a root user](/docs/containers/python-user-rights.md).
+1. If **Python: Django** or **Python: Flask** was selected, specify app port for local development. Django defaults to port 8000 while Flask defaults to port 5000; however, any unused port will work. We recommend selecting port 1024 or above to mitigate security concerns from [running as a root user](/docs/containers/python-configure-containers.md).
```
ghogen, line 58:
```suggestion
1. If **Python: Django** or **Python: Flask** was selected, specify app port for local development. Django defaults to port 8000, while Flask defaults to port 5000; however, any unused port will work. We recommend selecting port 1024 or above to mitigate security concerns from [running as a root user](/docs/containers/python-configure-containers.md).
```

docs/containers/python-configure-containers.md
```diff
@@ -137,41 +136,69 @@ Exception has occurred: PermissionError
...
 ```
 
-> **Note**: This is just one example of how to add permissions. There are many ways to do so and it is your responsibility give the least permission possible to specific files and folders.
+> **Note**: This is just one example of how to add permissions in a container. There are many ways to do so and it is your responsibility give the least permission possible to specific files and folders.
 
-### Binding to an inaccessible port
+### Invalid file permission on the host (Linux)
 
-If your container starts and stops immediately after `kb(workbench.action.debug.start)` without producing logs in the **Debug Console**, we can try to diagnose the issue through these steps:
+In the previous example, we showed you how to add permissions to a file or folder on the container as a non-root user. However, if you are trying to access a folder **on the host machine** from within the container as a non-root user, the user ID or group ID in the container must have access to the files on the host. To solve this issue in Linux, you may need to set file access control lists (setfacl).
```
ghogen, line 150:
```suggestion
In the previous example, we showed you how to add permissions to a file or folder on the container as a non-root user. However, if you are trying to access a folder **on the host machine** from within the container as a non-root user, the user ID or group ID in the container must have access to the files on the host. To solve this issue in Linux, you might need to set file access control lists (setfacl).
```

docs/containers/python-configure-containers.md
```diff
@@ -137,41 +136,69 @@ Exception has occurred: PermissionError
...
+# Creates a non-root user and adds permission to access the /app folder
 RUN useradd appuser && chown -R appuser /app
 
-# Adds permission to appuser (non-root) for access to the /extra folder
+# Adds permission for appuser (non-root) to access the /extra folder
 RUN chown -R appuser /extra
 ```
 
-> **Note**: This is just one example of how to add permissions. There are many ways to do so and it is your responsibility give the least permission possible to specific files and folders.
+> **Note**: This is just one example of how to add permissions in a container. There are many ways to do so and it is your responsibility give the least permission possible to specific files and folders.
```
ghogen, line 146:
```suggestion
> **Note**: This is just one example of how to add permissions in a container. There are many ways to do so, and it is your responsibility give the least permission possible to specific files and folders.
```

ghogen (approved):
Approving with a few small suggestions.

gregvanl:
@ucheNkadiCode The update has been published. 
The new URL is https://code.visualstudio.com/docs/containers/python-configure-containers
I've added a redirect from /docs/containers/python-user-rights to /docs/containers/python-configure-containers so that no bookmarks will be broken.

docs/containers/python-configure-containers.md
```diff
@@ -137,41 +136,69 @@ Exception has occurred: PermissionError
...
+
+```python
+PermissionError: [Errno 13] Permission denied: '/share/logs/log.txt'
+```
+
+In order to give access to a non-root user `appuser` from within the container, follow these steps:
+
+1. From the container command line, run one of these commands:
+
+    ```powershell
```
TheYarin, line 162: why powershell and not shell/bash?

TheYarin: same for line 173

docs/containers/python-configure-containers.md
```diff
@@ -137,41 +136,69 @@ Exception has occurred: PermissionError
...
+
+If you have a folder named `/share` on your host machine and try to access this folder before the access control list are properly set, you will likely receive this error:
+
+```python
+PermissionError: [Errno 13] Permission denied: '/share/logs/log.txt'
+```
+
+In order to give access to a non-root user `appuser` from within the container, follow these steps:
+
+1. From the container command line, run one of these commands:
```
TheYarin, line 160: Maybe it's worth mentioning how to enter a container's commandline

TheYarin, line 160: Are you sure the user/group ID will be the same if the container is removed and re-created? That's a common scenario when operating with `docker-compose up` &  `docker-compose down`

docs/containers/python-configure-containers.md
bwateratmsft: That's just telling the docs site how to syntax highlight this chunk of code. Those commands are actually shell/bash commands.

TheYarin: Yeah I know, this might seem a little petty but why not mark the block specifically as `bash`/`sh`? I'm pretty sure whichever markdown engine is used, it supports this language.

docs/containers/python-configure-containers.md
ucheNkadiCode: @bwateratmsft or @TheYarin Would it be sufficient to say 

1. From the **Containers** pane in the Docker View, right click on your running container, select **Attach Shell**, and run one of these commands:

TheYarin: hmm, I think the user will be doing all this permissions configuration in a production environment, so docker actions from within the developer's VSCode might not be relevant.

(Unless managing your remote docker host from within VSCode is really easy, of which I might not be familiar with)

ucheNkadiCode: @karolz-ms or @bwateratmsft, do you know about this?

bwateratmsft: I don't know for sure, but I would think that the uid/gid would stay the same, since it comes from the image, not the container. I think?

docs/containers/python-configure-containers.md
karolz-ms: ... or  `shell`  I agree with @TheYarin

docs/containers/python-configure-containers.md
karolz-ms: Good catch @TheYarin Best practice is to use and explicit UID in this case: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user

TheYarin: Yeah, I'm expecting  this scenario will mostly occur as a remote container. This _might_ happen in the development environment if it's a linux environment (when mounting a volume on my Windows dev environment it didn't cause any permissions errors), but honestly, I got a feeling that a lot of developers don't properly test their application in a containerized setup on their dev machine, because the developer probably has all the dependencies set up already so it's easier for him to just run the app outside the container.

But considering what @karolz-ms said below about the best practice regarding UID, I think mentioning how to enter a container's commandline is kind of irrelevant.

PR #4370, opened 2021-03-03
Provide direct link to JSON schemas

ugultopu:
So that it is easier to understand where the schemas are coming from.


gregvanl:
Hi @ugultopu Thank you for submitting a PR and you interest in the VS Code documentation.
I'm not sure most readers would want to see the underlying schema for the schema API. I believe most would link to a specific schema file rather than wanting to see the overall catalog file.

PR #4371, opened 2021-03-04
Fixed spelling mistake in Reset Intellisense Database

jtilles:
Changed **C_Cpp: Reset IntelliSense Database**  to **C/Cpp: Reset IntelliSense Database** to reflect the actual command in VS Code


gregvanl:
Hi @jtilles Thanks for pointing this out and submitting a PR.
Should the prefix actually be "C/C++:"? Here's what I see in the Command Palette:
image

jtilles:
@gregvanl You're correct, that is my mistake, it should be C/C++ instead of C/Cpp.  I fixed it and committed the change!

gregvanl:
Thanks!

PR #4374, opened 2021-03-08
Update Mac OS Setup M1 support documentation

hkennyv:
As of version [1.5.4](https://code.visualstudio.com/updates/v1_54), VS Code provides stable Apple Silicon builds 🎉. I think the Mac OS setup documentation should reflect this awesome achievement.

Any tips on word-smithing would be greatly appreciated :)

### DIFF
--- a/docs/setup/mac.md
+++ b/docs/setup/mac.md
@@ -97,6 +97,6 @@ If VS Code doesn't update once it restarts, it might be set under quarantine by 
 
 ### Does VS Code run on Mac M1 machines?
 
-Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip. Currently, only [Insiders](/insiders) macOS ARM64 builds are available.
+Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip starting with version [1.54](/release-notes/v1_54.md) and [Insiders](/insiders) builds.
 
 ![macOS ARM64 Insiders build](images/mac/arm64-insiders.png)


docs/setup/mac.md
```diff
@@ -97,6 +97,6 @@ If VS Code doesn't update once it restarts, it might be set under quarantine by
 
 ### Does VS Code run on Mac M1 machines?
 
-Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip. Currently, only [Insiders](/insiders) macOS ARM64 builds are available.
+Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip starting with version [1.54](/release-notes/v1_54.md) and [Insiders](/insiders) builds.
```
gregvanl, line 100: I think we can just drop the second sentence and don't need to mention the release number.

gregvanl:
Hi @hkennyv Thanks for noticing and submitting a PR. I left a comment that I think we can just remove the second sentence.

docs/setup/mac.md
hkennyv: Thanks @gregvanl ! There's only one sentence, but I _think_ I caught what you're saying. Check my edits?

docs/setup/mac.md
```diff
@@ -97,6 +97,6 @@ If VS Code doesn't update once it restarts, it might be set under quarantine by
 
 ### Does VS Code run on Mac M1 machines?
 
-Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip starting with version [1.54](/release-notes/v1_54.md) and [Insiders](/insiders) builds.
+Yes, VS Code supports macOS ARM64 builds that can run on Macs with the Apple M1 chip.
 
 ![macOS ARM64 Insiders build](images/mac/arm64-insiders.png)
```
hkennyv, line 102: should we remove this image as well? The image isn't in the repo, but if you view the site [here](https://code.visualstudio.com/docs/setup/mac), it looks like that image is outdated since it shows "Intel" and "ARM64". It seems like the naming has changed to be "Intel Chip" and "Apple Silicon" (see image below from insiders download page):

image

hkennyv (commented):
@gregvanl thoughts on this?

hkennyv:
@gregvanl thanks for the reviews! I've removed the stale image and mentioned the universal and and platform specific builds in my latest commits

PR #4391, opened 2021-03-12
Explicit UID Instructions

ucheNkadiCode:
Thanks once again to @TheYarin for giving us great feedback to improve the product! We have now changed our Python scaffolding to include the best practice of setting an explicit UID when creating a non-root user.

We changed our docs to Bash rather than powershell.

Lastly, we updated our instructions to help the user setfacls from their host with the knowledge that they already have an explicit UID declared

### DIFF
--- a/docs/containers/python-configure-containers.md
+++ b/docs/containers/python-configure-containers.md
@@ -158,7 +158,7 @@ PermissionError: [Errno 13] Permission denied: '/share/logs/log.txt'
 In order to give access to a non-root user `appuser` from within the container, follow these steps:
 
 1. Copy the explicit UID from your Dockerfile (`5678` in the example above).
-2. From the **host machine's** command line, run one of these commands:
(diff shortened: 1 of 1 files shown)


docs/containers/python-configure-containers.md
```diff
@@ -157,20 +157,10 @@ PermissionError: [Errno 13] Permission denied: '/share/logs/log.txt'
...
-    id -u appuser
-
-    # Or to find a specific user's GID
-    id -g username
-    ```
-
-1. Copy the output from the container command line.
-1. From the **host machine's** command line, run one of these commands:
+1. Copy the explicit UID from your Dockerfile (`5678` in the example above).
+2. From the **host machine's** command line, run one of these commands:
```
ghogen, line 161:
```suggestion
1. From the **host machine's** command line, run one of these commands:
```
use autonumbering

ghogen (approved):
Approved with one suggestion

PR #4397, opened 2021-03-16
Update extension-manifest.md

fiveisprime:
Added the education category to the manifest reference.


gregvanl:
@fiveisprime This category won't be available until 1.55 goes out, correct? 
We can either move this PR to the vscode-docs vnext branch or merge it after the 1.55 release goes out in two weeks.

fiveisprime:
It's already supported so it's safe to merge. The completion is the only thing that will come in 1.55.

The category is live: https://marketplace.visualstudio.com/search?target=VSCode&category=Education&sortBy=Installs

PR #4402, opened 2021-03-17
Bridge to Kubernetes tutorial: Clarify port info

ghogen:


ghogen:
@gregvanl This is ready now.

gregvanl:
@ghogen We'll need to verify that the change to use { } doesn't mess up the generated handlebar files.

Issue #3277, opened 2019-12-03
Clearer Documentation on Activating Interpreters: Global vs Virtual Environment


Anti-Distinctlyminty:
Related to https://github.com/microsoft/vscode-python/issues/8852

In my case it would've helped if the documentation was clearer about the difference between selecting an interpreter that is a regular python install vs selecting a virtual environment. Because all the available interpreters are in a single list I was under the assumption that they behave the same way when selected. E.g. if I were to switch to the Python 3 install, fired up a new terminal and did a `pip install` then the packages would be installed to Python 3, but this is not necessarily the case. But if I had selected a virtual environment then this _would_ be the case. Some demarcation of these behaviours in the documentation would be welcome.

JoshuaPartlow:
Hi @Anti-Distinctlyminty - Thanks for the feedback. I'll see if I can find a way to make it clearer in the https://code.visualstudio.com/docs/python/environments topic.

Anti-Distinctlyminty:
Hi @JoshuaPartlow ,
I think that when I was searching for information on this topic, I always ended up on the page that you linked to, so I think that's probably enough. If I think of any other locations I'll let you know.

PR #4411, opened 2021-03-22
Clarification for global environment package installation

JoshuaPartlow:


JoshuaPartlow:
Addresses https://github.com/microsoft/vscode-docs/issues/3277

PR #4419, opened 2021-03-24
Move tutorials to top of doc

jureid:
Can we move the tutorials section to the top of this doc so that people can more easily find the content for their platform?

BTW, we're going to start moving away from recommending MinGW for customers who want to target Linux from Windows and instead recommend WSL for those workflows. I'm going to propose changes to the GCC on Windows doc, and then we'll probably want to pick a different example for this landing page. I can help with that.


gregvanl:
Hi @jureid After the tuning to the topic done in December and the dramatic increase in CSAT (56% -> 77%), I'm reluctant to take changes to the ordering of this topic.
Let's do some planning around the C++ topics before we make changes here.

Issue #43, opened 2015-11-19
Add a PHP section under Languages


egamma:
Add the information from the release notes for how to setup PHP validation in there:

Language - PHP
VS Code is now using the official PHP linter (php -l) for PHP language diagnostics. VS Code will now stay current with PHP linter improvements.

There are two new settings to control the PHP linter:

php.validate.executablePath: points to the PHP executable on disk. Set this if the PHP executable is not on the system path.
php.validate.run: controls whether the validation is triggered on save (value: "onSave") or on type (value: "onType"). Default is on save.

gregvanl:
Found 10 verbatim on PHP in a 3 month window (November 2015-January 2016)

gregvanl:
Be sure to call out PHP debugging through extension in the debugging topic (maybe Common Questions).

gregvanl:
Topic added with https://github.com/Microsoft/vscode-docs/commit/ff59c0fc1c67e358dba8a3fe10e7f158f0446d25

PR #4435, opened 2021-03-28
add missing Education as valid value

jorenbroekema:
Follow-up from:
- https://github.com/microsoft/vscode/pull/118589
- https://github.com/microsoft/vscode-docs/pull/43

Seems like this one instance was missed ;)


gregvanl:
Thanks!

PR #4437, opened 2021-03-29
Automated use of developer command prompt in tasks

meatGUY:
Hi,

Recently I had to configure VS Code to use Microsoft C++ compiler in remote development scenarios and I have found the following way of automated initialisation of developer command prompt during execution of build task.

I am not sure that PR is the right way to share such an idea but I thought it would be better to see the changes in place, so it will be easier to discuss them or reject them right away.

### DIFF
--- a/docs/cpp/config-msvc.md
+++ b/docs/cpp/config-msvc.md
@@ -339,6 +339,51 @@ If you see the error "The term 'cl.exe' is not recognized as the name of a cmdle
 
 You can always check that you are running VS Code in the context of the Developer Command Prompt by opening a new Terminal (`kb(workbench.action.terminal.new)`) and typing 'cl' to verify `cl.exe` is available to VS Code.
 
+## Remarks
+
+In certain circumstances it isn't possible to run VS Code from **Developer Command Prompt for Visual Studio** (for example in Remote Development through SSH scenario). In that case you can automate initialization of **Developer Command Prompt for Visual Studio** during the build using the following configuration:
+
+```
+{
+    "version": "2.0.0",
+    "windows": {
+        "options": {
+            "shell": {
+                "executable": "cmd.exe",
+                "args": [
+                    "/C",
+                    "\"C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/Common7/Tools/VsDevCmd.bat\"",
+                    "&&"
+                ]
+            }
+        }
+    },
+    "tasks": [
+        {
+            "type": "shell",
+            "label": "cl.exe build active file",
+            "command": "cl.exe",
+            "args": [
+                "/Zi",
+                "/EHsc",
+                "/Fe:",
+                "${fileDirname}\\${fileBasenameNoExtension}.exe",
+                "${file}"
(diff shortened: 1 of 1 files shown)


jureid:
@meatGUY thank you for sharing this workaround! Before merging this PR, I want to double check with our engineering team about another potential workaround, which would be using the full path to cl.exe and passing the system headers path as an argument. If you do those two things, I'm not sure it's necessary to use the Developer Command Prompt.

meatGUY:
Hi @jureid!

It is been a while since I created this PR. I made my progress much further that it was originally and created a [post](https://olegkarasik.wordpress.com/2021/06/29/building-testing-and-debugging-visual-studio-c-project-in-visual-studio-code/) with detailed instructions of how to configure more than just developer command prompt.

If the info from the post seems reasonable I think I can update the PR to include all the core information without a story part.

docs/cpp/config-msvc.md
```diff
@@ -339,6 +339,51 @@ If you see the error "The term 'cl.exe' is not recognized as the name of a cmdle
...
+```
+{
+    "version": "2.0.0",
+    "windows": {
+        "options": {
+            "shell": {
+                "executable": "cmd.exe",
+                "args": [
+                    "/C",
+                    "\"C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/Common7/Tools/VsDevCmd.bat\"",
```
jureid, line 355: Let's add a comment explaining that this is the path to VsDevCmd.bat and that it could change depending on the VS version. I know you have a note about that below, but if people copy and paste this into tasks.json and then leave the doc, they might not see it.

"// the path to VsDevCmd.bat depends on the version of Visual Studio you have installed.

docs/cpp/config-msvc.md
```diff
@@ -339,6 +339,51 @@ If you see the error "The term 'cl.exe' is not recognized as the name of a cmdle
...
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            }
+        }
+    ]
+}
+```
+
+>**Note**: The path to `VsDevCmd.bat` might be different depending on the Visual Studio version or installation path.
```
jureid, line 385: We can also tell people that they can find the path to VsDevCmd.bat by opening a command prompt and running `dir "\VsDevCmd*" /s`

jureid:
Hi @meatGUY! Sorry for the delay in getting back to you. Thank you for creating and sharing that post - really impressive stuff! Let's merge this PR to get the Developer Command Prompt workaround in the docs. I think the rest of the information in your post is awesome, but would be better off as its own topic / blog post rather than in this tutorial.

I left two comments. Do you want to make those changes and then we can merge the PR?

gregvanl:
Follow up commit https://github.com/microsoft/vscode-docs/commit/4802e19ae03b91d20f1043d4a7b2a0b0243d6ee6

meatGUY:
Hi @jureid, @gregvanl.

I can see the PR is already merged and docs are updated according to the comments. Thank you both!

PR #4446, opened 2021-03-31
Update working-with-javascript.md

fire015:
Update config name as VS code complains that `javascript.` is now deprecated.


gregvanl:
Thanks!

PR #4452, opened 2021-04-02
Update telemetry docs with the additional information

lramos15:
This updates the telemetry documentation to contain information regarding the `--telemetry` command that has existed since 2019 but been a bit hidden. This also includes @sbatten work around the telemetry APIs which allow extension authors to respect the telemetry setting without utilizing our npm module.

### DIFF
--- a/docs/getstarted/telemetry.md
+++ b/docs/getstarted/telemetry.md
@@ -125,7 +125,7 @@ If you have created a VS Code extension, you can use the VS Code telemetry infra
 
 Follow this guide to set up [Azure Monitor](https://docs.microsoft.com/azure/azure-monitor/learn/nodejs-quick-start) and get your Application Insights instrumentation key.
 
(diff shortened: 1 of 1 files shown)


docs/getstarted/telemetry.md
```diff
@@ -119,6 +125,10 @@ If you have created a VS Code extension, you can use the VS Code telemetry infra
 
 Follow this guide to set up [Azure Monitor](https://docs.microsoft.com/azure/azure-monitor/learn/nodejs-quick-start) and get your Application Insights instrumentation key.
 
+If you would prefer to not utilize the npm module it is still recommended that extension authors respect the user's decision about whether or not to send telemetry data via the `telemetry.enableTelemetry` setting by utilizing the `isTelemetryEnabled` and `onDidChangeTelemetryEnabled` API.
```
sbatten, line 128: The API also recognizes the CLI flag and that's not mentioned here. Since this blurb is for extension authors though, I don't think that part is important. To simplify this sentence, I would drop the setting callout itself and just focus on the API values.

lramos15: Makes sense. We don't actually surface the `--disable-telemetry` flag anywhere in this doc or in the `--help` do you think it's worth mentioning?

PR #4453, opened 2021-04-06
Update remote troubleshooting

bamurtaugh:
### Context

One of my goals this iteration is to make some updates to help with troubleshooting and triaging issues in remote-release.

1) I've noticed that we often first ask users for their logs. We don't currently ask for logs in the issue reporter, and there's some complexity since we have 3 different extensions. I've added a section about reporting quality issues to the remote troubleshooting doc, which largely includes how to grab the correct logs.

2) In addition to logging info, I've also included a couple other pieces of info in this new section based on common questions we often ask users, i.e.:

- Can they SSH into their machine from an external terminal?
- Does their WSL issue lie on the WSL side or the Remote - WSL side?

3) I wanted to update the doc with any tips or settings that folks commonly use to troubleshoot. I added a section about `Remote-SSH: Kill VS Code Server on Host` and was considering any others that could be useful for any of the remote extensions.

4) Once we merge in any changes to this troubleshooting doc, I thought it'd make sense to update the remote-release issue reporter (I believe a few of us have discussed this) in hopes of folks taking some of those initial logging and troubleshooting steps to report higher quality issues.

This was my initial idea for the updated reporter template (additions are bolded):

reporter

### Getting your input

I'd really like to get your input here, since I hope this effort could help you all in some way. If you have input on common troubleshooting steps, settings, or commands we should document more clearly, any of the updates I've included here, or even just the format and overall idea of updating the issue reporter and docs, I'd love to hear it. 

Thank you!

### DIFF
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -1078,7 +1078,7 @@ If you run into an issue with one of the remote development extensions, it's imp
 
 Each remote extension has a command to view its logs.
 
-You can get the Remote - SSH logs with `F1` -> `Remote-SSH: Show Log`. This is the same as the output from the SSH Output channel: `F1` -> `Output: Focus on Output View`, and select "Remote - SSH" from the dropdown. When reporting Remote - SSH issues, please also verify if you're able to SSH into your machine from an external terminal (not using Remote - SSH).
+You can get the Remote - SSH logs with `F1` -> `Remote-SSH: Show Log`. When reporting Remote - SSH issues, please also verify if you're able to SSH into your machine from an external terminal (not using Remote - SSH).
 
 Similarly, you can get the Remote - Containers logs with `F1` -> `Remote-Containers: Show Log`.
 
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -1068,6 +1068,26 @@ Extensions that access locally attached devices will be unable to connect to the
 
 ## Questions and feedback
 
+### Reporting quality issues
+
+If you run into an issue with one of the remote development extensions, it's important to collect the correct logs so that we'll be able to help [diagnose your issue](https://aka.ms/vscode-remote/issues/new).
+
+We recommend turning on trace level telemetry to get the most detailed logs: `F1` -> `Developer: Set Log Level...` -> `Trace`. From here, the type of logs depends on the context of your issue.
+
+You can get the Remote - SSH logs from the SSH Output channel: `F1` -> `Output: Focus on Output View`, and select "Remote - SSH" from the dropdown.
+
+You can get the Remote - Containers logs with `F1` -> `Remote-Containers: Show Log`.
+
+Similarly, you can get the Remote - WSL logs with `F1` -> `Remote WSL: Show Log`.
+
+If you're experiencing issues using other extensions remotely (i.e. other extensions aren't loading or installing properly in a remote context), it's helpful to grab the log from the "Remote Extension Host" output channel: `F1` -> `Output: Focus on Output View`, and select "Log (Remote Extension Host)" from the dropdown.
+
+> **Note**: If you only see "Log (Extension Host)," this is the local extension host, and the remote extension host didn't launch. This is because the log channel is created only after the log file is created, so if the remote extension host does not launch, the remote extension host log file was not created, and this is not shown in the output view. This is helpful information to include in your issue.
+
+### Remote question and feedback resources
+
+We have a variety of other remote resources:
+
 - See [Remote Development FAQ](/docs/remote/faq.md).
 - Search on [Stack Overflow](https://stackoverflow.com/questions/tagged/vscode-remote).
 - Add a [feature request](https://aka.ms/vscode-remote/feature-requests) or [report a problem](https://aka.ms/vscode-remote/issues/new).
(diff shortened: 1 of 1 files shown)


docs/remote/troubleshooting.md
```diff
@@ -1068,6 +1072,26 @@ Extensions that access locally attached devices will be unable to connect to the
 
 ## Questions and feedback
 
+### Reporting quality issues
+
+If you run into an issue with one of the remote development extensions, it's important to collect the correct logs so that we'll be able to help [diagnose your issue](https://aka.ms/vscode-remote/issues/new).
+
+We recommend turning on trace level telemetry to get the most detailed logs: `F1` -> `Developer: Set Log Level...` -> `Trace`. From here, the type of logs depends on the context of your issue.
```
chrmarti, line 1079: Each of the extensions comes with a log level setting (the core log level is not shared).

chrmarti (approved):

docs/remote/troubleshooting.md
bamurtaugh: I see - so does setting trace log level not matter, since the log level is taken care of once the user selects their specific output (i.e. "Remote-Containers: Show Log" will display the same info regardless of whether a user selected trace)?

Or, does setting trace level still matter, but users need to select this while connected to their remote (i.e. I need to be connected to a dev container, then set trace level, then can Remote-Containers: Show Log)?

roblourens: The first - extensions don't know about vscode's log level.

docs/remote/troubleshooting.md
```diff
@@ -163,6 +163,10 @@ If you encounter an error when connecting, you may need to enable socket forward
 
 If you are running into problems with VS Code hanging while trying to connect (and potentially timing out), there are a few things you can do to try to resolve the issue.
 
+**Server fails to start**
+
+If you get a message along the lines of "Could not establish connection to `server_name`: The VS Code Server failed to start," try running the command: `Remote-SSH: Kill VS Code Server on Host`. This will remove the server and can help in scenarios such as it was installed in a bad state if the host ran out of disk space.
```
roblourens, line 168: Running this command can fix a lot of issues, it's a good general troubleshooting tip

docs/remote/troubleshooting.md
```diff
@@ -1068,6 +1072,26 @@ Extensions that access locally attached devices will be unable to connect to the
 
 ## Questions and feedback
 
+### Reporting quality issues
+
+If you run into an issue with one of the remote development extensions, it's important to collect the correct logs so that we'll be able to help [diagnose your issue](https://aka.ms/vscode-remote/issues/new).
+
+We recommend turning on trace level telemetry to get the most detailed logs: `F1` -> `Developer: Set Log Level...` -> `Trace`. From here, the type of logs depends on the context of your issue.
+
+You can get the Remote - SSH logs from the SSH Output channel: `F1` -> `Output: Focus on Output View`, and select "Remote - SSH" from the dropdown. Also please verify if you're able to SSH into your machine from an external terminal (not using Remote - SSH).
```
roblourens, line 1081: There's one command `Remote-SSH: Show Log`

docs/remote/troubleshooting.md
bamurtaugh: Thanks for the explanation. I've removed the section about trace level telemetry since it doesn't seem helpful or relevant for reporting most issues here.

docs/remote/troubleshooting.md
bamurtaugh: I see this command helps a lot of folks in remote-release, so I figured it'd make sense to call out in some higher level section in this doc. 

I've reworded this section to be a bit more of a general troubleshooting tip (i.e. now titled "General troubleshooting: Remove the server," updated the explanation). 

If you have any other feedback on the wording or placement (i.e. I felt like under the `Troubleshooting hanging or failing connections` made sense, but open to feedback here), please let me know.

docs/remote/troubleshooting.md
bamurtaugh: Yes that should be included, good point. Updated to include `Remote-SSH: Show Log` so that it's consistent with the other extensions' show log commands. Still include the mention of the Remote - SSH output channel in case folks are uncertain if they're the same thing or prefer one method over the other.

docs/remote/troubleshooting.md
```diff
@@ -1076,13 +1076,11 @@ Extensions that access locally attached devices will be unable to connect to the
 
 If you run into an issue with one of the remote development extensions, it's important to collect the correct logs so that we'll be able to help [diagnose your issue](https://aka.ms/vscode-remote/issues/new).
 
-We recommend turning on trace level telemetry to get the most detailed logs: `F1` -> `Developer: Set Log Level...` -> `Trace`. From here, the type of logs depends on the context of your issue.
+Each remote extension has a command to view its logs. You can get the Remote - SSH logs with `F1` -> `Remote-SSH: Show Log`, which is the same as the output from the SSH Output channel: `F1` -> `Output: Focus on Output View`, and select "Remote - SSH" from the dropdown. Also please verify if you're able to SSH into your machine from an external terminal (not using Remote - SSH).
```
roblourens: I don't think it's necessary to mention both ways of getting to the same log.

bamurtaugh: Sounds good - I removed the second way & just kept the first one (so that it's consistent with the other extensions).

roblourens (approved):

PR #4456, opened 2021-04-07
Update ssh.md

tyrionlannistar:
The suggested way "sc config ssh-agent start=auto" does not work.
Error: Set-Content : A positional parameter cannot be found that accepts argument 'startup'.

### DIFF
--- a/docs/containers/ssh.md
+++ b/docs/containers/ssh.md
@@ -15,7 +15,7 @@ We recommend using the Visual Studio Code [Remote-SSH extension](/docs/container
 
 1. Configure `ssh-agent` on the **local** system with the **private** key file produced above.
 
-    * **Windows (OpenSSH):** The latest version(s) of Windows 10 include OpenSSH by default. There is a Windows service, `ssh-agent` that is disabled by default, and needs to be re-enabled and set to automatic start. From an admin command prompt, run `sc config ssh-agent start=auto` and `net start ssh-agent`. Then, do `ssh-add <keyfile>`.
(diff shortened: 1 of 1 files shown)


roblourens:
I think this is the command for cmd.exe, not powershell. I would rather describe a command to run than some buttons to click in a gui. I think it would be better to have documentation for powershell, would that have been more clear?

roblourens:
I realized this is the containers topic so I'll leave it up to @chrmarti or whoever wrote this, but this is the powershell script we show for SSH 
https://github.com/microsoft/vscode-docs/blob/main/docs/remote/troubleshooting.md#setting-up-the-ssh-agent

gregvanl:
Giving a mention to @karolz-ms, who most recently worked on the topic.

karolz-ms:
I like how the remote docs use PS built-in cmdlets to make the service start automatically, we should do the same in containers docs. Thanks @roblourens for the tip!

tyrionlannistar:
@roblourens  It would be great to add the remark, that it is a cmd and not a powershell command. There are many discussions out there, on how to start the service. I was not aware, that cmd and powershell commands differ. :)

gregvanl:
Closing this PR. Assuming @karolz-ms will work on a new PR with input from @tyrionlannistar.
Thanks @tyrionlannistar for pointing out the possible confusion.

karolz-ms:
Hmm, I was not really volunteering to do a new PR ☺️ @ghogen could I use your help here?

If not, no worries, just LMK, I can squeeze this in sometime next week

gregvanl:
@karolz-ms No worries, looks like there is enough context here and in https://code.visualstudio.com/docs/remote/troubleshooting#_setting-up-the-ssh-agent for @ghogen to put something together.

PR #4461, opened 2021-04-09
created doc for cpp intro videos

jureid:
@gregvanl I created the Intro Videos doc for C++ that we talked about a few weeks ago. This is not ready to merge yet, I need to fix a few things. I'd appreciate your help whenever you get the chance.

Things to fix:
1. What should I put for ContentId?
2. I put "Order: 1" because we want this to be first in the the C++ TOC. Is there another document I need to edit to add this to the TOC? And then should I increment the Orders of all the other C++ docs in the TOC by 1?
3. The iframes aren't rendering, do you know what I'm doing wrong?

### DIFF
--- a/docs/cpp/introvideos-cpp.md
+++ b/docs/cpp/introvideos-cpp.md
@@ -1,10 +1,9 @@
 ---
-Order: 1
 Area: cpp
 TOCTitle: Intro Videos
-ContentId:
+ContentId: a7a9837e-f515-4644-8dae-b5d44715e63f
 PageTitle: Introductory Videos for C++ in Visual Studio Code
-DateApproved:
+DateApproved: 4/13/2021
 MetaDescription: Get started with C++ in Visual Studio Code by watching these introductory videos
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @jureid, answers below.
1. A generated GUID. I use the Create Guid extension (nwallace.createguid) to create a new GUID and paste it into ContentId. It uniquely identifies the topic.
2. Yep, if you are inserting the videos as the first item in the TOC, you'll need to increase all the other topic Order numbers.
3. Your iframe entries look fine but you won't be able to see them until they are on the staging server or website.

For testing, you could remove the Order: value and the new topic won't show up in the TOC. After the PR is merged, you will be able to review the page via the URL on the staging server. Once you are happy with how the page looks, we can add it to the TOC and sitemap for the rest of the world to enjoy.

jureid:
@gregvanl thank you! I added a GUID for Content Id and removed Order: value for testing purposes.

PR #4463, opened 2021-04-10
Fix link to #languages.registerColorProvider

wilhelmklopp:
👋 Hey! Just noticed this link points to an anchor that doesn't seem to be around anymore.

The working link should be: https://code.visualstudio.com/api/references/vscode-api#languages.registerColorProvider


gregvanl:
Thanks for noticing and submitting a PR!

PR #4464, opened 2021-04-10
Describe to use PPA priority

remcohaszing:
Without this, distributions may install their own version of `code`, which is likely outdated.

In ran into this issue in practive: https://www.reddit.com/r/pop_os/comments/mo1vv8/why_does_apt_downgrade_vscode/. The newly added instructions were the solution for me.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -56,6 +56,14 @@ sudo sh -c 'echo "deb [arch=amd64,arm64,armhf signed-by=/etc/apt/trusted.gpg.d/p
 rm -f packages.microsoft.gpg
 ```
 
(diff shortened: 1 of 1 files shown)


ItalyPaleAle:
Thanks @remcohaszing for pointing this out and proposing a PR! I have read the thread you linked to and it sounds like it's a not-too-uncommon issue.

Most distributions don't include `code` in their official repos (and we recommend they don't), however some do, such as Pop! OS. And this should help those users.

Because this impacts users on certain distributions only, I have moved this to the FAQ section where we cover some "edge cases". I would prefer not to put it in the middle of the Debian/Debian-based instructions because this won't apply to most users on Debian and Ubuntu (and derivatives), and so it might be best to not give them too much to worry about.

Let us know if this works for you. Otherwise, it's a #signoff from me and @gregvanl will be able to merge this in a few days.

remcohaszing:
That's fine with me :)

gregvanl:
Thanks!

PR #4465, opened 2021-04-11
Change compilerPath example to full path

Dibbyo456:
`compilerPath` should the full path to the compiler.


jureid:
This change looks good to me! We can merge this.

PR #4473, opened 2021-04-13
Added h2s and descriptions to intro videos docs, added intro videos to TOC

jureid:


docs/cpp/introvideos-cpp.md
```diff
@@ -7,12 +8,31 @@ DateApproved: 4/13/2021
 MetaDescription: Get started with C++ in Visual Studio Code by watching these introductory videos
 ---
 
-# Introductory Videos
+# Introductory Videos for C++
 
 Get started with C++ in Visual Studio Code by watching these introductory videos! These videos are designed to help you set up C++ IntelliSense and build and debug C++ projects in VS Code. After watching these quick tutorials, you'll be able to enjoy the rich C++ feature set VS Code has to offer.
 
+## Configure C++ IntelliSense in Visual Studio Code
```
gregvanl, line 15: You can take out the "in Visual Studio Code" in all three H2s.
These will cause line breaks in the right navigation list and isn't really necessary.

jureid: Good catch, thanks! Should be fixed now

Issue #4477, opened 2021-04-14
Update container doc based on updates


bamurtaugh:
Based on updates from this milestone (i.e. Simplify 'Clone Repository in Container Volume' command from the iteration plan), [our docs](https://code.visualstudio.com/docs/remote/containers#_quick-start-open-a-git-repository-or-github-pr-in-an-isolated-container-volume) will need an update.

Once we've confirmed any UI or functionality changes, I'll open a PR to update accordingly. 

cc @chrmarti

PR #4489, opened 2021-04-21
Update clone in volume information

bamurtaugh:
Updating the docs to reflect @egamma's latest updates to the Clone Repository in Container Volume flow.

- Update inspection command name to **Remote-Containers: Explore a Volume in a Development Container...**
- Add explanation and image for dev volumes section in Remote Explorer
- Update image for inspecting in Docker extension

I explored the other Remote - Containers docs (especially https://code.visualstudio.com/docs/remote/containers-advanced) to see if any content there needed to be updated, but the volume information there seems up-to-date (i.e. code-focused rather than impacted by the new command flow/UI). The [cloning quick start](https://code.visualstudio.com/docs/remote/containers#_quick-start-open-a-git-repository-or-github-pr-in-an-isolated-container-volume) also seemed pretty much unaffected. 

If there's anything else you think we should highlight regarding the new flow/functionality, please let me know. I considered adding extra info users may find interesting, i.e. how users can get more than one repo per volume, but since our docs are already rather long, I wasn't sure if that was necessary or would just get lost (and the command for creating a new volume is a Docker command `Docker volume create ...`, rather than something VS Code specific). 

Since these changes are live in Insiders but will ship with Stable, I think we can wait to push until the next Stable release. 

Fixes https://github.com/microsoft/vscode-docs/issues/4477.

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -198,11 +198,15 @@ To learn more about creating `devcontainer.json` files, see [Create a Developmen
 
 ## Inspecting volumes
 
-Occasionally you may run into a situation where you are using a Docker named volume that you want to inspect or make changes in. You can use VS Code to work with these contents without creating or modifying `devcontainer.json` file by selecting the **Remote-Containers: Inspect Volume in Container...** from the Command Palette (`kbstyle(F1)`).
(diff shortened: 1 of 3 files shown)


Chuxel (approved):

chrmarti (approved):

gregvanl:
Added the `don't merge` label as this should go out with the 1.56 updates.

PR #4493, opened 2021-04-23
fix(docs): corrects common typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to markdown docs only.

### DIFF
--- a/api/language-extensions/programmatic-language-features.md
+++ b/api/language-extensions/programmatic-language-features.md
@@ -383,7 +383,7 @@ export function activate(ctx: vscode.ExtensionContext): void {
 
 Allow the user to see all occurrences of a symbol in the open editor.
 
-![Select a symbol to highlight all occurences](images/language-support/document-highlights.gif)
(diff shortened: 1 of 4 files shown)


gregvanl:
Thanks!
The vscode-api.md fixes won't stick next time VS Code releases since the file is autogenerated from the vscode.d.ts Type Definition file. You can update that text with a PR to the VS Code product repo at https://github.com/microsoft/vscode to update https://github.com/microsoft/vscode/blob/main/src/vs/vscode.d.ts.

gregvanl:
Hi @plan-do-break-fix 
I mispoke, the API generation happens on each vscode-docs CI build so vscode-api.md can't be updated manually, but looks like you're up for submitting a PR to the product repo. Thanks!
The update to vscode-api.md won't happen until the next 1.56 release, sometime the first week of May.

plan-do-break-fix:
Yes, I think I can handle that. I appreciate your help.

PR #4498, opened 2021-04-27
Correct symbol navigation character

cmwendwa:
Go to symbol character is @ and not :.

### DIFF
--- a/docs/editor/editingevolved.md
+++ b/docs/editor/editingevolved.md
@@ -81,7 +81,7 @@ Some [languages](/docs/languages/overview.md) also support jumping to the type d
 
 ## Go to Symbol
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @cmwendwa Thank you for submitting a PR and showing interest in the VS Code documentation.
You are correct that '@' is the default Go to Symbol character:
image
The sentence is pointing out that by adding a ':', you can sort the dropdown by category:
image
Do you think this sentence should be clearer?

cmwendwa:
@gregvanl Yeah, the sentence can be more clearer. The screen shot below the sentence also just has `@` . What would make it clearer is maybe splitting the  content  so that  
"typing `:` " comes after the current screen shot and there's a different screenshot for `@:`.

PR #4499, opened 2021-04-27
Update extension-manifest.md

irvnriir:
fixed outdated . btw, i think References is more important to keep up to date, than shorts like /api/get-started/extension-anatomy.md#extension-manifest


gregvanl:
Thanks!
I didn't understand your second sentence. What do you mean my "References"?

PR #4502, opened 2021-04-28
fix python autoComplete configuration setting key

mikolajkalwa:
`python.autocomplete.addBrackets` doesn't exist.


JoshuaPartlow:
Hi @mikolajkalwa - Thank you for contributing to the docs. I've reviewed your change and it will be merged shortly.

PR #4515, opened 2021-05-04
Update extension-anatomy.md

HoseynHeydari:
Use `devDependencies` instead of `dependencies` as the main part of `package.json`.
At [extension-anatomy](https://code.visualstudio.com/api/get-started/extension-anatomy) `dependencies` introduced as the main part of `package.json`.
`yo` (version 4.1.0) recently generates a sample code for `Hello World` extension. Generated code contains `devDependencies` instead of `dependencies` in `package.json` file.
If `yeoman` is updated enough seems better to change this part of documentation.


gregvanl:
Thanks!

PR #4519, opened 2021-05-04
Update "end user" usage

josola:
Good morning!

This PR:

- Updates instances of "end user"/"end users" to "user"/"users".

Why:

- To follow [this article](https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/u/user-end-user) in the Microsoft Style Guide.
- "User" was used, rather than "audience" or "customer", as most of the instances being replaced live in a technical context.

Let me know if you have any questions or edits.


gregvanl:
Thanks!

PR #4530, opened 2021-05-06
Fix minor mistake in v1.56 release notes about KaTeX

Simran-B:
It says single dollar signs but shows `$$`. Also fix a typo.


gregvanl:
Thanks!

PR #4539, opened 2021-05-08
Remove stray character from Django tutorial

guarani:
It looks like a stray `f` was added to one of the code snippets so this removes that.


gregvanl:
Thanks!

PR #4540, opened 2021-05-08
Remove reference to unassigned keyboard shortcut

guarani:
The Save All command (`workbench.action.files.saveAll`) does not have a default keyboard shortcut (see [Visual Studio Code Key Bindings](https://code.visualstudio.com/docs/getstarted/keybindings#_file-management)) so is rendered `unassigned` in the tutorial. This doesn't make much sense to the reader of the tutorial, so I think it'd be best to remove the reference entirely.

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -183,7 +183,7 @@ To create a minimal Django app, then, it's necessary to first create the Django 
     ]
     ```
 
-1. Save all modified files with `kb(workbench.action.files.saveAll)`.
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @guarani Thanks for pointing this out. It looks like as some point the default keybinding (Ctrl+K S) for  `workbench.action.files.saveAll` moved to the global `saveAll` command.

image

Rather than remove the keyboard shortcut, can you change it to `saveAll`?

PR #4543, opened 2021-05-11
Compile VScode extension before running

flyme2bluemoon:
When using TypeScript to create a VScode extension, the first compile needs to be done in the command line. Otherwise, it does not compile properly. Without running `npm run compile`, the following error is produced. Optionally, `npm run watch` can also be used to achieve the same goal.

```
Activating extension 'undefined_publisher.helloworld' failed: Cannot find module '/Users/mshen/Developer/helloworld/out/extension.js'
Require stack:
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/loader.js
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-amd.js
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-fork.js.
-
```


gregvanl:
Hi @flyme2bluemoon I tried to reproduce your build error and I wasn't able to.
My steps were:
1. Create a new TypeScript extension project and select npm as the package manager. This causes npm install to run during creation.
2. From the main menu, Terminal > Run Build Task

The project built fine and I was able to Run > Start without Debugging to see the basic extension template running.

flyme2bluemoon:
Hey @gregvanl,

The error is only produced when a command is run. For instance, using the default boiler plate code, running the Hello World command from the command palette throws the error.

I believe the error is because the Typescript source code hasn't been transpiled into Javascript yet. Running `npm run compile` compiles the typescript code from the src directory into the out directory. However, the F5 key or the Run Build Task doesn't seem to run the Typescript compiler.

I have attached a video showing the error to hopefully help you reproduce it. I will also try this out on a Linux VM later to see if there might be some other issues.

https://user-images.githubusercontent.com/37197710/118332134-996da480-b4d7-11eb-83f1-628e268e0400.mov

gregvanl:
Hi @flyme2bluemoon Thanks for recording a video. There looks to be a terminal error when you Run Build Task and VS Code tries to run npm watch. 
Do you have any modified integrated terminal settings, which could be causing the task to fail? You can check by opening the Settings editor (Code > Preferences > Settings on macOS) and use the `@modified` filter.
You could also try running VS Code with the default settings by temporarily using a new user data directory.
```code helloworld --user-data-dir <dir>``` where <dir> is any temporary folder you want to use.
I'd be curious to hear if you see this on another machine with a clean installation of VS Code.

flyme2bluemoon:
Hey @gregvanl,

I have looked through my VScode settings and it turns out that the `terminal.integrated.inheritEnv` was set to false. Changing it to true seems to fix this issue and builds properly. Should I include this note in a new pull request? Additionally, I noticed that my zshrc config file is no longer sourced by the vscode integrated terminal which is an even bigger problem for me.

gregvanl:
That would explain why you see a difference between using Run Build Task and running from the command line. Glad that the integrated build is working now.
Thank you for offering, but we don't usually document tutorials with non-default settings so I don't think we'd need a new PR.
As to your .zshrc file no longer being picked up by the integrated terminal, there were some recent changes around `shellArgs` and the new terminal profiles feature. I know the recent 1.56.2 release had a fix to prefer using `shellArgs` if it was set for backwards compatibility.
Have at look at https://vscode-docs-main.azurewebsites.net/updates/v1_56#_terminal for information about profiles and issue https://github.com/microsoft/vscode/issues/123159 for the recent fix to see if this might explain your issue.

gregvanl:
Closing this PR now since running the build from the command line isn't a required step for the tutorial.
If you are still having trouble with your .zshrc settings being picked up, you can create or search for an issue in the VS Code product repo at https://github.com/microsoft/vscode/issues.
Thank you again for providing the nice video to help sort out the workflow.

PR #4547, opened 2021-05-13
docs: add gitlab ci example

SethFalco:
This updates the Continuous Integration documentation to include an example for GitLab CI.

This is based on the configuration here:
https://gitlab.com/Elypia/magick-image-reader/-/pipelines/302079622

---

This might not be ideal as installing packages in the pipeline isn't optimal, but it's a solution that works on the shared runners without creating ones own Docker image.

An alternative solution could be to make a Docker image just for testing Visual Studio Code Extensions, similarly to how there are containers for testing in Chromium. Like this, it could be based on a node:buster image that just installed the packages to allows us to run a headless instance of Visual Studio Code.

### DIFF
--- a/api/working-with-extensions/continuous-integration.md
+++ b/api/working-with-extensions/continuous-integration.md
@@ -209,8 +209,10 @@ before_script:
 
 test:
   script:
-    - apt update; apt install -y libasound2 libgbm-dev libgtk-3-0 libnss3 xvfb;
-      xvfb-run -a npm run test;
+    - |
+      apt update
+      apt install -y libasound2 libgbm-dev libgtk-3-0 libnss3 xvfb
+      xvfb-run -a npm run test
 ```
 
 ### GitLab CI automated publishing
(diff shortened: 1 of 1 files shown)


api/working-with-extensions/continuous-integration.md
```diff
@@ -197,6 +197,44 @@ In our example, the condition has three checks:
...
+
+```yml
+image: node:12-buster
+
+before_script:
+  - npm install
+
+test:
+  script:
+    - apt update; apt install -y libasound2 libgbm-dev libgtk-3-0 libnss3 xvfb;
```
connor4312, line 212: This would be easier to read if [split](https://docs.gitlab.com/ee/ci/yaml/script.html#split-long-commands). `;` if executed in the shell will also not do what you want most likely

connor4312 (approved):

gregvanl:
Thanks!

PR #4556, opened 2021-05-19
Fix #4555 Don't override the sudoers file by adding a user

### LINKED ISSUE #4555: Remote container docs: Demo code for adding a user will destroy sudoers file
The docs for adding a non-root user to your dev container suggests the following:

```Dockerfile
RUN groupadd --gid $USER_GID $USERNAME \
    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
    #
    # [Optional] Add sudo support. Omit if you don't need to install software after connecting.
    && apt-get update \
    && apt-get install -y sudo \
    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
    && chmod 0440 /etc/sudoers.d/$USERNAME
```

This will override and destroy the sudoers file. Use ">>" to append to the file instead of replacing it.

hegerdes:
While redirecting the output of echo `$USERNAME ALL=\(root\) NOPASSWD:ALL` the usage of ">" will override the sudoers file and destroy it. Use ">>" to append to the file

### DIFF
--- a/docs/remote/containers-advanced.md
+++ b/docs/remote/containers-advanced.md
@@ -471,7 +471,7 @@ RUN groupadd --gid $USER_GID $USERNAME \
     # [Optional] Add sudo support. Omit if you don't need to install software after connecting.
     && apt-get update \
     && apt-get install -y sudo \
-    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @hegerdes Thanks for creating a PR and your interest in the VS Code documentation.
This code block is creating a new file in the container so I think using the single pipe character is fine. 
https://linuxize.com/post/how-to-add-user-to-sudoers-in-ubuntu/ follows the same approach for creating a new user file.

PR #4573, opened 2021-05-27
Fix typo in "Build, test, and diagnose" section

EduardMaghakyan:
To configure the lint tool one should be using `go.lintTool` setting.


gregvanl:
Thanks!

PR #4582, opened 2021-05-31
Add missing f-string prefix

### LINKED ISSUE #4581: Missing f-string prefix in Python Django Tutorial
https://code.visualstudio.com/docs/python/tutorial-django#_define-models
String in line
`        return "'{self.message}' logged on {date.strftime('%A, %d %B, %Y at %X')}"`
must be prefixed with `f`.

[microsoft/python-sample-vscode-django-tutorial/blob/master/hello/models.py#L11](https://github.com/microsoft/python-sample-vscode-django-tutorial/blob/master/hello/models.py#L11) is OK.

kerbrek:
fixes #4581

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -666,7 +666,7 @@ class LogMessage(models.Model):
     def __str__(self):
         """Returns a string representation of a message."""
         date = timezone.localtime(self.log_date)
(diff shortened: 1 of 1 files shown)


JoshuaPartlow:
Hi @kerbrek - Thanks for catching the prefix was missing and submitting a PR to re-add it. Looks like it was accidentally removed in https://github.com/microsoft/vscode-docs/pull/4539.

Issue #4580, opened 2021-05-31
Document Remote-Containers adoption of trusted workspaces


chrmarti:
https://github.com/microsoft/vscode-remote-release/issues/5067

https://github.com/microsoft/vscode-internalbacklog/blob/main/proposals/core/trusted-workspace-dev-containers.md

# Docker deamon running locally + WSL

## Reopen folder in container

Requires trusting the local (or WSL) folder. The user will be asked to trust the local (or WSL) folder before the window reloads. (Exceptions: Clicking on a recent entry and 'Open Folder in Container' will ask after the window reloads - if trust is not already given.)

## Attach to existing container

User is asked to confirm that attaching means they trust the container. This is only confirmed once.

![image](https://user-images.githubusercontent.com/9205389/120207992-54fb3c00-c22d-11eb-89be-60368f49b63b.png)

## Clone repository in a volume

User is asked to confirm that cloning a repository means they trust the repository. This is only confirmed once.

![image](https://user-images.githubusercontent.com/9205389/120208302-9d1a5e80-c22d-11eb-9da4-929ace884180.png)

## Inspect volume

Starts in restricted mode and the user can trust the folder inside the container.

# Docker deamon running remotely

This implies trusting the machine the Docker daemon runs on. There are no additional prompts to confirm (only those listed for the local/WSL case above). This will be in documentation only.

PR #4585, opened 2021-06-02
Document Remote-Containers adoption of workspace trust

bamurtaugh:
Fixes https://github.com/microsoft/vscode-docs/issues/4580

@chrmarti @Chuxel I created a separate "Trusting your Workspace" section after the various sections about opening folders and cloning repos into volumes since users would run into workspace trust at/by this point. 

Another flow I thought of was adding the workspace trust points to each affected section (i.e. rather than an entire "Trusting your Workspace" section, add a description to the reopening folder section, cloning volume section, inspecting volume section, etc.). But I wasn't sure if this would feel repetitive and if it'd make more sense to keep all of the workspace trust in one section. 

Let me know your thoughts on location, wording, etc! Happy to adjust since I wasn't sure which flow would work best (or if another flow/location entirely might be better).


chrmarti:
Having all in one place looks good to me.

chrmarti (approved):

Chuxel:
LGTM - Though one question. Do we want to cross reference to the remote extensions from the main workspace trust article as well?

As Codespaces factors this in, there's a similar consideration there.

Chuxel (approved):

gregvanl:
This will go live once Workspace Trust is available in Stable so labeling with `don't merge` for now.

bamurtaugh:
Thanks for the reviews! @Chuxel I could see that being useful (since remote and trusting remotes could be a potentially big topic in the realm of workspace trust).

In reviewing the main Workspace trust doc, I'm trying to envision where it might make sense. Maybe somewhere like https://code.visualstudio.com/docs/editor/workspace-trust#_trusting-a-workspace since it describes "If you trust the authors and maintainers of a project, you can trust the project's folder on your local machine," and we could reference somewhere in/right at the end of this section that this also applies to trusting from a remote/in remote workflows?

chrmarti:
@bamurtaugh I'm adding `https://aka.ms/vscode-remote-containers-trust` to use on a Learn More button on the dialogs. Could you update that once this has gone live? Currently linking to the content of this PR, hope that's ok for Insiders.

Posted updated screenshots: https://github.com/microsoft/vscode-remote-release/issues/5136#issuecomment-854534310

bamurtaugh:
@chrmarti Sounds good! I took note of this to remind me, so when we merge in this PR, I'll update the aka.ms link to the live link rather than the PR.

I also updated this PR with the latest screenshots ("Got it" phrasing).

PR #4590, opened 2021-06-05
Add new dev container properties

bamurtaugh:


docs/remote/devcontainerjson-reference.md
```diff
@@ -104,6 +106,20 @@ Variables can be referenced in certain string values in `devcontainer.json` in t
...
 
+## Lifecycle scripts
+
+When creating and working with a dev container, you may need different commands to be run at different points in the container's lifecycle. The below are properties that shape how the container is created and takes shape over time, and the order in which they are run (i.e. `onCreateCommand` will run after `initializeCommand`).
+
+| Property | Description |
+|----------|-------------|
+| `initializeCommand` | A command to run locally before anything else. If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell. |
+| `onCreateCommand` | A command to run when creating the container. If this is a single string, it will be run in a shell. If this is an array of strings, it will be run as a single command without shell. <br> You may want to have some script steps happen before VS Code attaches (`onCreateCommand`) and some after (`postCreateCommand`). But, you can change this with `waitFor` to include even waiting for `postStartCommand` if you have something you want to happen every time. |
```
chrmarti, line 116: - 'before/after VS Code attaches' is somewhat misleading. I think the difficulty in explaining this is that we have split up `postCreateCommand` in a way that only makes sense when you have the prebuild feature.
- `waitFor` doesn't change if/when the commands are run, it only affects at what time the VS Code UI may connect to the dev container. By default it can connect after `updateContentCommand`. I'd keep the documentation for `waitFor` in its own row/paragraph.

docs/remote/devcontainerjson-reference.md
```diff
@@ -104,6 +106,20 @@ Variables can be referenced in certain string values in `devcontainer.json` in t
...
+| `updateContentCommand` | Reserved for future use. |
```
chrmarti, line 117: `onCreateCommand` is similar in that it does not add anything to `postCreateCommand` until we get the prebuild feature. Maybe document both as 'reserved'? Or explain a bit the future use? Have we added them to the JSON schema too early? @Chuxel

Chuxel: By default we should `waitFor` the `onCreateCommand` - so there is value here in that you can have something run that VS Code should wait for, and then something else that can happen after.  `updateContentCommand` is the weird one.

docs/remote/devcontainerjson-reference.md
Chuxel: Yeah see my other comment... I think we can say that by default we `waitFor` the `onCreateCommand`, but not `postCreateCommand`, but this can be changed using the `waitFor` property.

bamurtaugh: Thanks for reviewing!

I removed the sentence about "You may want to have some script steps..." as it may not be entirely clear given the extent of the context here. I also moved `waitFor` to its own section below the table, rather than as an entry in the table, and I mentioned users may have VS Code `waitFor` the `onCreateCommand`. 

@chrmarti @Chuxel Please let me know if these changes seem accurate and logical to you.

docs/remote/devcontainerjson-reference.md
chrmarti: @Chuxel `waitFor` currently defaults to `updateContentCommand`. Let me know if that should change.

bamurtaugh: I tried to word the doc in a way that could apply to any default (`By default, you may have VS Code `waitFor` the `onCreateCommand`...`), but if we'd rather change the wording here and/or default of `updateContentCommand`, happy to adjust.

Chuxel: Yeah updateContentCommand is fine as the default. So we could probably say:

By default, the `waitFor`  property is set to `updateContentCommand`. This allows you to use  use `onCreateCommand` or `updateContentCommend` for steps to it need to have happen before VS Code connects while still using `postCreateCommand` for steps that can happen behind the scenes afterwards.

We could probably say that it `updateContentCommand` is reserved for future scenarios where the source code may be updated sometime during the create process to disambiguate its purpose a bit without signaling specifically its purpose.

bamurtaugh: I like your wording ideas, and the way to disambiguate `updateContentCommand` for now. Updated both the `waitFor` and `updateContentCommand` descriptions, let me know what you think.

Chuxel: Looks great - one other thought is we could include an anchor link to the lifecycle section in the devcontainer.json propeprty section to help raise visibility to that since I do think its quite useful.

Chuxel (approved):
Looks great - one other thought is we could include an anchor link to the lifecycle section in the devcontainer.json property section to help raise visibility to that since I do think its quite useful. But otherwise sounds great!

bamurtaugh:
Thanks @chuxel! Added the following right before the devcontainer.json property table:

> You may need different commands to be run at different points in the container's lifecycle. The [lifecycle scripts](#_Lifecycle-scripts) section later in this doc describes how some of the below properties may shape the container over time.

@gregvanl I believe this doc is ready for your review and then merging if you think it is acceptable.

PR #4602, opened 2021-06-10
Fix default pytest path

The-Compiler:
The default path is "pytest", not "py.test":
https://github.com/microsoft/vscode-python/blob/f6100335a5260967031a2a1ee3684f701e95a932/package.json#L1837

Originally introduced in https://github.com/microsoft/vscode-python/commit/5b3cd6ee51754621c7be776c50157214d22a55da as "pytest".


gregvanl:
Looks right to me.

image

Thanks!

Jayboy989:
I bought this computer from SAM's it was one of their computers. It had windows Professional 64bit 86bit  on it and XBOX all free to play . I sent my computer to the shop because it would not come on. When I got it back it had WINDOWS HOME 64 bit on it. All I want is my computer back

Jayboy989:
I don't know what ENTERPRISE is I won't windows Professional work back!

PR #4605, opened 2021-06-10
Add Terraform & HCL language identifiers

radeksimko:


gregvanl:
Hi @radeksimko Thank you for submitting a PR and your interest in the VS Code documentation. 
Given that VS Code doesn't support this language ID out of the box and there seem to be several extensions using different language IDs for Terraform files (`tf`, `terraform`), we'd rather not take this PR.
Extension users can find the appropriate Terraform language ID by reviewing the extension's Feature Contributions page in the Extension view Details pane or by finding the language ID in the Change Language Mode dropdown.

radeksimko:
> Given that VS Code doesn't support this language ID out of the box

Do you mind briefly explaining on what does "out of the box" mean in this context?

> there seem to be several extensions using different language IDs for Terraform files (tf, terraform), we'd rather not take this PR.

That was the main motivation for that PR actually. We (HashiCorp; the company behind both Terraform & HCL) maintain [an extension](https://github.com/hashicorp/vscode-terraform) ourselves. And while we are happy for other extensions to be created and maintained, I think it would be helpful if they all used the same language IDs, so that they can integrate with language server(s) more easily.

gregvanl:
By "out of the box", I mean that VS Code doesn't doesn't recognize the .tf or .hcl file extensions nor ship with any basic language features such as syntax highlighting or bracket matching.  For example, without an extension installed, if you create a file `blah.tf` and copy in some Terraform code, the editor treats the file type as Plain Text.

One option would be to reach out to other open-source Terraform extension authors and propose they use the same language IDs.

PR #4618, opened 2021-06-16
Correct quickstart-python.md prompt order

kevbutler:
Updated the order of the prompts after selecting `Python: Django` as the app type when using `Docker: Add Docker Files to Workspace...`. For the Docker extension (v1.13.0), the correct order should have the prompt to select whether or not to include Docker Compose files as the last prompt, not the first.


ghogen:
@ucheNkadiCode can you confirm this is an ok change?

ucheNkadiCode:
Thanks for the change! Looks good!

PR #4565, opened 2021-05-26
Data Science TOC first iteration

tanhakabir:
Changes:
1. Add Data Science overview page for the Data Science topic
2. Moves Jupyter Notebook content and related images into the Data Science topic. (No actual content was changed, it's still webview version, I just a dragged and dropped)
2. Moves Data Science Tutorial content and related images into the Data Science topic. (No actual content was changed, I just a dragged and dropped)
3. Remove the Jupyter and Data Science tutorial articles under python and replaced them with just "Data Science" to point to the Data Science overview page.
3. Any shared images were copied over into the Data Science TOC
4. An empty page for Python Interactive to link to the Python topic's Python Interactive

**Important:** Need to merge in https://github.com/microsoft/vscode-website/pull/973 beforehand for this to work

PR #4623, opened 2021-06-16
Move Pytorch content from Python node to be under Data Science node

tanhakabir:
@jmew 

Related to https://github.com/microsoft/vscode-docs/pull/4565#issuecomment-849984050

We want to move the current [PyTorch article](https://code.visualstudio.com/docs/python/pytorch-support) under Python to the Data Science node while keeping the old url redirecting to this new page

**Need to merge first: https://github.com/microsoft/vscode-website/pull/985**


docs/datascience/pytorch-support.md
```diff
@@ -1,8 +1,8 @@
 ---
-Order: 9
-Area: python
+Order: 5
+Area: datascience
 TOCTitle: PyTorch Support
-ContentId: 7B5266AD-3D3E-491F-BD7C-B883C592D943
+ContentId: 265ac51a-200d-4016-96da-fa52d8e7ef35
```
gregvanl, line 5: You can leave the ContentId the same since we're just moving the topic.

docs/python/python-on-azure.md
```diff
@@ -1,5 +1,5 @@
 ---
-Order: 13
+Order: 12
```
gregvanl, line 2: You don't need to revert these changes but, in the future, you can save yourself time and leave the Order values the same. It's OK to have gaps in the Order values.

tanhakabir: oh LOL thanks Greg!

docs/datascience/pytorch-support.md
tanhakabir: will do!

PR #4628, opened 2021-06-22
Update markdown.md

PeterWone:
Mention the existence of printing extensions that can automate the rendering of Markdown.

### DIFF
--- a/docs/languages/markdown.md
+++ b/docs/languages/markdown.md
@@ -19,7 +19,9 @@ In addition to the functionality VS Code provides out of the box, you can instal
 
 > Tip: Click on an extension tile above to read the description and reviews to decide which extension is best for you. See more in the [Marketplace](https://marketplace.visualstudio.com).
 
-## Markdown preview
+## Markdown printing and preview
+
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @PeterWone Thank you for your PR and interest in the VS Code documentation.
Rather than add a mention of printing extensions in the Markdown preview section and short rendering walkthrough, what about if we add the VS Code Printing Free extension as a tile in the Markdown extensions section:

image

There is a space available as a previously called out extension has been removed from the Marketplace.

If this sounds good to you, I wonder if you could update the extension description in a few ways?
For example, "markdown" should always capitalized as "Markdown".
Also it would look better on the tile if "Fully rendered Markdown" came first.

PeterWone:
Yes, that would be fantastic. Thank you for the opportunity.

Updating now to mention Markdown first in the extension description and generally capitalize “Markdown”.

gregvanl:
Added the tile. It uses a live query to the Marketplace so when you update the extension description and republish the extension, it will be reflected on the tile.
I would be curious to hear how this mention affects your install count and user engagement. Maybe you can report back here after a month?

PeterWone:
Fixed capitalisation of Markdown in all my docs and missed it in the extension description. I have bug fix going out next weekend, I'll fix it then.

PR #4634, opened 2021-06-25
Add additional extensionKind info

bamurtaugh:
After recent extension investigations, we thought perhaps there could be documentation improvements to clarify `extensionKind` to users, i.e. the concept itself, differences in extension kinds, when to define it.

Upon reviewing the docs even more closely, I feel as though they are quite well-defined already (but perhaps that's also because I'm rather close to the problem at this point).

One takeaway I thought could make the experience a bit easier to read/grasp would be linking users directly to the [troubleshooting section](https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location) as that took a bit of scrolling to find and may not be immediately obvious if we just link users to the top of the [Supporting Remote Development](https://code.visualstudio.com/api/advanced-topics/remote-extensions) doc. I also tried to add a bit more context around the docs links.

This is an initial proposal - please let me know if you have other ideas in how we could improve this (or if this change even feels necessary). Thank you!


Chuxel (approved):
LGTM!

egamma (approved):

PR #4635, opened 2021-06-28
fix: Make sure `setContext` docs match behavior

Jason3S:
I tried to follow the directions on setting the context for when clauses, but it didn't work as documented.

The `setContext` value must match the expected read value.

It is correct in the release document. See: [vscode-docs/v1_49.md at main · microsoft/vscode-docs](https://github.com/microsoft/vscode-docs/blob/main/release-notes/v1_49.md#in-operator-for-when-clauses)

Two ways to document based upon preferred convention:

## Using `.`

```ts
vscode.commands.executeCommand('setContext', 'ext.supportedFolders', [ 'test', 'foo', 'bar' ]);
```

```json
{
  "when": "explorerResourceIsFolder && resourceFilename in ext.supportedFolders"
}
```

## Using `:`

```ts
vscode.commands.executeCommand('setContext', 'ext:supportedFolders', [ 'test', 'foo', 'bar' ]);
```

```json
{
  "when": "explorerResourceIsFolder && resourceFilename in ext:supportedFolders"
}
```

### DIFF
--- a/api/extension-guides/command.md
+++ b/api/extension-guides/command.md
@@ -214,7 +214,7 @@ Last, menus showing commands, like the Command Palette or context menus, impleme
 
 If you are authoring your own VS Code extension and need to enable/disable commands, menus, or views by using a `when` clause context and none of the existing keys suit your needs, then you can add your own context.
 
-The first example below sets the key `myExtension:showMyCommand` to true, which you can use in enablement of commands or with the `when` property. The second example stores a value that you could use with a `when` clause to check if the number of cool open things is greater than 2.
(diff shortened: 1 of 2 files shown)


SchoofsKelvin (approved):
I also just noticed (while adding when clause contexts to my extension) that the documentation is incorrect regarding this. I added `ext.key` and `Developer: Inspect Context Keys` only lists `ext.key`, even though the documentation makes it seem as if it should be `ext:key`.

Might also be worth mentioning that using `extensionId.someKey` is a good convention to follow, although it seems like the command accepts any kind of key, not requiring a dot/colon and even accepting Unicode characters. _whether the `"when"` fields accept Unicode, not sure_

gregvanl:
Thanks!

PR #4637, opened 2021-06-28
Update Notebook API docs

tanhakabir:


api/extension-guides/notebook.md
```diff
@@ -210,62 +211,68 @@ Rich outputs are the most advanced form of displaying cell outputs. They allow f
...
 npm install -g yo generator-code
 ```
 
-Then, run `yo code --insiders`.
+Then, run `yo code` and choose `New Notebook Renderer (TypeScript)`.
 
 ### A Simple, Non-Interactive Renderer
 
-Renderers are declared for a set of mimetypes by contributing to the `contributes.notebookOutputRenderer` property of an extension's `package.json`. This renderer will work with input in the `ms-vscode.github-issue-notebook/github-issue` format, which we will assume some installed kernel is able to provide:
+Renderers are declared for a set of mimetypes by contributing to the `contributes.notebookRenderer` property of an extension's `package.json`. This renderer will work with input in the `ms-vscode.github-issue-notebook/github-issue` format, which we will assume some installed kernel is able to provide:
```
connor4312, line 265: I think this should be "some installed controller" now

api/extension-guides/notebook.md
```diff
@@ -278,25 +285,26 @@ Renderers are declared for a set of mimetypes by contributing to the `contribute
 
 Output renderers are always rendered in a single `iframe`, separate from the rest of VS Code's UI, to ensure they don't accidentally interfere or cause slowdowns in VS Code. The contribution refers to an "entrypoint" script, which is loaded into the notebook's `iframe` right before any output needs to be rendered. Your entrypoint needs to be a single file, which you can write yourself, or use a bundler like Webpack, Rollup, or Parcel to create.
 
-When it's loaded, your entrypoint script should immediately call `acquireNotebookRendererApi()` with your renderer ID, and start listening to notebook output events. For example, this will put all your GitHub issue data as JSON into the cell output:
+When it's loaded, your entrypoint script should immediately call `ActivationFunction` from `vscode-notebook-renderer` to render your UI once VS Code is ready to activate and render your renderer. For example, this will put all your GitHub issue data as JSON into the cell output:
```
connor4312, line 288: The renderer is no longer calling anything, it's exporting the activation function.

api/extension-guides/notebook.md
```diff
@@ -312,49 +320,46 @@ const GithubIssues: FunctionComponent<{ issues: GithubIssue[]; }> = ({ issues })
...
-  }
 });
 ```
 
-It's important to bear in mind that all outputs for a notebook are rendered in different elements in the same iframe. If you use functions like `document.querySelector`, make sure to scope it to the specific output you're interested in to avoid conflicting with other outputs. In this example, we use `evt.element.querySelector` to avoid that issue.
+It's important to bear in mind that all outputs for a notebook are rendered in different elements in the same iframe. If you use functions like `document.querySelector`, make sure to scope it to the specific output you're interested in to avoid conflicting with other outputs. In this example, we use `element.querySelector` to avoid that issue.
+
+> **Note**: The following parts of Notebook API is still proposed and under development, which means some Notebook functionality, like the following, is only available on VS Code [Insiders](/insiders) and requires adding `vscode.proposed.d.ts` to your extension project. You can learn more in [Using Proposed APIs](/api/advanced-topics/using-proposed-api).
 
 ### Interactive Notebooks
```
connor4312, line 360: The content you have about the messaging API is useful, but it's actually not the right use case here--the original approach in this file remains the currently recommended way to do things.

The renderer messaging API is useful to allow for interactions between a specific renderer and the machine or filesystem. For example, the renderer might have a button that opens a folder in the file explorer. However, it _shouldn't_ be used to communicate directly to the controller like this, since that creates a hard dependency between the kernel and a specific renderer, and prevents other extension authors from being able to create renderers that work with the controller (or even at all, depending on how tight the coupling is). Instead, the controller preloads should be used to provide a API for all renderer extensions to use.

connor4312 (changes requested):

api/extension-guides/notebook.md
```diff
@@ -57,122 +56,128 @@ A content provider is declared in `package.json` under the `contributes.notebook
...
+        execution.replaceOutput([new vscode.NotebookCellOutput([vscode.NotebookCellOutputItem.text('Dummy output text!')])])
+        execution.end(true, Date.now());
     }
-};
+}
 ```
 
+Samples:
+
+* [GitHub Issues Notebook](https://github.com/microsoft/vscode-github-issue-notebooks/blob/main/src/extension/notebookProvider.ts): Controller to execute queries for GitHub Issues
```
jrieken, line 173: better use perma links?

api/extension-guides/notebook.md
```diff
@@ -57,122 +56,128 @@ A content provider is declared in `package.json` under the `contributes.notebook
...
+* [REST Book](https://github.com/tanhakabir/rest-book/blob/main/src/extension/notebookKernel.ts): Controller to run REST queries.
+
```
jrieken, line 175: another, very simple sample is the regexp notebook: https://github.com/jrieken/vscode-regex-notebook/blob/master/src/extension/extension.ts#L56

jrieken (commented):
👍

api/extension-guides/notebook.md
tanhakabir: I think I've reverted back to preload scripts but let me know if anything looks off!

connor4312 (approved):

PR #4641, opened 2021-06-29
fix(when-clause-contexts): clause context variable name typo

avinoamsn:
I believe the `activiewFocus` flag mentioned under _Active/focused view or panel when clause context_ is supposed to be `focusedView`.

### DIFF
--- a/api/references/when-clause-contexts.md
+++ b/api/references/when-clause-contexts.md
@@ -185,7 +185,7 @@ Panel Identifiers:
 * workbench.panel.comments - Comments
 * workbench.view.search - Search when `search.location` is set to `panel`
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @avinoamsn Thanks for catching this! Reviewing the sentence, I think it is talking about view or panel so that the fix would be `or activePanel`.
If that also sound right to you, would you like to update your PR?

gregvanl:
I went ahead and fixed the typo with https://github.com/microsoft/vscode-docs/commit/60c901e4a2d979bc53f01ff9f255a61b4b90a0b5.
Thanks again for noticing!

PR #4645, opened 2021-06-30
Update tutorial to use up-to-date version of mingw-w64 via MSYS2

jureid:
Still need to update screenshots that show terminal output after running the build task.

@sean-mcmanus, when I use the mingw-w64 that I installed with MSYS2, the default build task changes cwd to from "${workspaceFolder}" to "C:/msys64/mingw64/bin". This seems weird to me, but it works. When I changed it back to "${workspaceFolder}" the build failed. Do you know why we're doing this?

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -25,17 +25,13 @@ To successfully complete this tutorial, you must do the following steps:
 
     ![C/C++ extension](images/cpp/cpp-extension.png)
 
-1. Install Mingw-w64 via the SourceForge website. Click [Mingw-w64](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/installer/mingw-w64-install.exe/download) to download the Windows Mingw-w64 installer.
-    1. Run the installer.
-    1. For **Architecture** select **x86_64** and then select **Next**.
-    1. On the Installation Folder page, use the default installation folder. Copy the location as you will need it later.
-    1. Select **Next** to start the installation.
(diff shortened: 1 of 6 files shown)


gregvanl:
@jureid Should we also update /docs/languages/cpp since it also uses the MinGW toolset? I added some duplication between the MinGW tutorial and C++ languages doc since SEO was causing people to land on /docs/languages/cpp.

jureid:
> @jureid The "cwd" needs to be "C:/msys64/mingw64/bin" when that path is not in the user's PATH. Once that is added, it should change to "${fileDirname}", i.e. I think your "cwd" should be changed to "${fileDirname}" since you have a step for updating the PATH.
> 
> Also, I'm not sure why you're seeing the "cwd" be ${workspaceFolder} since that was changed with the 1.4.1 to be ${fileDirname}.

Ah, that makes sense because I created the build task / debug configuration before adding this mingw-w64 to my PATH. Just changed cwd to ${fileDirname} in both tasks.json and launch.json and it works perfectly. I'll update the doc.

I said ${workspaceFolder} before because that's what was in the sample tasks.json in this doc... we should change it to ${fileDirname} in our other tutorials, too.

jureid:
> @jureid Should we also update /docs/languages/cpp since it also uses the MinGW toolset? I added some duplication between the MinGW tutorial and C++ languages doc since SEO was causing people to land on /docs/languages/cpp.

Yes we should! I can add those changes to this PR so that they are updated together. Sound good?

jureid:
Alright, I think all the updates are complete! One thing to consider: I didn't copy over instructions from the MSYS2 website for how to run the MSYS2 installer and use it to install mingw-w64. I just said to follow the instructions on the MSYS2 website. Let me know if you think we should pull those instructions into these docs instead.

PR #4650, opened 2021-07-02
Adding Live Preview to 1.58 Notes

andreamah:
Since this is the first time that Live Preview will be in the release notes, it would be great for somebody to double-check how I wrote this.


gregvanl:
Looks good!

PR #4653, opened 2021-07-04
Fix fish config file location

kidonng:
Fish does not use `~/.fishrc` but `~/.config/fish/config.fish`.


gregvanl:
Hi @Tyriar This looks correct to me. Can you review? Thanks

Tyriar (approved):
Thanks 👍

PR #4655, opened 2021-07-05
fix: custom-editors docs typo

doouding:
It seems that there's misdescription in custom editors section of extension guide. But I'm not sure if it's really a misdescription or misunderstanding.


mjbvz:
Thanks!

PR #4669, opened 2021-07-09
Fix docker --cache-from option name

felipecrs:


gregvanl:
Thanks! This is a confusing. I'll follow up with a mention of the devcontainer.json `cacheFrom` to make it clearer.

PR #4672, opened 2021-07-09
fix invalid syntax in contribution-points.md

boneskull:
removed an extra backtick in the `contributes.taskDefinitions` body


gregvanl:
Thanks!

PR #4673, opened 2021-07-12
Add entry for `cacheFrom`

stuartleeks:
Add entry for `cacheFrom` as per https://github.com/microsoft/vscode-docs/blob/main/remote-release-notes/v1_58.md#devcontainerjson-support-for---cache-from


gregvanl:
Thanks!

PR #4681, opened 2021-07-15
Update reference name to NotebookSerializer

kiri11:
NotebookContentProvider no longer exists, and the link points to NotebookSerializer. 
Should remove the old name to avoid confusion.


gregvanl:
Thanks!

PR #4686, opened 2021-07-17
Add working (proper) way to send ctrl+f to shell to find section

corneliusroemer:
Adds workaround for https://github.com/microsoft/vscode/issues/128914 to the documentation.

What is currently listed in the documentation doesn't align with the way keybindings should be handled in the shell as explained here: https://code.visualstudio.com/docs/editor/integrated-terminal#_keybindings-and-the-shell


Tyriar (approved):
I like it, I removed the other example all together 👍

PR #4690, opened 2021-07-20
Fix broken url related to python isort

msgongora:
This PR fixes broken links from the python documentation

### DIFF
--- a/docs/python/editing.md
+++ b/docs/python/editing.md
@@ -203,9 +203,9 @@ Custom arguments to isort are specified in the `python.sortImports.args` setting
 
 To use a custom isort script, use the `python.sortImports.path` setting to specify the path.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
@msgongora Thanks for spotting this! The updated URLs look good.
@luabud Any idea why aka.ms links where used in this topic? Should the aka.ms links be updated rather than replaced with direct links?

luabud:
@gregvanl I actually don't remember why we used aka.ms links. For transparency sake and for making community contributions easier, I personally prefer to have the direct links instead of the aka.ms ones. Direct links are commonly used throughout the VS Code docs as well, right?

gregvanl:
@luabud Just wanted to double-check before merging this PR. I agree that direct links are more transparent and community friendly.

Issue #4167, opened 2020-12-08
Can we have c++ demo project using VStudio also with clang-cl.exe ?


DBJDBJ:
https://github.com/microsoft/vscode-remote-try-cpp/issues/10#issue-759671095

gregvanl:
Closing as duplicate of https://github.com/microsoft/vscode-docs/issues/4168, which covers adding documentation.

PR #4694, opened 2021-07-22
Add workflows for different browsers in macOS installation instructions

### LINKED ISSUE #4617: missing step in macos installation instructions
The installation instructions should have a step to unzip the .zip file that has been downloaded before "Drag Visual Studio Code.app to the Applications folder, making it available in the macOS Launchpad.". 

Installation#
Download Visual Studio Code for macOS.
Open the browser's download list and locate the downloaded archive.
Select the 'magnifying glass' icon to open the archive in Finder.
Drag Visual Studio Code.app to the Applications folder, making it available in the macOS Launchpad.
Add VS Code to your Dock by right-clicking on the icon to bring up the context menu and choosing Options, Keep in Dock.

meowmeowmeowcat:
- Fixes #4617


gregvanl:
Hi @meowmeowmeowcat Thank you for submitting a PR. 
I'm holding on this PR since I'm not sure that the extra step is needed. I have an open question in #4167 about whether the automatic unzipping is not running for the user.
The current steps in the macOS setup documentation work for me on Big Sur 11.4.

meowmeowmeowcat:
Hi, @gregvanl . Did you open the achieve from broswer? I've tested from two different browsers (Safari and Google Chrome), and the archive will be unzipped automatically when we open it from Safari, while Google Chrome needs this extra steps.

meowmeowmeowcat:
Hi @gregvanl! Maybe we could add `If the archive is not unzipped automatically, double click it in Finder to unzip it`? Just in case if someone disabled auto unzipping.

meowmeowmeowcat:
@gregvanl Sorry for opening this PR. I made a mistake while trying to reproduce #4617. I've tested again and the archive should be unzipped automatically when we open it.

gregvanl:
@meowmeowmeowcat I think you and @sndik16 have pointed out that the instructions depend on the particular browser used for the download. The original instructions matched the flow for Chrome but then was changed to follow the flow for Safari (magnifying glass icon). 
I think it makes sense to either make the instructions generic or call out that different browsers have different workflows.
I'm happy to work with you on this PR.

gregvanl:
Looking at the browsers for that page, Chrome is the most popular followed by Safari then Firefox.

image

So I think it make sense to document the Chrome browser flow (as you did in your PR) and add a mention of how Safari is different.

meowmeowmeowcat:
@gregvanl I think both Google Chrome and Safari can unzip the archive automatically. I made a mistake when I was testing with Google Chrome so I opened this PR.

> The original instructions matched the flow for Chrome but then was changed to follow the flow for Safari (magnifying glass icon).

I wonder what magnifying glass icon is actually, since I can't find it in Safari.

> So I think it make sense to document the Chrome browser flow (as you did in your PR) and add a mention of how Safari is different.

Yep, we may add workflows for Chrome, Safari and Firefox in Step 3 so folks will know what to do.

PR #4695, opened 2021-07-23
Fix non-working URLs

handreistoicescu:
The two markdown URLs are leading to two wrong locations which redirect to the 404 page. The two wrongly formed URLs are the following:
- `https://code.visualstudio.com/docs/editor/docs/editor/tasks.md`
- `https://code.visualstudio.com/docs/editor/docs/editor/debugging.md`

Adding the leading forward-slash should fix this.


gregvanl:
Thanks for noticing and fixing these!

PR #4696, opened 2021-07-25
Fix broken link

meowmeowmeowcat:
- Link of `get-pip.py` has broken in [Linux section of python-tutorial](https://code.visualstudio.com/docs/python/python-tutorial#_linux)
- Change https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py to https://pip.pypa.io/en/stable/installation/#get-pip-py


gregvanl:
Thanks!

PR #4700, opened 2021-07-27
fix #4699: how to sync user global state across machines

### LINKED ISSUE #4699: document using setKeysForSync in remote development
In Gitpod we see many extensions are showing annoying webviews on start of a new machine: https://github.com/gitpod-io/gitpod/issues/4566 I believe it should be the same in Codespaces.

Usual issue that extensions do check previous version from the global state, but such key is not marked as sync.

It would be good to document such pattern in https://code.visualstudio.com/api/advanced-topics/remote-extensions#persisting-extension-data-or-state One could then reach out to extension devs to fix it.

akosyakov:
[Rendered](https://github.com/akosyakov/vscode-docs/blob/akosyakov/document-using-setkeysforsync-4699/api/advanced-topics/remote-extensions.md#sync-user-global-state-between-machines)

### DIFF
--- a/api/advanced-topics/remote-extensions.md
+++ b/api/advanced-topics/remote-extensions.md
@@ -219,6 +219,28 @@ export function activate(context: vscode.ExtensionContext) {
 }
 ```
 
+### Sync user global state between machines
+
+If your extension needs to preserve some user state across different machines
+then provide the state to sync using `vscode.ExtensionContext.globalState.setKeysForSync`.
+
+The common problem is to prevent welcome and update pages. You can use the following pattern:
+
+```TypeScript
+// on activate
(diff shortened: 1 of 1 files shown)


gregvanl:
Thanks for this PR and apologies for the delay in getting it merged.
I've discussed this PR with a VS Code engineer and I'm going to move this example to https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage and add links from remote-extensions and also the settings sync topic.
Thanks again!

gregvanl:
Follow up commit https://github.com/microsoft/vscode-docs/commit/a69374a52f8be39ecdfb0887c9ed630b85f849e0

PR #4707, opened 2021-07-28
fix minor typos on devcontainer reference page

qmacro:
Was reading through this page just now (very helpful, thanks!) and spotted a few typos that I could fix. Hope that's OK.


gregvanl:
PRs always welcome. Thanks!

PR #4708, opened 2021-07-28
Remove uneeded sudo from `dnf check-update`

yaayes:
For RHEL, Fedora, and CentOS based distributions the command `dnf check-update` can be executed without sudo.


gregvanl:
Thanks!

PR #4734, opened 2021-08-05
Update repositories extension

bamurtaugh:
- Updated content in Working with GitHub section
- Outstanding question: should we also update the blog post https://code.visualstudio.com/blogs/2021/06/10/remote-repositories? Or is that considered more of a point-in-time?


gregvanl:
I've updated a few more references to the name Remote Repositories across the docs with https://github.com/microsoft/vscode-docs/commit/0efacda3f7bee1f092e205fbdd9f7678d1f6a4d7.
I think it's fine to leave the blog post as a point in time document since the page views at this point are few (100 per day compared to 2000 per day for the Working with GitHub topic).

PR #4745, opened 2021-08-10
Remove old -S flag from sample `npm run`.

yaymukund:
This is no longer needed past npm v5, and is not documented anywhere. It
may confuse people.

**Note**: I saw the "Do not touch" note at the top of the file and read the contribution guidelines. I'm not sure if editing this `md` is the correct way to contribute, so please let me know if an alternate workflow is preferred. Thanks.

### DIFF
--- a/api/working-with-extensions/bundling-extension.md
+++ b/api/working-with-extensions/bundling-extension.md
@@ -31,10 +31,10 @@ Merge these entries into the `scripts` section in `package.json`:
 
 ```json
 "scripts": {
-    "vscode:prepublish": "npm run -S esbuild-base -- --minify",
+    "vscode:prepublish": "npm run esbuild-base -- --minify",
     "esbuild-base": "esbuild ./src/extension.ts --bundle --outfile=out/main.js --external:vscode --format=cjs --platform=node",
-    "esbuild": "npm run -S esbuild-base -- --sourcemap",
-    "esbuild-watch": "npm run -S esbuild-base -- --sourcemap --watch",
+    "esbuild": "npm run esbuild-base -- --sourcemap",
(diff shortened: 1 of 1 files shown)


lnicola:
`MetaDescription` also seems slightly outdated.

yaymukund:
@lnicola Ah, oops, I misunderstood the warning - I thought it applied to the entire file rather than just the header. Fixed! :sweat_smile:

lnicola:
I don't know, let's see what the reviewers say.

gregvanl:
Sorry, if the comment in the metadata section is frightening. I think it was put there so that contributors would know not to worry about the header and that it is maintained by the writers. Nothing actually bad would happen if that section is changed. :-)

gregvanl:
It looks like the npm -s (silent) switch still works in npm version 7.

NPMsilent

Wouldn't suppressing build output be a good thing to do for these build scripts?

yaymukund:
The `-S` flag refers to `--save`, which is different from `-s/--silent`. See: https://stackoverflow.com/a/58475949

gregvanl:
Oh, I see, the current code blocks use "-S" rather than "-s".
I wonder if that was a mistake and the author meant to use "-s"? Do you know that upper "-S" used to do?

gregvanl:
I'm going to check with the original author of the code blocks to see if that should be "-s". From your links above, "-S" was for "npm install" to modify the package.json.

connor4312:
Yea I think it should just be lowercased

lnicola:
`-s` seems to hide [even error messages](https://stackoverflow.com/questions/34426332/how-to-suppress-output-when-running-npm-scripts#comment92033276_34426599). That's probably not something you want when starting a new project.

PR #4752, opened 2021-08-12
Remove generator-aspnet from Additional tools

sam-EV:
Front npm: https://www.npmjs.com/package/generator-aspnet

This package has been deprecated
Author message:

Use dotnet new instead


gregvanl:
Thanks for catching this!

PR #4753, opened 2021-08-13
Various updates to the Java language overview

waldyrious:
As I read through this page, I noticed a few passages that could be improved. These edits are meant to improve readability, consistency and correctness. Some of these changes may be matters of style, so I'm happy to integrate any editorial feedback.

### DIFF
--- a/docs/languages/java.md
+++ b/docs/languages/java.md
@@ -11,7 +11,7 @@ MetaDescription: Learn about Visual Studio Code editor features (code completion
 
 Support for Java in Visual Studio Code is provided through a wide range of [extensions](/docs/java/extensions.md). Combined with the power of core VS Code, these extensions give you a lightweight and performant code editor that also supports many of the most common Java development techniques.
 
-This article will give you an overview of different capabilities of Visual Studio Code for Java developers. For a quick walkthrough of editing, running, and debugging a Java program with Visual Studio Code, use the button below.
+This article will give you an overview of different capabilities of Visual Studio Code for Java developers. For a quick walkthrough of editing, running, and debugging a Java program with Visual Studio Code, use the **Java Getting Started Tutorial** button below.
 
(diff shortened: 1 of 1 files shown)


gregvanl:
@nickzhums

docs/languages/java.md
```diff
@@ -9,9 +9,9 @@ MetaDescription: Learn about Visual Studio Code editor features (code completion
 ---
 # Java in Visual Studio Code
 
-Support for Java in Visual Studio Code is provided through a wide range of [extensions](/docs/editor/extension-marketplace.md). Combined with the power of core VS Code, these extensions give you a lightweight and performant code editor that also supports many of the most common Java development techniques.
+Support for Java in Visual Studio Code is provided through a wide range of [extensions](/docs/java/extensions.md). Combined with the power of core VS Code, these extensions give you a lightweight and performant code editor that also supports many of the most common Java development techniques.
```
gregvanl, line 12: The link here is to point readers to the general VS Code extensions topic, in case they aren't familiar with how VS Code extensions are installed or work.

nickzhums: Maybe we can change the paragraph to mention both
a) indicate that we have a marketplace that provides many extensions
b) mention that for Java specifically, we have the Jave extension pack

waldyrious: Yeah, I think if we want to mention VS Code extensions in general, the sentence would need to be rewritten. The way it currently is worded doesn't combine well with a general link — in particular the start of the sentence, which explicitly limits its scope to "Support for Java". If you agree with a rewording, @gregvanl, I'd be happy to put up a proposal.

macotouenaca (approved):

docs/languages/java.md
gregvanl: Reading this section again, it does seem to be specific to Java extensions, so I'm fine with the change.

PR #4003, opened 2020-09-14
1. refined supported versions on tutorial doc; 2. updated dependency …

yaojinyang:
…view to project view on build doc

PR #4754, opened 2021-08-13
Small edits to the Java tutorial page

waldyrious:
Related to #4753, this time focused on the [Java tutorial](https://code.visualstudio.com/docs/java/java-tutorial) page.

### DIFF
--- a/docs/java/java-tutorial.md
+++ b/docs/java/java-tutorial.md
@@ -12,7 +12,7 @@ MetaDescription: Java tutorial showing basic Java language support in the Visual
 
 This tutorial shows you how to write and run Hello World program in Java with Visual Studio Code. It also covers a few advanced features, which you can explore by reading other documents in this section.
 
-For an overview of the features available for Java in VS Code, see [Java Language Overview](/docs/languages/java.md)
+For an overview of the features available for Java in VS Code, see [Java Language Overview](/docs/languages/java.md).
 
 If you run into any issues when following this tutorial, you can contact us by clicking the **Report an issue** button below.
 
@@ -32,13 +32,14 @@ To help you set up quickly, you can install the **Coding Pack for Java**, which 
 
 ### Installing extensions
 
-If you are an existing VS Code user, you can also add Java support by installing [Java Extension Pack](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack), which includes these extensions:
+If you are an existing VS Code user, you can also add Java support by installing the [Java Extension Pack](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack), which includes these extensions:
 
 * [Language Support for Java(TM) by Red Hat](https://marketplace.visualstudio.com/items?itemName=redhat.java)
 * [Debugger for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-debug)
 * [Java Test Runner](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-test)
 * [Maven for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-maven)
 * [Project Manager for Java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-dependency)
+* [Visual Studio IntelliCode](https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode)
 
 <a class="tutorial-install-extension-btn" href="vscode:extension/vscjava.vscode-java-pack">Install the Java Extension Pack</a>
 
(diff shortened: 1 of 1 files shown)


docs/java/java-tutorial.md
```diff
@@ -59,21 +60,21 @@ For this tutorial, the only required extensions are:
 
 The supported version for running the VS Code for Java extension and the supported version for your projects are two separate runtimes. To run VS Code for Java, Java SE 11 or above version is required; for projects, VS Code for Java supports projects with version 1.5 or above. For more details, refer to [Configure JDK](/docs/java/java-project.md#configure-jdk).
```
waldyrious, line 61: There doesn't seem to be a "VS Code for Java extension". Was this perhaps an old name for the Java extension pack, or something?

For the record, I've also left a comment to this effect in #4003, which is where this language was first introduced.

gregvanl:
@nickzhums for visibility

docs/java/java-tutorial.md
nickzhums: yes, we should probably make it "Extension pack for Java" for consistency

nickzhums:
> @nickzhums for visibility

Thanks for tagging me! Looks good just left a few minor comments

docs/java/java-tutorial.md
waldyrious: Ok, will change!

docs/java/java-tutorial.md
```diff
@@ -111,19 +112,19 @@ To learn more about editing Java, see [Java Editing](/docs/java/java-editing.md)
...
 
-To run and debug Java code, set a breakpoint, then either press `kb(workbench.action.debug.start)` on your keyboard or use the **Run** > **Start Debugging** menu item. You can also use the **Run|Debug** CodeLens options in the editor. After the code compiles, you can see all your variables and threads in the Run view.
+To run and debug Java code, set a breakpoint, then either press `kb(workbench.action.debug.start)` on your keyboard or use the **Run** > **Start Debugging** menu item. You can also use the **Run|Debug** CodeLens option in the editor. After the code compiles, you can see all your variables and threads in the Run view.
 
 <video autoplay loop muted playsinline controls>
   <source src="/docs/java/java-tutorial/run-debug.mp4" type="video/mp4">
 </video>
 
-The debugger also supports advanced features such as Hot Code replacement and conditional breakpoints.
+The debugger also supports advanced features such as Hot Code Replacement and conditional breakpoints.
```
gregvanl, line 121: Looks like the feature is called "Hot Code Replace". 
@nickzhums Should we be using that term wherever the docs currently say "Hot Code replacement"?

docs/java/java-tutorial.md
waldyrious: Actually, on a closer look, I'm wondering if we could be a little more specific. I mean, the extension pack itself doesn't require Java 11. This seems to be specifically related to the language server, which IIUC is provided by the [Language Support for Java](https://marketplace.visualstudio.com/items?itemName=redhat.java) extension (which is included in the extension pack). Indeed, that page says exactly this:

> (latest Java 11 is the current minimum requirement).

This is consistent with what's written in [java-project.md](https://github.com/microsoft/vscode-docs/blob/main/docs/java/java-project.md):

> Although the Java language server requires JDK version 11 or above to run, this is NOT a requirement to your project's runtime.

It also matches what the configuration wizard says:

Screenshot from 2021-08-13 18-21-59_

(Although clearly the second message is too ambiguous and should be fixed too — I've submitted https://github.com/microsoft/vscode-java-pack/issues/729 to track that.)

So, maybe we should make this a reference to the `redhat.java` extension (in whatever phrasing is appropriate here), rather than to the extension pack?

docs/java/java-tutorial.md
gregvanl: I'll update to be "Hot Code Replace" to match the feature name

docs/java/java-tutorial.md
gregvanl: I'm merging this PR since it has a lot of great edits. If you want to add clarification here, please open another PR.

waldyrious: Good point. This was not part of the PR's diff, and probably warrants discussion in a separate issue.

docs/java/java-tutorial.md
waldyrious: For future reference, that was done in commit 3a2469482075d54750f7856d3e2c2cde12f78e10.

docs/java/java-tutorial.md
waldyrious: Ok, I've opened #4786, and will mark this conversation as resolved, to keep this PR's thread focused on the actual changes made here.

PR #4755, opened 2021-08-13
Fix page title in java-project.md

waldyrious:
Looks like this change might have been missed in #3846 (which did the same change for the `java-build.md` page).


gregvanl:
@nickzhums for visibility

gregvanl:
Thanks!

PR #4761, opened 2021-08-17
Make capitalization of language ID consistent

dnicolson:
This pull request replaces instances of "id" and "Id" with "ID". One instance of "Id" was replaced with "identifier", in another instance "id" was used to match the Yeoman screenshot.


gregvanl:
Thanks!

PR #4765, opened 2021-08-19
doc: fix kb shortcut remote development with SSH

kevinnls:
[Remote Development using SSH](https://code.visualstudio.com/docs/remote/ssh) lists the keyboard shortcut for the Command Palette as F1 repeatedly. It uses a (presumably) older style of listing shortcuts on the site.

This PR changes it to use the proper `kb(COMMAND-ID)` format to automatically resolve and display the keybindings based on the user's client platform.

### DIFF
--- a/docs/remote/ssh.md
+++ b/docs/remote/ssh.md
@@ -73,7 +73,7 @@ To connect to a remote host for the first time, follow these steps:
     ssh user@domain@hostname
     ```
 
-2. In VS Code, select **Remote-SSH: Connect to Host...** from the Command Palette (`kbstyle(F1)` or `kb(workbench.action.showCommands)`) and use the same `user@hostname` as in step 1.
+2. In VS Code, select **Remote-SSH: Connect to Host...** from the Command Palette (`kbstyle(F1)`, `kb(workbench.action.showCommands)`) and use the same `user@hostname` as in step 1.
 
     ![Illustration of user@host input box](images/ssh/ssh-user@box.png)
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @kevinnls Thank you for submitting a PR and your interest in the VS Code documentation.
F1 opens the Command Palette on all platforms and, while you are correct that we usually prefer to use the command ID mapping, in this case the topic author wanted to promote the F1 keybinding rather than Ctrl/Cmd+Shift+P.

kevinnls:
Hi @gregvanl . Understandable. Would you consider having both bindings, with `F1` coming first?

Having mentally mapped the command palette to `ctrl + shift + p` I have found it confusing every time I revist this page. And while ideally people would have access to VS Code while reading the docs, it is not always the case as I personally experienced recently.

Having both bindings listed appears to me as a good approach.

gregvanl:
@kevinnls Sure, that would be fine if you wanted to also include the command ID version after F1. We can have a look and see if it looks too bulky.

kevinnls:
Will i have to make a new PR? 

I have already pushed to the same branch, but I am not familiar with closed PRs

docs/remote/ssh.md
```diff
@@ -73,7 +73,7 @@ To connect to a remote host for the first time, follow these steps:
     ssh user@domain@hostname
     ```
 
-2. In VS Code, select **Remote-SSH: Connect to Host...** from the Command Palette (`kb(workbench.action.showCommands)`) and use the same `user@hostname` as in step 1.
+2. In VS Code, select **Remote-SSH: Connect to Host...** from the Command Palette (`kbstyle(F1)` or `kb(workbench.action.showCommands)`) and use the same `user@hostname` as in step 1.
```
gregvanl, line 76: Rather than using "or", let's use a comma between the two keybindings.

kevinnls:
thanks!

gregvanl:
Looks good. 

image

Thanks!

kevinnls:
been thinking. would `/` work better?

i had initially thought `,` could cause confusion with multi combination bindings

while that doesn't seem to be the case, i feel `/` might convey the "or" better

what do you think @gregvanl ?

gregvanl:
I'm reluctant to use '/' since this might be confused with the first keystroke being either F1 or Ctrl along with Shift and P.
Sometimes in the docs we use "Ctrl/Cmd" to indicate the platform specific keystroke. For example "Ctrl/Cmd+Click" to open a link.

kevinnls:
aha! makes sense. `,` it is 👌✨

PR #4799, opened 2021-09-01
Fix section link in Debugging

TyMick:
A link for "Automatically open a URI when debugging a server program" is currently directed at the section above that, "Remote debugging".


gregvanl:
Thanks!

PR #4807, opened 2021-09-03
Remove duplicate section on default profiles

Livven:


gregvanl:
Thanks! Looks like a bad merge between branches. Thanks for catching and fixing this.

PR #4813, opened 2021-09-04
 Change "environment" to "env" in launch.json

JayDouglass:
I'm on VSCode 1.60 and the current docs suggested this format for launch profile environment variables:

```json
"environment": [
    {"name": "config", "value": "Debug"}
]
```
however

```json
"env": {
   "config": "Debug"
}
```
is the format that worked for me.

There was some discussion here https://github.com/microsoft/vscode-cpptools/issues/116 about this issue, but I can't find where this was updated.

Also this link https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes says `env` is the correct property to use. So it looks like the vscode cpp docs are out of sync for this property.

### DIFF
--- a/docs/cpp/launch-json-reference.md
+++ b/docs/cpp/launch-json-reference.md
@@ -99,7 +99,7 @@ Sets the working directory of the application launched by the debugger.
 
 ### environment
 
-Environment variables to add to the environment for the program. Example: `[ { "name": "config", "value": "Debug" } ]`, not `[ { "config": "Debug" } ]`.
+Environment variables to add to the environment for the program. Example: `[ { "config": "Debug" } ]`.
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @JayDouglass Thank you for submitting a PR. I'm checking with the C++ extension team to see if the extension's launch.json attributes changed.

jureid:
Hi @JayDouglass - the C++ debugger should only support the "environment" variable with [ { "name": "foo", "value": "bar" } ] format. This is evident in MIEngine's JsonLaunchOptions:

https://github.com/microsoft/MIEngine/blob/c9ea7b2766261b5181f9925596a3f67c14be5d26/src/MICore/JsonLaunchOptions.cs#L289-L293

And in the C++ extension's package.json:

https://github.com/microsoft/vscode-cpptools/blob/f26f88b8e182a72704d867ddf898d5e58f22b82c/Extension/package.json#L1587-L1603

Were you able to successfully start a C++ debug session using "env"? Are you using [Microsoft's C++ extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools)?

Thanks!

JayDouglass:
Yes, env what works for me. My apologies, I think it's because I'm using
the CodeLLDB extension
<https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb>

gregvanl:
Closing PR as `env` setting came from another extension.

PR #4814, opened 2021-09-06
Update nodejs-debugging.md

caperso:
remove oral expression of an in-list item


gregvanl:
Hi @caperso, thank you for pointing out the confusing section.
The author's intent was to make the debugging options read like a sentence, even though they are laid out as bullets.
I agree that this can be confusing.

PR #4823, opened 2021-09-10
[codespaces.md] Replace jargon

r3m0t:


gregvanl:
Hi @r3m0t Thank you for submitting a PR and your interest in the VS Code documentation.
We'd like to leave the sentence as it is since "user agent" is the term used in the linked W3C spec and also people may be using the VS Code desktop as the Codespaces client instead of their browser.

r3m0t:
OK, it's just jarring as it's the first use in the article and the rest uses "browser". I didn't realise the VS Code is also considered a user agent. Especially as we don't expect it to follow W3C specs.

PR #4824, opened 2021-09-12
Add a note for `contributes.commands`

Pablion:
When using icons form product icons, `light` and `dark` will disable the icon. Correct syntax is `"icon": "$(book)"`


Pablion:
Hi @gregvanl, thank you for the reply.
It seems my PR has been accidentally modified. I think the official document should be clearly readable and mention all unusual behaviors. 
For me, both changes I made was implicit, and I had to test them to make sure if they would work. I personally think its incomplete.
If you think the propose works, I'll repair the PR.

gregvanl:
Sure, if I reviewed an incomplete PR, feel free to resubmit or update the existing PR and I can take a look.

PR #4827, opened 2021-09-15
Update host requirement

bamurtaugh:
It may be beneficial to mention the space requirement directly on the extension doc.

Would love to hear your thoughts on placement, wording, etc.


Chuxel:
We may also want to call out that 2-cores and 2GB of RAM is recommended to get reasonable performance.  1GB is pretty rough.

bamurtaugh:
@gregvanl The wording here has been discussed, and this PR should be good to go!

PR #4830, opened 2021-09-16
Clarify kind of languages for contributes.language

LinqLover:
Not to be confused with natural/display languages.


gregvanl:
Thanks! Good clarification

PR #4835, opened 2021-09-20
Aeschli/plattform-specific

aeschli:
proposal


isidorn:
Looks good. My only concern is that we lost the information that this is supported from VS Code `1.61.0`. I think this is important because `vsce` will throw an error if the platform specific extension does not specify that version as an `engine`.

isidorn (approved):

PR #4839, opened 2021-09-21
use native mac open -a to open vs code

gcsgithub:
use native mac open -a to open vs code


talks2much:
In which scenarios it can be useful? `code` is more elegant

PR #4840, opened 2021-09-22
Update references to the Create New Terminal command

oadugmore:
Updated references to the **Terminal: Create New Terminal** command (current documentation contains what is likely the former name of this command)


rjmolyneaux:
Thanks!

PR #4841, opened 2021-09-22
update the return types

AkatQuas:
With reference to [here](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostApiCommands.ts#L27) and [here](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostApiCommands.ts#L33), it might be possible a big typo for the original document.

So here is my quick fix.

### DIFF
--- a/api/references/commands.md
+++ b/api/references/commands.md
@@ -26,12 +26,12 @@ let success = await commands.executeCommand('vscode.openFolder', uri);
 
 * _uri_ - Uri of a text document
 * _position_ - A position in a text document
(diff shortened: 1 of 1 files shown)


AkatQuas:
Will this PR be accepted ?

gregvanl:
Hi @AkatQuas Sorry for the delay. As you pointed out, this documentation is auto-generated from sources so needed to check in the VS Code product repo.
Thanks for spotting and fixing this!

PR #4846, opened 2021-09-24
C++ docs: Add redundancy to avoid pitfalls

rzhao271:
Recently, I wanted to get a small C demo going (https://github.com/rzhao271/named-pipe-sample).

I read through the documentation too quickly, assumed that I knew what I was doing, and ended up stuck for a while, especially on the `cl.exe` setup.

This PR adds docs to and expands the troubleshooting sections of https://code.visualstudio.com/docs/cpp/config-mingw and https://code.visualstudio.com/docs/cpp/config-msvc to emphasize some steps that a user could've forgotten to do (or assumed were unimportant).

Edit: Commit https://github.com/microsoft/vscode-docs/pull/4846/commits/50045832a2f30f72703bdadde8b0c9e4e4fc435a says it is unverified, but that is because I don't have code signing set up on Codespaces.

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -43,7 +43,7 @@ g++ --version
 gdb --version
 ```
 
-If you don't see the expected output or `g++` or `gdb` is not a recognized command, make sure your PATH entry matches the Mingw-w64 binary location where the compilers are located.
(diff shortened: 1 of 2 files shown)


jureid (approved):
Thank you @rzhao271 for creating this PR!

PR #4858, opened 2021-10-03
Add Microsoft Build of OpenJDK to list of recommended JDKs

Brettb-gpu:
The Microsoft Build of OpenJDK has been released for some time, I think it would be a great idea to list it under the recommended sources for JDK installs.
https://code.visualstudio.com/docs/java/java-tutorial#_installing-a-java-development-kit-jdk

### DIFF
--- a/docs/java/java-tutorial.md
+++ b/docs/java/java-tutorial.md
@@ -74,6 +74,7 @@ If you prefer, you can configure JDK settings using the VS Code Settings editor 
 
 If you need to install a JDK, we recommend you to consider installing from one of these sources:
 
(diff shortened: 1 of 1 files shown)


nickzhums:
I have no issues with this. @eskibear @jdneo any thoughts?

jdneo:
+1 for @ItalyPaleAle's suggestion, we may need to take the ordering into consideration. Same thing applies to the runtime configuration page in vs code java.

Brettb-gpu:
Would it be better if the link to the Microsoft Build of OpenJDK linked to the [home page](https://www.microsoft.com/openjdk) or the[ downloads page](https://docs.microsoft.com/en-us/java/openjdk/download)?

PR #4920, opened 2021-10-28
Add tip from remote-release on non-manual build

bamurtaugh:
Including @fyyyyy's tip from https://github.com/microsoft/vscode-remote-release/issues/5779 in our section in pre-building.

### DIFF
--- a/docs/remote/devcontainer-cli.md
+++ b/docs/remote/devcontainer-cli.md
@@ -71,9 +71,11 @@ You can optionally specify the path to the folder to open, for example `devconta
 
 The `devcontainer build` command allows you to build the dev container image for a folder. As with the `open` command, `build` accepts a path to the folder to build the image for and defaults to the current working folder in your shell. For example, `devcontainer build` will build the dev container image for the current folder and `devcontainer build ~/source/my-folder` will build the container image for the `~/source/my-folder` folder.
 
-In order for your dev container to build properly, it must be built by VS Code and the Remote - Containers extension rather than a manual user build with Docker.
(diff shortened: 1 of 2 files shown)


docs/remote/devcontainer-cli.md
```diff
@@ -71,6 +71,27 @@ You can optionally specify the path to the folder to open, for example `devconta
 
 The `devcontainer build` command allows you to build the dev container image for a folder. As with the `open` command, `build` accepts a path to the folder to build the image for and defaults to the current working folder in your shell. For example, `devcontainer build` will build the dev container image for the current folder and `devcontainer build ~/source/my-folder` will build the container image for the `~/source/my-folder` folder.
 
+In order for your dev container to build properly, it must be built by VS Code and the Remote - Containers extension rather than a manual user build with Docker.
+
+You can promote pre-building and ensuring VS Code builds the container through something like the following in your Dockerfile:
```
Chuxel, line 76: Maybe add a heading and mention that the CLI and GH Codespaces work too.

```suggestion
### [Optional] Avoiding problems with images built using Docker

In order for your dev container to build properly, it may need to be built by the dev container CLI, the Remote - Containers extension, or GitHub Codespaces rather than a manual user build with Docker. To avoid this problem, you can encourage the use of the dev container CLI instead by adding a required, unset build argument in your Dockerfile that is only set in `devcontainer.json` or a dev container specific Docker Compose file. 

For example, you could add the following to your Dockerfile:
```

bamurtaugh:
Thanks for the review @Chuxel - implemented the change.

@gregvanl I think this is ready for your review & merge.

PR #4929, opened 2021-11-01
Removed Microsoft from supported language servers

rjmolyneaux:
Removed Microsoft from the list of supported language servers, in addition to removing MPLS content [here](https://github.com/microsoft/vscode-docs/commit/a97c8de28220315ce64efbdaae02adb193bb5239).


savannahostrowski:
LGTM!

PR #4930, opened 2021-11-02
fix: broken link to virtual-workspaces page

simoneb:
Missing a leading slash, without which the path is wrong


gregvanl:
Hi @simoneb Thanks for finding and fixing this!

I found another instance by searching for `(api/` and it's in api/extension-guides/virtual-workspaces.md.

"This separation also helps with [Untrusted Workspaces](api/extension-guides/workspace-trust) running in [Restricted Mode](/docs/editor/workspace-trust#restricted-mode). Rich language extensions often require trust while basic language features can run in any setup".

Would you like to fix that as well with your PR?

gregvanl:
Merging as don't want to keep a broken link around.
I'll fix the other link.
Thanks again for noticing!

PR #4931, opened 2021-11-03
Add docs for verified publisher feature

chgill-MSFT:
Adding initial docs for the verified publisher feature explaining the benefits, criteria, and how to become verified.


isidorn:
@chgill-MSFT the wording is smooth and good. However I am not sure if this deserves a separate page in the Extension docs. I would suggest that we simply put this as a new section "Verify a Publisher" after https://code.visualstudio.com/api/working-with-extensions/publishing-extension#create-a-publisher

Also I suggest to make this much more concise. I love how this reads for example in release notes where we are trying to sell features, but docs should be more concise. For example I would cut the "benefits off", I think people will just get it.

The current content we could use for the release notes, so feel free to enrich this section https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_62.md#verified-extension-publishers

I would love to hear @gregvanl thoughts as well.

api/working-with-extensions/publishing-extension.md
```diff
@@ -93,6 +93,29 @@ Test your publisher's personal access token using [`vsce`](#vsce), while at the
...
+Verifying your publisher will:
+* Help users confirm your publisher's authenticity.
+* Add a verified publisher badge to your extensions.
+* Give users quick access your brand's landing page or professional portfolio.
+
+![Verified publisher indicators in VS Code](images/verified-publishers/verified-publishers.png)
+
+Eligible domains meet the following criteria:
+
+* Cannot be a subdomain (i.e. <subdomain>.github.io, <subdomain>.contoso.com, or similar).
```
gregvanl, line 109: We try to avoid Latin abbreviations.
Could this be "such as" or "for example"?

gregvanl:
I agree that we don't need a new short topic. I like the section that you've put together and it should keep extension authors in the main publishing doc, rather than having to jump to a new topic.

api/working-with-extensions/publishing-extension.md
chgill-MSFT: It's probably clear enough if I just remove the "i.e." altogether. What do you think?

gregvanl: Especially as you have "or similar"

isidorn:
This is great, thanks a lot @chgill-MSFT 
Merging in.

@gregvanl can we make sure that this changes are live when we do the release today?
I will also make a change in the release notes to point to these docs.

gregvanl:
@isidorn The changes will flow from vnext to main with the rest of the 1.62 content at release time.

PR #4934, opened 2021-11-03
updated definition for Python projectType

ucheNkadiCode:
updating project type


ucheNkadiCode:
I would like to wait on sign off from @bwateratmsft 

Thank you,

bwateratmsft:
I think we should make it clear that the primary purpose is for them to match the framework for their Python project. That it sets the port and command is an effect, but not why they need to do it.

So I'd say something along the lines of "The type of your Python project, `flask` for Flask projects, `django` for Django projects, `fastapi` for FastAPI, and `general` for others. The project type will be used to set the port and commands used for debugging."

bwateratmsft:
Looks good to me!

ghogen (approved):

PR #4943, opened 2021-11-07
Fix the Sass compile command

rojserbest:


gregvanl:
Hi @rojserbest Thank you for submitting a PR and your interest in the VS Code documentation.
I believe sass supports both command line forms, using a space between the input and output as well as using a `:`.

image

Is using a colon the more popular form? I can see how it would be better if you want to compile multiple files.

gregvanl:
Leaving with the simpler syntax that uses a space

PR #4945, opened 2021-11-08
Update web-extensions to simplify setup for chromium browsers

orta:
Chrome allows any website to access random localhost ports on your computer. Is this a good idea? IMO not really, but it's the reality we're in - I use this to do the same thing when making plugins for the TS playground. This means if you want to quickly test an extension in vscode.dev you have to do less work.

### DIFF
--- a/api/extension-guides/web-extensions.md
+++ b/api/extension-guides/web-extensions.md
@@ -194,7 +194,7 @@ There are currently three ways to test a web extension before publishing it to t
 
 * Use VS Code running on the desktop with the `--extensionDevelopmentKind=web` option to run your web extension in a web extension host running in VS Code.
 * Use the [@vscode/test-web](https://github.com/microsoft/vscode-test-web) node module to open a browser containing VS Code for the Web including your extension, served from a local server.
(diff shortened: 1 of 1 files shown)


api/extension-guides/web-extensions.md
```diff
@@ -315,7 +315,11 @@ npx: installed 78 in 2.196s
 
 ```
 
-Then, in another terminal, run `npx localtunnel -p 5000`:
+If you are using a Chromium based browser, open vscode.dev, run Developer: Install Web Extension... from the Command Palette and give "http://localhost:5000" as the address.
+
+<details><summary>If you are not using a Chromium browser</summary>
```
tanhakabir, line 320: Suggestion: write out `If you are not using Edge, Chrome, nor any other Chromium browser:`

tanhakabir:
Makes sense to me but wondering what @joaomoreno and @aeschli think?

aeschli:
+1 Didn't know this. That that simplifies the steps.

gregvanl:
Thanks!

gregvanl:
Unfortunately, the website rendering doesn't support using a details flyout. I'll need to put in inline.

PR #4949, opened 2021-11-08
Update web extension info

bamurtaugh:
Include info on where extensions are persisted and how users can persist them through Settings Sync.

@gregvanl I know we discussed a new H2, but since we already have an Extensions H3, I thought this content might fit in the existing section. Happy to chat further and potentially move to a new section if we think it'd fit better elsewhere.


gregvanl:
@bamurtaugh Looks good. We can reuse the existing H3 and just keep an eye out if users aren't finding the information under the H2 "Limitations" or if more content comes in for extensions and we think an H2 with subsections is needed.

PR #4953, opened 2021-11-09
Remove all remaining references for python.pythonPath setting

luabud:
I believe I caught all the remaining ones, but if someone could please double check that would be great!

### DIFF
--- a/docs/python/editing.md
+++ b/docs/python/editing.md
@@ -80,7 +80,7 @@ If autocomplete and IntelliSense aren't working for a custom module, check the f
 
 | Cause | Solution |
 | --- | --- |
-| The path to the python interpreter is incorrect | Check the `pythonPath` setting. Restart VS Code if you make a correction. |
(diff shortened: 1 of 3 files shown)


docs/python/editing.md
```diff
@@ -80,7 +80,7 @@ If autocomplete and IntelliSense aren't working for a custom module, check the f
 
 | Cause | Solution |
 | --- | --- |
-| The path to the python interpreter is incorrect | Check the `pythonPath` setting. Restart VS Code if you make a correction. |
+| The path to the python interpreter is incorrect | Make sure you selected a valid interpreter path by running the `Python: Select Interpreter` command (see [Environments](/docs/python/environments.md)). |
```
gregvanl, line 83: For commands that are displayed in the VS Code UI like the Command Palette, use bold to format the text.
"by running the **Python: Select Interpreter** commmand"

luabud: Thank you! Fixed it 😊

brettcannon (approved):

rjmolyneaux:
Thanks @luabud !

PR #4956, opened 2021-11-12
Docker extension overview SEO

### LINKED ISSUE #4933: Change Title of Overview page to Docker Extension for VS Code
Right now our title calls us the Container tools extension for VS Code
searching "VS Code Docker Docs" on Google and Bing puts us at 2nd and 4th respectively. 

I suggest we change our title to Docker Extension for VS Code as is the actual name of our extension. 

https://code.visualstudio.com/docs/containers/overview

@ghogen for visibility.

We should also: 
- Make sure Container is included in the metadata body of the Overview (right now remote containers is beating us). 
- We should update Date of the metadata as well because it says April 14 **2016**

ghogen:
Fixes #4933


gregvanl:
Thanks!

PR #4965, opened 2021-11-15
correct minor punctuation and grammar

MeatSim:
- insert commas separating dependent clauses
- delete extraneous article


gregvanl:
Thanks!

PR #4971, opened 2021-11-20
alternate file exclusions

c0a3bd:
Adding an alternate method to modify workspace settings for file exclusions.


gregvanl:
Good suggestion to use the Settings editor UI. Thanks

PR #4972, opened 2021-11-21
Chage the contribution point from grammar to grammars

zhuzhuman978:
Chage the contribution point from grammar to grammars in the Tokenizations part of yntax-highlight-guide


gregvanl:
Thanks for catching this!

PR #4973, opened 2021-11-21
Add gpg config for agent loopback

dlech:
To get gpg agent forwarding to work on macOS 12 with gpg and pinentry installed from homebrew, I had to add the following config. Otherwise, I would get the error: `Inappropriate ioctl for device`. I haven't tried it on other OSes.

Source: https://d.sb/2016/11/gpg-inappropriate-ioctl-for-device-errors

### DIFF
--- a/docs/remote/containers.md
+++ b/docs/remote/containers.md
@@ -512,6 +512,23 @@ If you do not have GPG set up, you can configure it for your platform:
   * Register a `pinentry` GUI in your WSL distro. `echo pinentry-program /mnt/c/Program\ Files\ \(x86\)/Gpg4win/bin/pinentry.exe > ~/.gnupg/gpg-agent.conf`
   * Reload the `gpg` agent in WSL. `gpg-connect-agent reloadagent /bye`
 
+Then, still on the host system, you need to enable loopback for `pinentry`.
+
+`~/gnupg/gpg.conf` should contain the lines:
+
+    use-agent
+    pinentry-mode loopback
+
+And `~/gnupg/gpg-agent.conf` should contain the line:
+
+    allow-loopback-pinentry
+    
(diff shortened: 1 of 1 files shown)


chrmarti:
@dlech Thanks for your PR!

I'm not entirely sure how the lookback feature works. Could you describe what `pinentry-mode loopback` does? Would that make gpg ask for the passphrase inside the container?

Would installing a pinentry on macOS solve the original problem too? E.g., https://formulae.brew.sh/formula/pinentry-mac?

dlech:
I wasn't able to find any good information explaining why this works. There are just lots of blog posts and stack overflow answers the say this is how you fix the problem.

The [best info I could find](https://www.gnupg.org/documentation/manuals/gpgme/Pinentry-Mode.html) says this of the loopback function:

> Redirect Pinentry queries to the caller. This enables the use of gpgme_set_passphrase_cb because pinentry queries are redirected to gpgme.

dlech:
> Would that make gpg ask for the passphrase inside the container?

No, the passphrase is requested on the host OS.

chrmarti:
The proposed instructions depend on how GPG was installed. We suggest GPG Suite for macOS above and that seems to come with a pinentry installed (reading its release notes, not verified by installing) and should work without additional configuration.

If we add a section for installing using Homebrew, I'd suggest to check if https://formulae.brew.sh/formula/pinentry-mac solves the issue too since that uses the more common 'pinentry' approach.

My suggestion is to check if there is good external documentation for installing GPG using Homebrew and just link to that. Adding too much detail to our documentation about external tools will make it harder to use.

gregvanl:
Closing PR as author hasn't responded back and the engineer's recommendation is to leave the documentation as is since it should cover most configurations.

PR #4998, opened 2021-12-02
Don't download the GPG key instead read the bytes.

Prajwal-Koirala:
- Its a pretty simple change, instead of downloading the GPG key it will read it off the bytes and than save that.


Prajwal-Koirala:
Dear developers,
Any feedback provided on this PR, would be nice.
- prajwal

dlech:
Please read https://www.linuxuprising.com/2021/01/apt-key-is-deprecated-how-to-add.html to understand why we should not use `apt-key`.

We should probably also be using `/usr/local/share/keyrings/` instead of `/etc/apt/trusted.gpg.d/`. Otherwise, if Microsoft's key is ever compromised, it could be used to spoof signing for other apt repositories.

PR #5009, opened 2021-12-06
Add TOC entry that will redirect to CMake docs on github

TylerMSFT:


gregvanl:
Can we change the filename to build-with-cmake.md? 
That will make a better looking URL if people want to share it and for SEO. 
The example bridge-to-kubernetes.md was actually the former topic name and not so named to mean a "bridge" to another topic.

PR #5011, opened 2021-12-07
Update docs for Java Debugging page

testforstephen:
Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>

- Previous page lists too much videos, i will remove those unimportant videos to simplify the page.
- Categorize the videos under **Use** part and adjust the order.
- Re-record those out-of-date videos.


testforstephen:
// cc: @nickzhums 

Here is one question to @gregvanl, Is there any way to display the ToC for all sub headers at the right sidebar?

Current ToC only lists the second-level headers, that seems not enough for the Java debugging page. Because the debugging page contains lots of videos and makes the page too long, it's a challenge for users to discover the following materials without an efficient ToC.

image

testforstephen:
Yes. If this is a limitation on the underlying document processing system to render more headers to the ToC, we can currently add more H2 headers for better navigation.

testforstephen:
Here is the new on-page ToC.

Install
Configure
Run and Debug
Debugging single files
Debug session inputs
Breakpoints
Expression evaluation
Hot Code Replace
Step filtering
Configuration options
Troubleshooting
Feedback and questions
Next steps

nickzhums (approved):

PR #5014, opened 2021-12-08
Fix/update Python invocations from the command-line

skycommand:
Hello

The subject of this pull request asks the reader to run a `python3` command in macOS/Linux, but a `py` command in Windows instead. In my test on Windows 10, Visual Studio Code 1.62.3, and Python 3.10 (Microsoft Store):

- The `python3` command works in Windows too. 
- The `py` command doesn't work.
- The `python` command works in Windows and seems to be the equivalent of the `python3` command.

This PR eliminates all platforms distinctions and uses the `python3` command uniformly throughout the document.

This PR probably needs a thorough review.

### DIFF
--- a/docs/python/python-tutorial.md
+++ b/docs/python/python-tutorial.md
@@ -67,23 +67,13 @@ The built-in Python 3 installation on Linux works well, but to install other Pyt
 
 ## Verify the Python installation
 
-To verify that you've installed Python successfully on your machine, run one of the following commands (depending on your operating system):
-
-- Linux/macOS: open a Terminal Window and type the following command:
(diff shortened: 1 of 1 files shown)


skycommand:
I think I've discovered what's going on.

- The Microsoft Store version of Python 3.10 does not come with Python Launcher for Windows (`py`). It is an exclusive feature of Python.org's installer and is **only available when the installing user has administrative privileges.**
- `Python3.exe` is a feature of Windows 10 called "App execution aliases." It attempts to locate Python 3. Failing that, it opens Python's page on Microsoft Store.
- There are several implementations of Python 3 on Windows that come with `python3.exe`. MSYS64 is one of them.

PR #5018, opened 2021-12-08
Intro videos suggestions

Logan-Cope:
Created a directory tree with the intention of keeping the VS Code Introductory Videos up to date. This is in response to vscode-docs/issues/4854 which states how the current "Getting Started" video is not up to date.

The directory contains 2 sub-directories: 1) Scripts and 2) Video Attempts where contributors can attempt to write new scripts or create new introductory videos.

The idea is that we can reduce the load of recreating these videos by allowing a place to contribute ideas while keeping them separate so that project owners can pick and choose when/how to use these video attempts/scripts.

I have also included my own attempt at creating a new "Getting Started" video under Intro-video-suggestions/Video Attempts/Getting Started.


digitarald:
Thanks @Logan-Cope, this is really cool work! I love the initiative.

I personally dig the idea of community contributed intro videos using shared transcripts. The downside is keeping consistency in language and style, so the end to end watching experience feels connected. The other problem is timing, as we have started to brainstorm on new story scripts for the intro videos that are less welcome page focused.

I realized and appreciate the time and effort that went into this. Another option would have been to create an issue discussing your proposal to gather feedback. We might come back to your idea of community powered videos as we think about localization and exploring more integrated videos within our docs.

Thanks again for this contribution. I'll close this pull request for now.

PR #5035, opened 2021-12-16
Move content from separate Python configuration article to Tips and Tricks

ghogen:


ghogen:
@gregvanl This requires one redirect - from containers/python-configure-containers to containers/troubleshooting#running-as-a-non-root-user

PR #5043, opened 2021-12-20
added language identifier hyperlink for better clarification

GorvGoyl:
It is not clear to some users what `language identifiers` are when setting up `Language snippet scope`. I added a hyperlink so that users can learn more about it.


gregvanl:
Thanks! Good addition

PR #5054, opened 2021-12-25
Change rpm repo id from code to vscode

junjieyuan:
Make repo id same as repo filename.

Signed-off-by: Junjie Yuan <yuan@junjie.pro>


Tyriar:
I'm concerned about unforeseen issues this could cause. Right now the repo id matches the CLI name

PR #5055, opened 2021-12-26
Fix editor.renderIndentGuides deprecation

pallxk:
I guess it's better to use non deprecated settings in docs. So here is the PR.

Ref:

> https://github.com/microsoft/vscode-docs/blob/main/release-notes/v1_61.md#indentation-guides-settings
>
> The `editor.renderIndentGuides` and `editor.highlightActiveIndentGuide` settings have been deprecated in favor of `editor.guides.indentation` and `editor.guides.highlightActiveIndentation`.


gregvanl:
Thanks!

PR #5071, opened 2022-01-03
Reverts #4945 

orta:
re: https://github.com/microsoft/vscode/issues/138623 where CSP errors occurs with using localhost


gregvanl:
Thanks for following up with this!

PR #5072, opened 2022-01-03
Note on browser support

bamurtaugh:
Based on recent discussions, I wanted to add a note on official browser support in VS Code for the Web and any potential limitations.

Added an initial note on which browsers are supported - it's my understanding there may be something to note for webviews in Firefox (@alexr00?), and @jrieken I think you've done notable testing in Safari. Would love to know if there's anything specific we should highlight for these or other browsers (cc @digitarald too).

### DIFF
--- a/docs/editor/vscode-web.md
+++ b/docs/editor/vscode-web.md
@@ -143,7 +143,9 @@ There are also extensions that run in the browser with partial support only. A g
 
 ### Browser support
 
-You can use VS Code for the Web in Chrome, Edge, Firefox, and Safari. Webviews might appear differently or have some unexpected behavior in Firefox and Safari.
+You can use VS Code for the Web in Chrome, Edge, Firefox, and Safari.
+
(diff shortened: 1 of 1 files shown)


alexr00:
Known GHPRI Firefox issue: https://github.com/microsoft/vscode-pull-request-github/issues/3068
and Safari issue: https://github.com/microsoft/vscode-pull-request-github/issues/2426

bamurtaugh:
Thanks @alexr00! Added: `Webviews might appear differently or have some unexpected behavior in Firefox and Safari.` If it should be generalized to the web rather than Firefox and Safari, just let me know.

jrieken:
Safari has come a long way and pain has gotten better, the most annoying issue is Safari reloading the page because of high load/energy consumption. This is the tracking issue https://github.com/microsoft/vscode-dev/issues/285 - tho we have no clue what to do and we aren't only the page receiving the "banner of shame". Also seeing it in outlook and GH issues with many GIFs etc... 

What do you think about about adding links to issue queries? We usually label with [`safari`](https://github.com/microsoft/vscode/labels/safari), `ios`, `upstream` etc... That would always be up to date

bamurtaugh:
Thanks for the insight @jrieken. I agree the link to issue queries is a great idea to get the most up-to-date status. I've added a note on this - let me know what you think!

PR #5081, opened 2022-01-07
Add cuda language id

hokein:
The cuda language is supported in vscode, see https://github.com/microsoft/vscode/pull/119444


gregvanl:
Hi @hokein Thank you for noticing the omission and submitting this PR.
I think "CUDA" is usually capitalized (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html) and we should match the in-product language alias as "CUDA C++".

image

hokein:
Thanks for the comment, done.

PR #5086, opened 2022-01-10
Point to Remote-Containers feature contributions

bamurtaugh:
Fixes https://github.com/microsoft/vscode-remote-release/issues/6128

@chrmarti @Chuxel @gregvanl Open to feedback on wording and location in the doc!


chrmarti (approved):

raghavthind2005 (approved):

Chuxel:
@bamurtaugh I almost wonder if this can just be under frequently asked questions. Having it before the quick start makes it almost look like you have to know this to work with the extension, which is not the case.  Anyone familiar with VS Code will also know this already.

bamurtaugh:
That's a good point @Chuxel. I initially moved it as the last point under `Common questions` on the Remote-Containers doc, but since settings configuration applies to any of the remote extensions, I then moved it to the [actual FAQ page](https://code.visualstudio.com/docs/remote/faq#_general). Please let me know if you have any thoughts.

Chuxel (approved):

Issue #5061, opened 2021-12-26
"jinja" and "code" are not mentioned as launch configuration option for Python debugging


rioj7:
In the section [Set configuration options](https://code.visualstudio.com/docs/python/debugging#_set-configuration-options) on Python debugging there is no mentioning of the `jinja` option.

It is mentioned in the Flask debugging section but Jinja can be used in any Python Web Framework.

Using the setting `"jinja":true` allows me to debug the template (if language set to Jinja) in a non-Flask framework. The template needs to be loaded with the Environment/File Loader.

rioj7:
Also `code` is not mentioned as a possible alternative for `program` or `module`

rjmolyneaux:
Thank you for contributing to the docs @rioj7! We've made the appropriate changes to the page.

PR #5087, opened 2022-01-11
Adding jinja and code to configs

rjmolyneaux:
In response to issue #5061 The following commit includes `jinja` as an option and `code` as an alternative to `module`. @karthiknadig @luabud Can you verify the accuracy please? ☺️

### DIFF
--- a/docs/python/debugging.md
+++ b/docs/python/debugging.md
@@ -355,7 +355,7 @@ You can also rely on a relative path from the workspace root. For example, if th
 
 ### `module`
 
-Provides the ability to specify the name of a module to be debugged, similarly to the `-m` argument when run at the command line. For more information, see [Python.org](https://docs.python.org/3/using/cmdline.html#cmdoption-m)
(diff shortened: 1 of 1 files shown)


docs/python/debugging.md
```diff
@@ -355,7 +355,7 @@ You can also rely on a relative path from the workspace root. For example, if th
 
 ### `module`
 
-Provides the ability to specify the name of a module to be debugged, similarly to the `-m` argument when run at the command line. For more information, see [Python.org](https://docs.python.org/3/using/cmdline.html#cmdoption-m)
+Provides the ability to specify the name of a module to be debugged, similarly to the `-m` argument when run at the command line. The `code` option is also a possible alternative. For more information, see [Python.org](https://docs.python.org/3/using/cmdline.html#cmdoption-m)
```
karthiknadig, line 358:
```suggestion
Provides the ability to specify the name of a module to be debugged, similarly to the `-m` argument when run at the command line. The `code` option is also a possible alternative. For more information, see [Python.org](https://docs.python.org/3/using/cmdline.html#cmdoption-m)

### `code`

Provides the ability to specify text as code to be debugged, similar to the `-c` argument when run at the command line. For more information, see [Python.org](https://docs.python.org/3/using/cmdline.html#cmdoption-c)
```

rjmolyneaux: Thanks for the review @karthiknadig! ☺️

PR #5090, opened 2022-01-12
Make testing.md more pythonic

bedebignell:
Replace 'true/false' with pythonic True/False


rjmolyneaux (approved):
Hi @bedebignell, 
Thank you for bringing this to our attention and contributing to the docs!

Issue #4691, opened 2021-07-20
Update VS Code docs to reflect support for live reloading of flask and django


DonJayamanne:

gregvanl:
Fixed by mentioned PR

PR #5099, opened 2022-01-17
Include steps in tutorial to reflect support for live reloading of flask and django 

luabud:
For https://github.com/microsoft/vscode-docs/issues/4691. 

I figured it was worth removing the note where it said users had to restart the debugger to apply the changes, and adding a step to trigger live reload of the web servers in the Flask and Django tutorials.

### DIFF
--- a/docs/python/tutorial-flask.md
+++ b/docs/python/tutorial-flask.md
@@ -279,7 +279,7 @@ Debugging gives you the opportunity to pause a running program on a particular l
 
     ![Flask tutorial: result of the modified program](images/flask-tutorial/debug-run-result.png)
 
-1.  Change the line in the code to use different datetime format, for example `now.strftime("%a, %d %b, %y at %X")`, and then save the file. The Django server will automatically reload, which means the changes will be applied without the need to restart the debugger. Refresh the page on the browser to see the update.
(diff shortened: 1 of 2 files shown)


docs/python/tutorial-flask.md
```diff
@@ -275,12 +275,12 @@ Debugging gives you the opportunity to pause a running program on a particular l
     'Wed, 31 Oct, 18 at 18:13:39'
     ```
 
-    > **Note**: If you see a change you like, you can copy and paste it into the editor during a debugging session. However, those changes aren't applied until you restart the debugger.
-
 1. Step through a few more lines of code, if you'd like, then select Continue (`kb(workbench.action.debug.continue)`) to let the program run. The browser window shows the result:
 
     ![Flask tutorial: result of the modified program](images/flask-tutorial/debug-run-result.png)
 
+1.  Change the line in the code to use different datetime format, for example `now.strftime("%a, %d %b, %y at %X")`, and then save the file. The Django server will automatically reload, which means the changes will be applied without the need to restart the debugger. Refresh the page on the browser to see the update.
```
karthiknadig, line 282:
```suggestion
1.  Change the line in the code to use different datetime format, for example `now.strftime("%a, %d %b, %y at %X")`, and then save the file. The Flask server will automatically reload, which means the changes will be applied without the need to restart the debugger. Refresh the page on the browser to see the update.
```

karthiknadig (approved):
A minor copy paste error, but rest looks good.

rjmolyneaux:
Thanks @luabud @karthiknadig, merging now. ☺️

PR #5108, opened 2022-01-19
Update name property definition in devcontainer.json reference

bamurtaugh:
From discussion in https://github.com/microsoft/vscode-remote-release/issues/6143, which led to https://github.com/microsoft/vscode-remote-release/issues/6143#issuecomment-1013318036


KoscheyMC (approved):

chrmarti (approved):
Thanks!

PR #5112, opened 2022-01-24
Add missing step for keeping in the Dock

javierruizjimenez:


gregvanl:
Thanks! Good clarification. Apologies for the delay in merging

PR #5115, opened 2022-01-24
Edits in Configure Tests section

rjmolyneaux:
Addressing the doc-bug for PR #5109. @karthiknadig can you please confirm that having the Python extension and an open project are enough to get the beaker to render? ☺️

### DIFF
--- a/docs/python/testing.md
+++ b/docs/python/testing.md
@@ -88,7 +88,7 @@ With this code, you can experience working with tests in VS Code as described in
 
 ## Configure tests
 
-Once you have the Python extension installed and a Python project open within the Editor, a test beaker icon will be available on the VS Code Activity bar. The beaker icon is called the **Test Explorer**. When opening the Test Explorer, you will see a **Configure Tests** button if you don't have a test framework enabled. Once you select **Configure Tests**, you will be prompted to select a test framework and a folder containing the tests. If you're using unittest, you will also be asked to select the file glob pattern used to identify your test files.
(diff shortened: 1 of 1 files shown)


docs/python/testing.md
```diff
@@ -88,7 +88,7 @@ With this code, you can experience working with tests in VS Code as described in
 
 ## Configure tests
 
-Once you install the Python extension, a test beaker icon will be available on the VS Code Activity bar - that's the **Test Explorer**. When opening the Test Explorer, you will see a **Configure Tests** button if you don't have a test framework enabled. Once you select **Configure Tests**, you will be prompted to select a test framework and a folder containing the tests. If you're using unittest, you will also be asked to select the file glob pattern used to identify your test files.
+Once you have the Python extension installed and a Python project open within the Editor, a test beaker icon will be available on the VS Code Activity bar. The beaker icon is called the **Test Explorer**. When opening the Test Explorer, you will see a **Configure Tests** button if you don't have a test framework enabled. Once you select **Configure Tests**, you will be prompted to select a test framework and a folder containing the tests. If you're using unittest, you will also be asked to select the file glob pattern used to identify your test files.
```
karthiknadig, line 91: The test beaker icon can appear for any language, not just python. As long as there is an extension that supports testing is installed and activated. The key point here is that python extension must be activated for users to see the beaker icon and see options for python testing. The python extension in most cases activates if a python file is opened or some project folder where there are known files that trigger activation.

I am not sure if we should provide that clarification, but I just wanted to provide some context on this.

rjmolyneaux: Thanks for the context @karthiknadig! ☺️

PR #5127, opened 2022-01-30
Fix TypeScript website link in typescript.md docs

JoshuaKGoldberg:
The TypeScript language spec is no longer updated.

### DIFF
--- a/docs/languages/typescript.md
+++ b/docs/languages/typescript.md
@@ -11,7 +11,7 @@ MetaSocialImage: images/typescript/Languages_typescript.png
 
 # TypeScript in Visual Studio Code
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @JoshuaKGoldberg Thanks for pointing this out. 
I think we can just remove the sentence calling out the language spec all together and avoid having two links to typescriptlang.org in the same paragraph.
There are also two references in docs/typescript/typescript-compiling.md and docs/typescript/typescript-tutorial.md.
Would you like to update your PR to remove these three sentences?

JoshuaKGoldberg:
That makes a ton of sense @gregvanl, agreed! 71469fd has that.

gregvanl:
Thanks!

PR #5128, opened 2022-01-30
object used for `in` can't map to functions

verhovsky:
This tripped me up, "the value doesn't matter" is making a stronger claim than it should be.

https://github.com/microsoft/vscode/issues/140359


gregvanl:
Thanks for adding the clarification!

PR #5139, opened 2022-02-03
fix small typos in testing.md

lemonez:
Small typo fixes. This article on Python testing was super helpful!!

### DIFF
--- a/docs/python/testing.md
+++ b/docs/python/testing.md
@@ -66,7 +66,7 @@ The exact structure of the code depends on the test framework you're using, and 
 
 The combined results of all the tests is your test report, which tells you whether the function (the unit), is behaving as expected across all test cases. That is, when a unit passes all of its tests, you can be confident that it's functioning properly. (The practice of **test-driven development** is where you actually write the tests first, then write the code to pass increasingly more tests until all of them pass.)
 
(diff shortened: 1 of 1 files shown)


docs/python/testing.md
```diff
@@ -186,7 +186,7 @@ You can run tests using any of the following actions:
...
 
-After a test run, VS Code displays results directly in the editor as gutter decorations. Failed tests will also be highlighted in the editor, with a Peek View that displays the test run error message and a history of all of the tests' runs. You can press `kbstyle(Escape)` to dismiss the view, and you can disable it by opening the User settings (**Preferences: Open Settings (UI)** command in the **Command Palette**) and changing the value of the `Testing: Automatically Open Peek View` setting to `never`.
+After a test run, VS Code displays results directly in the editor as gutter decorations. Failed tests will also be highlighted in the editor, with a Peek View that displays the test run error message and a history of all of the tests' runs. You can press `kbstyle(Escape)` to dismiss the view, and you can disable it by opening the User settings (**Preferences: Open Settings (UI)** command in the **Command Palette**) and changing the value of the `testing.automaticallyOpenPeekView` setting to `never`.
```
gregvanl, line 189: Hi @lemonez The setting is being referenced by its Settings editor text. I think we should leave it as is but since it is a UI string, our convention is to use **bold** rather than `inline code block` formatting.

image

lemonez: Hi @gregvanl , that makes a lot of sense! The formatting is what threw me, and also the fact that some of the other settings are referenced by the JSON key string so it was unclear which was being referenced. So yeah I'd agree, I think **bold** formatting would do the trick; I'll make that change. Thanks for the quick response.

rjmolyneaux:
Hi @lemonez! We're happy to hear that this document has been useful to you. Thank you for helping us improve Microsoft's documentation with your contribution. You can expect to see this update go live soon!

Cheers!

PR #5144, opened 2022-02-06
Easier way to get mingw.

AdhTri001:
I added two more ways to install mingw. Mysys2 provides very old version of gcc(it was like v8.0 for me last time i used it). The path that I added is much more easier and faster. Also it provides much ahead versions than mysys2. It's upto the maintainer to accept or reject the changes but hope this pr show up to searchs of those who can manage to get mingw on windows, if not the docs.

### DIFF
--- a/docs/languages/cpp.md
+++ b/docs/languages/cpp.md
@@ -57,9 +57,15 @@ If you don't have a compiler installed, in the example below, we describe how to
 
 ## Example: Install MinGW-x64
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @AdhTri001 Thanks for creating this PR and your interest in the VS Code documentation.
For getting started content, we don't want to present users with options and we're currently presenting MSYS2 for the installation.
I agree that your PR will be a helpful resource for people looking for other installation options and as you point out, they should be able to find your PR via search.
Thank again!

PR #5162, opened 2022-02-09
fix: 404 links

SimonSiefke:
not sure if this is the best solution, but it is a bit better than having 404 links


gregvanl:
HI @SimonSiefke Thanks for catching this and this looks fine for a quick solution.
I'll work with the VS Code engineers to see how we should handle this going forward

PR #5179, opened 2022-02-13
Fixed typo in contributes.menus

Empiire00:
fix: changed typo in docs (see [here](https://github.com/microsoft/vscode/blob/5c85bdc1998f501306e305e48825747c6a12c8fa/src/vs/workbench/services/actions/common/menusExtensionPoint.ts#L84) )


gregvanl:
Thanks!

PR #5181, opened 2022-02-15
Add working hours for domain verification SLA

prashantvc:


isidorn:
@prashantvc can you please explain the motivation for this PR?
@gregvanl and me feel like within 5 business days is precise enough. It is not like customers are counting hours.

@chgill-MSFT thoughts?

prashantvc:
We have received the complaint lately about working hours of support. So, making it explicit

isidorn:
I suggest to wait for more user feedback. We should not change docs just for one complaint. 
Let's assign to backlog until we gather more user feedback. Thanks!

gregvanl:
Closing for now. We can reopen the PR if there is more user feedback

PR #5192, opened 2022-02-20
Extra `'` typo in `CMD` causes container to break

mdrakiburrahman:
There's an extra '`' in the CMD that causes the container to break when following instructions


gregvanl:
Thanks!

PR #5194, opened 2022-02-20
VS Code no longer adjusts given diagnostic ranges

matangover:
This behavior was removed in the following commit:
https://github.com/microsoft/vscode/commit/7443621e3b4366b7dabca94b0ba988cf8f47e164


matangover:
This might need another change after https://github.com/microsoft/vscode/issues/142810 is fixed

alexdima:
This doc change should no longer be necessary, we will bring back the old behavior.

PR #5197, opened 2022-02-22
Add links to string vs array section

bamurtaugh:
Fixes https://github.com/microsoft/vscode-dev-containers/issues/1315

I added `[Learn more](#formatting-string-vs-array-properties) about string vs array formatting.` to the description of any properties that can be a string or array. 

@Chuxel If you have thoughts on location or wording, please just let me know! In some sections, it felt a bit duplicative (i.e. the lifecycle scripts), but I wanted to ensure users who just focused on one property could still get the information, without expecting them to review other descriptions in the table.


Chuxel:
@bamurtaugh We may also want to add something along the lines of "Note that the array syntax will the command without a shell. See ...." so we make that difference super apparent. That was the primary point of confusion in the vscode-dev-containers issue.

bamurtaugh:
@Chuxel Makes sense. Updated the statement to be: `Note that the array syntax will execute the command without a shell. You can [learn more](#formatting-string-vs-array-properties) about formatting string vs array properties.` Let me know what you think!

Chuxel:
LGTM!  Probably should add this same change over at dev-container-spec as I think about it.

Chuxel (approved):

bamurtaugh:
Thanks @chuxel, and good point! Will add a to-do to dev-container-spec.

PR #5198, opened 2022-02-23
Repository Containers: Clarify lookup

chrmarti:
https://github.com/microsoft/vscode-remote-release/issues/6013

The mentioned log messages will be in the next stable release.


bamurtaugh (approved):
LGTM, thanks for adding this!

This appears to be the only spot in docs where we describe repository configuration folders (not mentioned in our main containers.md and tips & tricks), so just updating here seems fine as it's where users will be getting their info on this setup.

As the user in the issue mentioned reviewing https://github.com/microsoft/vscode-dev-containers/tree/main/repository-containers for info, perhaps we add a link to this section of docs to that readme? Opened https://github.com/microsoft/vscode-dev-containers/pull/1321 to discuss.

PR #5212, opened 2022-03-04
Clarify dev container CLIs

bamurtaugh:
Refer to the current "`devcontainer` CLI" as the "VS Code `devcontainer` CLI", to differentiate from the [reference implementation CLI](https://github.com/microsoft/dev-container-spec/issues/9).

@Chuxel @chrmarti Would love any thoughts or feedback! i.e. on the new naming, mentioning the reference implementation at this point or in this way.


Chuxel (approved):

Abrdjalw:
Duplicate of #

chrmarti (approved):
LGTM! We can think about renaming it from `devcontainer` to `remotecontainer` or similar. Something to look into after we have the reference implementation CLI.

bamurtaugh:
Thank you both for the reviews!

@chrmarti that's a good idea - opened https://github.com/microsoft/vscode-remote-release/issues/6424 to track (let me know if we should track elsewhere, like dev-container-spec repo).

@gregvanl This is ready for your review and merge!

felipecrs:
> LGTM! We can think about renaming it from `devcontainer` to `remotecontainer` or similar. Something to look into after we have the reference implementation CLI.

Or its functionalities could be added to the VSCode's `code` CLI (which of course would require the VSCode itself to expose APIs for making it possible for an extension to contribute features to the CLI).

And then:

```
$ cd my-repo
$ code --remote-container .
```

Considering that `build` would not be required by it, as it would be in the open-source `devcontainer` instead.

PR #5213, opened 2022-03-04
Remove references to insiders program

luabud:
The Python extension now has support for pre-release versions, so the insiders settings are now deprecated.


docs/languages/python.md
```diff
@@ -25,12 +25,6 @@ You can configure the Python extension through settings. Learn more in the [Pyth
 
 >**Windows Subsystem for Linux**: If you are on Windows, WSL is a great way to do Python development. You can run Linux distributions on Windows and Python is often already installed. When coupled with the [Remote - WSL](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl) extension, you get full VS Code editing and debugging support while running in the context of WSL. To learn more, go to [Developing in WSL](/docs/remote/wsl.md) or try the [Working in WSL](/docs/remote/wsl-tutorial.md) tutorial.
 
-## Insiders program
```
karthiknadig, line 28: Should we remove this or point them to the pre-release installation steps here.

luabud: I thought about including it but noticed the other extensions that also support pre-release versions don't have it in their docs? 🤔

In any case since the "install pre-release" is the default option when installing the extension in VS Code insiders, I don't think discoverability will be an issue.

karthiknadig (approved):

PR #5217, opened 2022-03-06
running into alot of system without wget or gpg

Prajwal-Koirala:


gregvanl:
Good suggestion to let people know they may need to install these tools. Thanks!

PR #5233, opened 2022-03-15
Update mac.md

twocs:
The main download link is for an .app file, but the instructions assumed it's a .zip file. No need to extract an .app file.


gregvanl:
Hi @twocs The download is always a .zip but some browsers like Safari will automatically extract the .zip to a .app file.
It doesn't look like Chrome has this behavior or at least not by default.
I think your changes are a good clarification for the cases when the .app after downloading is available.
Thanks!

gregvanl:
I found the "Open "safe" files after downloading" setting in the Safari Preferences > General section.
I'm not sure if checked is the default but some users could have this behavior disabled.

PR #5238, opened 2022-03-16
Use `viewType` instead of `viewId` to match naming in other APIs

jruales:


gregvanl:
Hi @jruales Thank you for your interest in the VS Code documentation and taking the time to create a PR.
This page is automatically generated from the command definitions in VS Code and the current source code uses `viewId` not your suggestion `viewType`.
You can see this code at https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostApiCommands.ts

```ts
new ApiCommand(
		'vscode.openWith', '_workbench.openWith', 'Opens the provided resource with a specific editor.',
		[
			ApiCommandArgument.Uri.with('resource', 'Resource to open'),
			ApiCommandArgument.String.with('viewId', 'Custom editor view id or \'default\' to use VS Code\'s default editor'),
			new ApiCommandArgument<vscode.ViewColumn | typeConverters.TextEditorOpenOptions | undefined, [vscode.ViewColumn?, ITextEditorOptions?] | undefined>('columnOrOptions', 'Either the column in which to open or editor options, see vscode.TextDocumentShowOptions',
				v => v === undefined || typeof v === 'number' || typeof v === 'object',
				v => !v ? v : typeof v === 'number' ? [typeConverters.ViewColumn.from(v), undefined] : [typeConverters.ViewColumn.from(v.viewColumn), typeConverters.TextEditorOpenOptions.from(v)],
			).optional()
		],
		ApiCommandResult.Void
```
If you think the API should be changed, you can open an issue in the vscode product repository.
Since we don't want the commands reference to diverge from the actual source code, I'll close this PR.

jruales:
@gregvanl thanks for the explanation. I think it's important to change it since it's confusing that it's called something else in different places of the documentation. It wasn't clear what "view id" meant. I'll create a GitHub Issue over there. Thanks!

jruales:
Created https://github.com/microsoft/vscode/pull/145270 over there

PR #5244, opened 2022-03-21
api: further clarify super type & sub type relationship

radeksimko:
This just further clarifies the relationship as per discussion in https://github.com/microsoft/vscode/issues/145312

cc @aeschli - thanks for your earlier update in https://github.com/microsoft/vscode-docs/commit/1f45283bd3183837a47f776549ea993dcdafe379

### DIFF
--- a/api/language-extensions/semantic-highlight-guide.md
+++ b/api/language-extensions/semantic-highlight-guide.md
@@ -151,8 +151,9 @@ If necessary, extensions can declare new types and modifiers or create sub types
   }
 }
 
-In the example above, an extension declares a new type `templateType` and a new modifier `native`. By naming `type` as the super type, styling rules for `type` will also apply to `templateType`:
-```
+In the example above, an extension declares a new type `templateType` and a new modifier `native`. By naming `type` as the `superType`, theme styling rules for `type` will also apply to `templateType`:
+
+```json
 {
   "name": "Red Theme",
   "semanticTokenColors": {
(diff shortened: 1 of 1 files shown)


api/language-extensions/semantic-highlight-guide.md
```diff
@@ -151,19 +151,20 @@ If necessary, extensions can declare new types and modifiers or create sub types
...
     "type": "#ff0011"
   }
 }
 ```
-applies to both `type` and all it's subtypes, including `templateType`
 
+`"#ff0011"` above applies to both `type` and all it's subtypes, including `templateType`.
 
-Along with custom token types, extensions can define how these are mapped to TextMate scopes. This is described in the [Custom Mappings](#custom-textmate-scope-mappings) section. Note that custom mapping rules are not automatically inherited from the super type. Instead, subtypes need to refefine the mapping, preferrably to more specific scopes.
+Along with custom token types, extensions can define how these are mapped to TextMate scopes. This is described in the [Custom Mappings](#custom-textmate-scope-mappings) section. Note that custom mapping rules are not automatically inherited from the super type. Instead, subtypes need to redefine the mapping, preferably to more specific scopes.
```
radeksimko, line 167: https://www.grammarly.com/blog/preferably-preferrably/

gregvanl:
I think independently I covered most of this PR with https://github.com/microsoft/vscode-docs/commit/77f415a11eddff2c9ae19539a3cb708eeec78cc5
I'll apply some of the other changes in vnext as well. 
Apologies for not seeing this PR earlier

PR #5245, opened 2022-03-21
api: document missing semantic highlighting contribution points

radeksimko:
Semantic highlighting is already documented in the linked page but it was missing in the reference overview page, so this PR addresses that.


gregvanl:
Thanks!

PR #5248, opened 2022-03-21
Add Settings editor language filter and override notes

rzhao271:
Refs https://github.com/microsoft/vscode/issues/136068#issuecomment-1073445298, and the fact that I also learned more about settings precedence last week from Sandeep.
Additionally, refs https://github.com/microsoft/vscode/issues/58038, and the fact that part of that issue has been fixed, so part of the docs is now unnecessary and incorrect.

We can merge these notes into vnext, because the Settings editor language filter feature and the fix for 58038 are not in Stable yet.

This PR does two things:
- It adds notes on language filters in the Settings editor
- It refines the notes on settings precedence, and also considers remote and language-specific scopes.


gregvanl:
Thanks! 
I had to resolve several merge conflicts due to my updates to settings.md in main. I think I did it correctly but I'll do another review of your changes so that I got everything

PR #5253, opened 2022-03-25
remote: update 'Kill VS Code Server on Host' command

phil-blain:
The current commands fail to show any processes corresponding to the VS Code server. Grep for `vscode-server` instead.
   
   Also add a separate command to clean up the most recent  `.$hash.{pid,log,token}` files.

For bonus points, mark Markdown files with the `diff=markdown` attribute in `.gitattributes`

### DIFF
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -516,8 +516,9 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
 If you want to run these steps manually, or if the command isn't working for you, you can run a script like this:
 
 ```bash
-kill -9 `ps ax | grep "remoteExtensionHostAgent.js" | grep -v grep | awk '{print $1}'`
-kill -9 `ps ax | grep "watcherService" | grep -v grep | awk '{print $1}'`
+# Kill server processes
+kill -9 `ps aux | \grep vscode-server | \grep USER | \grep -v grep | awk '{print $2}'`
+# Delete related files and folder
 rm -rf ~/.vscode-server # Or ~/.vscode-server-insiders
 ```
 
--- a/docs/remote/troubleshooting.md
+++ b/docs/remote/troubleshooting.md
@@ -519,7 +519,7 @@ If you want to run these steps manually, or if the command isn't working for you
 # Kill server processes
 kill -9 `ps aux | \grep vscode-server | \grep USER | \grep -v grep | awk '{print $2}'`
 # Delete related files and folder
-rm -rf ~/.vscode-server # Or ~/.vscode-server-insiders
+rm -rf $HOME/.vscode-server # Or ~/.vscode-server-insiders
 ```
 
 The VS Code Server was previously installed under `~/.vscode-remote` so you can check that location too.
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.mp4 filter=lfs diff=lfs merge=lfs -text
 *.jpg filter=lfs diff=lfs merge=lfs -text
 *.png filter=lfs diff=lfs merge=lfs -text
+*.md diff=markdown


docs/remote/troubleshooting.md
```diff
@@ -516,8 +516,10 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
 If you want to run these steps manually, or if the command isn't working for you, you can run a script like this:
 
 ```bash
-kill -9 `ps ax | grep "remoteExtensionHostAgent.js" | grep -v grep | awk '{print $1}'`
-kill -9 `ps ax | grep "watcherService" | grep -v grep | awk '{print $1}'`
+# Kill server processes and clean up related files
+kill -9 `ps ax | \grep vscode-server | \grep -v grep | awk '{print $1}'`
+rm $(\ls ~/.vscode-server/.$(\ls -1t ~/.vscode-server/bin | head -1).*)
```
tanhakabir, line 521: > Also add a separate command to clean up the most recent .$hash.{pid,log,token} files.

This seems like an unnecessary command. These files are generated on installation of the VS Code server and are overwritten to be correct every time the server is installed or started.

What was the scenario in which removing the most recent of these files were helpful?

docs/remote/troubleshooting.md
```diff
@@ -516,8 +516,10 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
 If you want to run these steps manually, or if the command isn't working for you, you can run a script like this:
 
 ```bash
-kill -9 `ps ax | grep "remoteExtensionHostAgent.js" | grep -v grep | awk '{print $1}'`
-kill -9 `ps ax | grep "watcherService" | grep -v grep | awk '{print $1}'`
+# Kill server processes and clean up related files
+kill -9 `ps ax | \grep vscode-server | \grep -v grep | awk '{print $1}'`
```
tanhakabir, line 520:
```suggestion
kill -9 `ps ax | \grep "out/server-main.js" | \grep -v grep | awk '{print $1}'`
```

To be more consistent with what the Remote SSH extension looks for. This should yield the same result since all other server processes are children of `server-main`.

docs/remote/troubleshooting.md
```diff
@@ -516,8 +516,10 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
 If you want to run these steps manually, or if the command isn't working for you, you can run a script like this:
 
 ```bash
-kill -9 `ps ax | grep "remoteExtensionHostAgent.js" | grep -v grep | awk '{print $1}'`
-kill -9 `ps ax | grep "watcherService" | grep -v grep | awk '{print $1}'`
+# Kill server processes and clean up related files
```
tanhakabir, line 519:
```suggestion
# Kill server processes
```

If you remove the script to clean up most recent files

docs/remote/troubleshooting.md
```diff
@@ -516,8 +516,10 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
...
+kill -9 `ps ax | \grep vscode-server | \grep -v grep | awk '{print $1}'`
+rm $(\ls ~/.vscode-server/.$(\ls -1t ~/.vscode-server/bin | head -1).*)
+# Delete the folder
```
tanhakabir, line 522:
```suggestion
# Delete related files and folder
```

tanhakabir (changes requested):
Thanks for making this PR! I had forgotten these docs were out of date.

docs/remote/troubleshooting.md
phil-blain: This is needed if you want to stop the server, but not delete the whole `~/.vscode-server` folder. For example, let's say you are on a slow connection and want to avoid having to download the server again.

If I only kill the process, I do not delete these files,  and I then try to reconnect, it seems the server looks in these files for its PID and is confused because the processes were killed; so nothing works.

I we remove the whole folder I agree this command is not necessary, but I think it's nice to be able to kill the process and restart the server without downloading it again. Then this separate command is needed.

docs/remote/troubleshooting.md
phil-blain: Using that command does not find any processes for me (the server hash is `899d46d82c4c95423fb7e10e68eba52050e30ba3`, i.e. microsoft/vscode@899d46d82c4c95423fb7e10e68eba52050e30ba3)

I also noticed that on a multi-user system, both my command and yours return PIDs for other user's servers, which we don't want to (and won't be allowed to anyway) kill. So I think a better command would be:

~~~bash
kill -9 `ps aux | \grep vscode-server | \grep -v grep | \grep $USER | awk '{print $2}'`
~~~

phil-blain: I just pushed an updated version with this change.

docs/remote/troubleshooting.md
phil-blain: I just pushed an updated version with that justification in the commit message.

tanhakabir: Hm that seems like a bug if you've successfully killed the VS Code Server process then it will try to find and verify the running process at that PID, if that fails it will start the VS Code Server process again and overwrite the `<hash>.pid` file. If you're seeing something other than this behavior could not create an issue for the behavior you're seeing in https://github.com/microsoft/vscode-remote-release

docs/remote/troubleshooting.md
tanhakabir: You're on a very old version of VS Code `899d46d82c4c95423fb7e10e68eba52050e30ba3` is from 4 months ago. The USER grep is a good addition

phil-blain: Unfortunately I can't upgrade at will on my work computer :) 

Since the docs at https://code.visualstudio.com/docs are not versioned, I think it would be nice if the command provided works not only for the most recent VSCode version, but for previous versions as well. That's why I would keep the `grep vscode-server`. What do you think ?

docs/remote/troubleshooting.md
phil-blain: I've removed the additional command in the latest version, I'll mark this conversation as resolved :)

docs/remote/troubleshooting.md
tanhakabir: Ah I see, okay yeah I agree that `vscode-server` works well here. 

> 
```
kill -9 `ps aux | \grep vscode-server | \grep -v grep | \grep $USER | awk '{print $2}'`
```

Looks good. Could you update your PR with this script?

phil-blain: I updated it this morning: https://github.com/microsoft/vscode-docs/pull/5253#commits-pushed-2660759

tanhakabir: I don't see the grep for USER?

phil-blain: woupsi, I went too fast this morning, you are right :) I just pushed a new version with the added `grep USER`

docs/remote/troubleshooting.md
```diff
@@ -516,8 +516,9 @@ The SSH extension provides a command for cleaning up the VS Code Server from the
 If you want to run these steps manually, or if the command isn't working for you, you can run a script like this:
 
 ```bash
-kill -9 `ps ax | grep "remoteExtensionHostAgent.js" | grep -v grep | awk '{print $1}'`
-kill -9 `ps ax | grep "watcherService" | grep -v grep | awk '{print $1}'`
+# Kill server processes
+kill -9 `ps aux | \grep vscode-server | \grep USER | \grep -v grep | awk '{print $2}'`
+# Delete related files and folder
 rm -rf ~/.vscode-server # Or ~/.vscode-server-insiders
```
tanhakabir, line 522:
```suggestion
rm -rf $HOME/.vscode-server # Or ~/.vscode-server-insiders
```
One last suggestion

tanhakabir (changes requested):

docs/remote/troubleshooting.md
tanhakabir: Looks good to me!

phil-blain:
Thanks for the review ! :)

PR #5258, opened 2022-03-28
Add Rancher Desktop info

bamurtaugh:
Add info about Rancher Desktop to the Docker options doc.

### DIFF
--- a/remote/advancedcontainers/docker-options.md
+++ b/remote/advancedcontainers/docker-options.md
@@ -110,6 +110,14 @@ Podman also has its own implementation of the Compose Spec with [Podman Compose]
 
 ## Other container engines
 
+### Rancher Desktop
+
+[Rancher Desktop](https://docs.rancherdesktop.io/) provides container management and Kubernetes on Windows, macOS, and Linux.
+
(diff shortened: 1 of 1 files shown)


Chuxel:
I think we'd probably say Rancher Desktop is another way to install docker, so it can likely move up into the sections on that.  While it does support nerdctl as well, their guide routes people to the docker CLI.  Otherwise looks good!

Chuxel (approved):

bamurtaugh:
Thanks @Chuxel, makes sense! 

I moved it to the intro section overviewing alternate ways to install Docker.

@gregvanl This is ready for your review and merge.

remote/advancedcontainers/docker-options.md
```diff
@@ -16,6 +16,8 @@ You can use Docker with the [Remote - Containers](https://marketplace.visualstud
 * Docker installed on another machine or remote environment.
   * You only need Docker installed on the remote host, rather than Docker installed locally.
 * Other Docker compliant CLIs, installed locally or in a remote environment.
+  * For instance, [Rancher Desktop](https://docs.rancherdesktop.io/) is another way to install Docker, providing container management and Kubernetes on Windows, macOS, and Linux.
```
gregvanl, line 19: Should we add a link to https://mobyproject.org/ for the Moby reference?

bamurtaugh: Sounds good to me! I added a link.

PR #5265, opened 2022-03-30
Fix backtick excape

todd-xander:
Fix backtick escape in doc here: https://code.visualstudio.com/api/references/vscode-api#3363

### DIFF
--- a/api/references/vscode-api.md
+++ b/api/references/vscode-api.md
@@ -646,7 +646,7 @@ will appear in the uri that is passed to the <a href="#UriHandler">UriHandler</a
   }
 });
 
-const callableUri = await vscode.env.asExternalUri(vscode.Uri.parse(`${vscode.env.uriScheme}://my.extension/did-authenticate`));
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @todd-xander Thanks for noticing this rendering issue.
The API reference topic is auto generated from https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts so the fix will need to be made in the vscode product repo.
I'll do a little investigation and see if we've run into this rendering issue before and how we fixed it.

gregvanl:
Hi @todd-xander Thanks for noticing this rendering issue.
The API reference topic is auto generated from https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts so the fix will need to be made in the vscode product repo.
I'll do a little investigation and see if we've run into this rendering issue before and how we fixed it.

gregvanl:
Fixed with https://github.com/microsoft/vscode/commit/4ebdb61c694953688d172c5600e11310643d1e41 to avoid the backticks.

Here's what it looks like now on our staging server:

image

This will go live with our next 1.67 release at the beginning of May.

Thanks again for catching this!

PR #5270, opened 2022-04-01
Updated docs to use deploy stage instead of after_script

Aerex:
**Context**: The use of the `after_script` for deployment may not be the best usecase for deployment. Since [2012](https://blog.travis-ci.com/after_script_behavior_changes), the `after_script` will **always** run regardless if the result of the test result.  Using the `after_script` will lead to attempts at publishing a new version of an extension in every build. 

The condition `env(TRAVIS_TAG) =~ ^v` has no effect so on each build it will still try to publish a new version.

I think a better solution is to use the `deploy` script which is design for deploy artifacts to a repository, such as the Azure Marketplace.

### DIFF
--- a/api/working-with-extensions/continuous-integration.md
+++ b/api/working-with-extensions/continuous-integration.md
@@ -277,24 +277,23 @@ cache: yarn
 }
 ```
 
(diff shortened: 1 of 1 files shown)


connor4312 (approved):

gregvanl:
Thanks!

PR #5280, opened 2022-04-05
Updated the Django tutorial

jsma:
* Many updates are simply adding new screen shots, updating menu item references, etc. that have changed in VS Code since the tutorial was written in 2018
* Added notes to clarify that VS Code does not currently support proper syntax highlighting for Django HTML templates (VS Code was breaking some of the template tags when copying/pasting from the examples) and does not support auto-discovering (let alone running/debugging) Django tests.
* Updated outdated screen shots, ran ImageOptim on them to reduce file sizes
* Removed duplicate instructions in code comments since these are likely to be copied and pasted, and the preceding content already explains what needs to be done
* Fixed issue with using timezone naive datetime
* Updated product name from SQLite Browser => DB Browser for SQLite
* Simplified configuration of class-based view by setting properties on the view itself instead of importing the model and making a separate variable in urls.py
* Updated custom snippet to make it easier for users to tab from the title block to the content block when filling in the example content


jsma:
I used the dev container while working on these edits but I had issues with VS Code changing the ordered lists. E.g. they are all 1., 1., 1. but VS Code would randomly change some sections to 1., 2., 3., etc. Any tips on how to prevent that? It was a bit tedious to go back through and correct especially if I had other changes on the same line (in other words, I couldn't simply discard the changes on a line by line basis).

jsma:
I spent hours painstakingly following the tutorial (despite claims [no one reads tutorials](https://code.visualstudio.com/blogs/2022/03/08/the-tutorial-problem) 😉), taking notes on what didn't work and turning all of that into this PR. But at the risk of throwing all of that out the window, and without meaning to criticize anyone that has contributed to this guide in the past, it strikes me that this tutorial should be rewritten from the ground up. 

It can't seem to make up its mind whether it's trying to teach a complete beginner Django itself, or if it's trying to teach someone who is somewhat-to-very familiar with Django how they can use VS Code to increase their productivity as they develop, test, and deploy their project. It often leans too heavily on the former, trying to explain some of the most basic Python and Django concepts (how Django template tags work or that single element Python tuples require a trailing comma, etc.).

In other words, it should assume they are capable of reading the [Django tutorial](https://docs.djangoproject.com/en/4.0/intro/tutorial01/) or are maybe even a seasoned Django developer but are now curious how VS Code might fit into their toolbox and make their programming lives better. 

Obviously without proper built-in Django HTML and Django test support, VS Code is a little bit of a hard sell, not to mention other bugs, oddities, and gaps in the documentation that make it a chore to get a functioning environment going (I documented a "newbie" journey of trying to use VS Code with the most basic Wagtail project possible [here](https://github.com/jsma/wagtail-vs-code-demo)). I noted these template/testing limitations front-and-center in this PR to hopefully save others the same frustration and lost hours, so I do think this PR provides some value in the short term.

At any rate, assuming those issues are eventually addressed _someday_, the _ideal_ guide should likely start with using dev containers for development rather than continue to encourage bad habits of polluting host systems and all the related problems. The learner would instead follow a pathway of modern best-practices to create or modify a real-world app with all the local setup/install pain removed. It would be much easier to focus on teaching how to gradually add/enhance Docker Compose configuration files, devcontainer.json/launch.json, and adjusting Django settings to get a powerful setup that can realistically be deployed (here's how to get a real database, a real Elastic Search, etc. locally). The promise of an "it just works!" setup, but for really real this time 😉

docs/python/tutorial-django.md
```diff
@@ -443,17 +450,20 @@ In production, you also need to set `DEBUG=False` in `settings.py`, which necess
...
-    {% load static %}
     <link rel="stylesheet" type="text/css" href="{% static 'hello/site.css' %}" />
     ```
 
 1. Also in `templates/hello/hello_there.html`, replace the contents `<body>` element with the following markup that uses the `message` style instead of a `<strong>` tag:
 
     ```html
-    <span class="message">Hello, there \{{ name }}!</span> It's \{{ date | date:'l, d F, Y' }} at \{{ date | time:'H:i:s' }}.
+    <span class="message">Hello, there \{{ name }}!</span> It's \{{ date|date:'l, d F, Y' }} at \{{ date|time:'H:i:s' }}.
     ```
```
jsma, line 467: Not only are the spaces between variables, "|", and the following filters non-standard practice, VS Code insisted on formatting this by placing the closing `}}` on a new line after I pasted it into the template, which of course breaks the tag. I could only get this template snippet to work by removing the spaces and then VS Code left it alone, all on one line. I documented this behavior above, but it's unclear what the guidance should be for new users. "If VS Code keeps breaking your code on save, turn off all formatting or try another IDE that doesn't do that"?

rjmolyneaux:
Hi @jsma! Thanks for taking the time to review and contribute to VS Code docs. You've left some valuable feedback that will be taken into consideration. We'll keep this PR open and update with any comments if needed. Otherwise, we'll ping you as soon as it's merged! ☺️

Patusia98 (approved):

jsma:
Out of curiosity, how do spam bots keep becoming reviewers?

rjmolyneaux:
Hi @jsma, thank you for taking the time to contribute to the docs! After reviewing this PR with the product team, we agree there's opportunity to improvement here. As a result, we will be taking your feedback to the drawing board and adding some additional edits to provide you with a more pleasant doc experience! Closing this for the time being. Thank you for the insight.

PR #5286, opened 2022-04-14
Replace "replaced list a list" with "replaced with a list".

ErrorGamer2000:
A small mistake, but "replaced list a list" is (I think) not what we are going for here.

## Changes made:
Replaced "list a list" with "with a list" in `docs/nodejs/nodejs-debugging`.


gregvanl:
Thanks for noticing and fixing this!

PR #5295, opened 2022-04-24
Java: Update description and demos for language status item

CsCherrYY:
Since vscode-java has [adopted](https://github.com/redhat-developer/vscode-java/pull/2363) the language status item, we'd better update our tutorial here.

cc: @nickzhums 

Signed-off-by: CsCherrYY <chenshi@microsoft.com>

### DIFF
--- a/api/references/icons-in-labels.md
+++ b/api/references/icons-in-labels.md
@@ -330,6 +330,7 @@ VS Code extensions can use these icons in labels, views, and trees.
 |<i class="codicon codicon-bold"></i>|bold|
 |<i class="codicon codicon-book"></i>|book|
 |<i class="codicon codicon-bookmark"></i>|bookmark|
+|<i class="codicon codicon-bracket-dot"></i>|bracket-dot|
 |<i class="codicon codicon-briefcase"></i>|briefcase|
 |<i class="codicon codicon-broadcast"></i>|broadcast|
 |<i class="codicon codicon-browser"></i>|browser|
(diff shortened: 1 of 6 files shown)


api/references/icons-in-labels.md
```diff
@@ -330,6 +330,7 @@ VS Code extensions can use these icons in labels, views, and trees.
 |<i class="codicon codicon-bold"></i>|bold|
 |<i class="codicon codicon-book"></i>|book|
 |<i class="codicon codicon-bookmark"></i>|bookmark|
+|<i class="codicon codicon-bracket-dot"></i>|bracket-dot|
```
CsCherrYY, line 333: this codicon can be found here: https://github.com/microsoft/vscode/issues/134102#issuecomment-932422528, it's already on vscode codicon repository.

nickzhums:
LGTM - Should we mention the three features it contains? eg. build status, pom file, etc

nickzhums (approved):

CsCherrYY:
> LGTM - Should we mention the three features it contains? eg. build status, pom file, etc

IMO it's more important to let users know we have supported the language status item in Java than what is included in this item, since it's simple enough to find the features and the language status item is the common place for other languages as well. 

If you think we should put a standalone section here. I can also add it🙂

nickzhums:
> > LGTM - Should we mention the three features it contains? eg. build status, pom file, etc
> 
> IMO it's more important to let users know we have supported the language status item in Java than what is included in this item, since it's simple enough to find the features and the language status item is the common place for other languages as well.
> 
> If you think we should put a standalone section here. I can also add it🙂

makes sense, let us keep it this way for now

PR #5301, opened 2022-04-28
socat no longer needed

chrmarti:
(Found in https://github.com/microsoft/vscode-remote-release/issues/6620.)

The forwarding for SSH and GPG now works without socat.


bamurtaugh (approved):
Thanks!

PR #5303, opened 2022-04-29
point to lazy loading NVM article for those with NVM issue

jasonwilliams:
After having an issue with the terminal failing to load I found a useful blog post which talks about the same thing and how to resolve it via lazy loading. I think it's worth adding to the FAQ's so others can be assisted from this too.

Fixes: https://github.com/microsoft/vscode/issues/148434

### DIFF
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -239,7 +239,7 @@ The process outlined below may help you identify which parts of your shell initi
 * Save and fully restart VS Code.
 * Continue commenting out operations until the error disappears.
 
->**Note**: While `nvm` is a powerful and useful Node.js package manager, it can cause slow shell startup times, if being run during shell initialization. You might consider package manager alternatives such as [asdf](https://asdf-vm.com) or search on the internet for `nvm` performance suggestions.
+>**Note**: While `nvm` is a powerful and useful Node.js package manager, it can cause slow shell startup times, if being run during shell initialization. You might consider [lazy loading](https://spin.atomicobject.com/2022/01/18/terminal-lazy-loading/) nvm or other package manager alternatives such as [asdf](https://asdf-vm.com) or search on the internet for `nvm` performance suggestions.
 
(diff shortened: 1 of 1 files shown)


Tyriar:
Would prefer not to recommend/endorse a plugin that we don't know anything about and doesn't seem to be often updated. There's already a suggestion to avoid using nvm which I think is sufficient.

jasonwilliams:
Hey @ljharb just CCing you to let you know nvm doesn't seem to play nice with VSCode. The only option seems to be using the lazy load plugin, but that doesn't seem to be an accepted solution here. I'm not sure if you're (or anyone from nvm) are aware or plan to put in place any performance/lazy loading mechanisms in future. 

Original issue: https://github.com/microsoft/vscode/issues/148434
Issue also referenced here: https://code.visualstudio.com/docs/supporting/faq#_resolving-shell-environment-fails

ljharb:
@Tyriar why would nvm be avoided with vscode?? I use them together exclusively and they work perfectly fine.

@jasonwilliams thanks for the heads up. I’m not sure when that advice was written, but about a year ago i shipped a large performance update, and the problems should be largely eliminated.

If there’s still a problem, filing an issue would be much more appropriate than recommending users avoid the most used node package manager (and than recommending a much more obscure one).

jasonwilliams:
> Would prefer not to recommend/endorse a plugin that we don't know anything about and doesn't seem to be often updated. There's already a suggestion to avoid using nvm which I think is sufficient.

It’s inconsistent with the WSL docs which say to install nvm https://docs.microsoft.com/en-us/windows/dev-environment/javascript/nodejs-on-wsl you may want to update those also if that’s the message you want to send.

ljharb:
fwiw the WSL docs here are flawless - kudos to that team for being one of the few instances of "how to install node with nvm" instructions i've ever seen on the internet that's actually _accurate_!

PR #5320, opened 2022-05-05
Update .NET Core -> .NET

jamesmontemagno:


gregvanl:
Hi @jamesmontemagno Thanks for the PR!
Should we also update https://code.visualstudio.com/docs/languages/csharp?
Looking at https://dotnet.microsoft.com/en-us/download, it appears that we still say "ASP.NET Core", is that correct?

jamesmontemagno:
@gregvanl let me do it in this PR..

It is still "ASP.NET Core".... but ".NET Core" is now just ".NET" :)

jamesmontemagno:
Updated @gregvanl

gregvanl:
Thanks!

PR #5328, opened 2022-05-09
Update activity bar image

rzhao271:
I retook the activity bar hidden image so that it's on Stable rather than on Insiders.


gregvanl (approved):
Thanks!

PR #5329, opened 2022-05-09
Rebrand .NET Core > .NET

ghogen:
.NET 5 and later drop the "Core" and just use the brand ".NET"
ASP.NET Core still uses the "Core" name 
Product UI still uses .NET Core in places, so those refs are not updated
@ucheNkadiCode @gregvanl

### DIFF
--- a/docs/containers/debug-common.md
+++ b/docs/containers/debug-common.md
@@ -13,7 +13,7 @@ With version 0.9.0 and later, the Docker extension provides more support for deb
 
 The Docker extension provides a `docker` debug configuration provider that manages how VS Code will launch an application and/or attach a debugger to the application in a running Docker container. This provider is configured via entries within `launch.json`, with configuration being specific to each application platform supported by the provider.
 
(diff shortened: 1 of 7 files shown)


docs/containers/debug-common.md
```diff
@@ -13,7 +13,7 @@ With version 0.9.0 and later, the Docker extension provides more support for deb
 
 The Docker extension provides a `docker` debug configuration provider that manages how VS Code will launch an application and/or attach a debugger to the application in a running Docker container. This provider is configured via entries within `launch.json`, with configuration being specific to each application platform supported by the provider.
 
-The Docker extension currently supports debugging [Node.js](#node-js), [Python](#python), and [.NET Core](#net-core) applications within Docker containers.
+The Docker extension currently supports debugging [Node.js](#node-js), [Python](#python), and [.NET](#net-core) applications within Docker containers.
```
gregvanl, line 16: I believe you'll need to update this header link since the header is changed below

ghogen:
@ucheNkadiCode I think it is necessary to align with the .NET team's branding.

ucheNkadiCode:
Okay, let's go for it! Sign-off

PR #5342, opened 2022-05-15
Rename Azure Spring Cloud to Azure Spring Apps

silenceJialuo:


silenceJialuo:
Hi @gregvanl, Thanks for your reply in time.

Actually, I think this PR could be published now. Since from the [ plugin page in the marketplace](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-azurespringcloud&ssr=false#overview), we have updated these information.  

Besides, for the official website [https://azure.microsoft.com/services/spring-cloud](https://azure.microsoft.com/services/spring-cloud), they will update the service name in the latter of this month  and we could publish this PR before them.

gregvanl:
Hi @silenceJialuo Merging this PR so that the extension rename is documented and assuming the azure.com updates time this month.
As a follow up, should the screenshots at the end of the article be retaken? They still have the "AZURE: SPRING CLOUD" title in the Azure view.
If someone is going to update the images, they should be sure to enable Git LFS in their local repo per https://github.com/microsoft/vscode-docs#cloning

PR #5364, opened 2022-05-25
update shell integration docs

### LINKED ISSUE #5319: Update shell integration docs

meganrogge:
fix #5319


docs/editor/integrated-terminal.md
```diff
@@ -228,11 +228,11 @@ Extensions make use of links in the terminal, such as GitLens, to identify branc
 
 ## Shell integration
 
-Shell integration is an experimental feature, which will turn on certain features like enhanced command tracking and current working directory detection.
+Shell integration turns on features like enhanced command tracking and current working directory detection.
 
 ![Command decorations show up on the left of the command as well as in the scroll bar when shell integration is enabled](images/integrated-terminal/shell-integration.png)
 
-Shell integration works by injecting a script that is run when the shell is initialized and lets the terminal gain additional insights into what is happening within the terminal. Note that the script injection may not work if you have custom arguments defined in the terminal profile.
+Shell integration works by injecting a script that is run when the shell is initialized and lets us gain and leverage insight into what is happening within the terminal. Note that the script injection may not work if you have custom arguments defined in the terminal profile.
```
Tyriar, line 235: I'd avoid using "us" here, seems better to avoid the leverage part as well since we're repeating what's above the screenshot?

```suggestion
Shell integration works by injecting a script that is run when the shell is initialized and lets VS Code gain insight into what is happening within the terminal. Note that the script injection may not work if you have custom arguments defined in the terminal profile.
```

PR #5365, opened 2022-05-26
Update tips-and-tricks.md - fix links

ilyachalov:
Fix three links. The value of one of the parameters, 'sortBy', has been changed in the links.

Instead of the 'Downloads' value, it should be 'Installs', because there is no 'Downloads' value in the VS Code extension Marketplace:

sortby


gregvanl:
@ilyachalov Thanks for catching this!

PR #5369, opened 2022-05-31
Clarified automatic variables loading from .env

GarrisonD:
https://github.com/microsoft/vscode-remote-release/issues/3150#issuecomment-644010315

### DIFF
--- a/remote/advancedcontainers/environment-variables.md
+++ b/remote/advancedcontainers/environment-variables.md
@@ -67,7 +67,7 @@ If you've already built the container and connected to it, run **Remote-Containe
 
 ## Option 2: Use an env file
 
-If you have a large number of environment variables that you need to set, you can use a `.env` file instead. VS Code will automatically pick up a file called `.env` in your workspace root, but you can also create one in another location.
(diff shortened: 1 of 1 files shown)


remote/advancedcontainers/environment-variables.md
```diff
@@ -67,7 +67,7 @@ If you've already built the container and connected to it, run **Remote-Containe
 
 ## Option 2: Use an env file
 
-If you have a large number of environment variables that you need to set, you can use a `.env` file instead. VS Code will automatically pick up a file called `.env` in your workspace root, but you can also create one in another location.
+If you have a large number of environment variables that you need to set, you can use a `.env` file instead. `docker compose` will automatically pick up a file called `.env` in your workspace root, but you can also create one in another location.
```
chrmarti, line 70: It picks it up next to the (first) docker-compose.yml (not the workspace root, this changed in docker-compose some time ago).

@gregvanl Maybe we should also move the sentence to the Docker Compose-specific text below?

GarrisonD: For me it makes sense to move to the Docker Compose section. This will save from getting into this trap. When I read it I thought that VSCode would load environment variables from .env for me (that would be cool, by the way).

PR #5375, opened 2022-06-01
Update tips-and-tricks.md - fix keybinding

ilyachalov:
The section title says about the folder, so keybinding should be `kb(workbench.action.closeFolder)`, not `kb(workbench.action.closeActiveEditor)`.


gregvanl:
Thanks!

Issue #5378, opened 2022-06-02
Misleading section 'Navigation history' in article 'Visual Studio Code Tips and Tricks'


ilyachalov:
Section '[Navigation history](https://code.visualstudio.com/docs/getstarted/tips-and-tricks#_navigation-history)' in article '[Visual Studio Code Tips and Tricks](https://code.visualstudio.com/docs/getstarted/tips-and-tricks)' says
> Navigate entire history: `kb(workbench.action.quickOpenPreviousRecentlyUsedEditorInGroup)`
> Navigate back: `kb(workbench.action.navigateBack)`
> Navigate forward: `kb(workbench.action.navigateForward)`

<br>**Issue 1.** These are two _different_ ways to navigate around the code. So you should write
> Navigate entire history: `kb(workbench.action.quickOpenPreviousRecentlyUsedEditorInGroup)`
> 
> Alternatively, you can use 
> Navigate back: `kb(workbench.action.navigateBack)`
> Navigate forward: `kb(workbench.action.navigateForward)`

<br>**Issue 2.** The _essence_ of these ways of navigating the code is not described (or described incorrectly). You can write like this
> Navigate open files in current editor group: `kb(workbench.action.quickOpenPreviousRecentlyUsedEditorInGroup)`
> 
> Alternatively, you can use (to navigate between edit locations in one or different files)
> Navigate back: `kb(workbench.action.navigateBack)`
> Navigate forward: `kb(workbench.action.navigateForward)`

<br>**Issue 3.** The _unusual_ use of keyboard shortcut in the first method of code navigation is not described. You can write like this
> Navigate open files in current editor group: `kb(workbench.action.quickOpenPreviousRecentlyUsedEditorInGroup)`
> | `kbstyle(Ctrl)` | `kbstyle(Tab)` | Result of Command |
> | --- | --- | --- |
> | Down and hold... | Down and Up | Open the file list |
> | ...hold... | Down and Up | Moving through the file list |
> | ...up | - | Selecting a position in list and closing the file list |
> 
> Alternatively, you can use (to navigate between edit locations in one or different files)
> Navigate back: `kb(workbench.action.navigateBack)`
> Navigate forward: `kb(workbench.action.navigateForward)`
> 
> See [Quick file navigation](https://code.visualstudio.com/docs/editor/editingevolved#_quick-file-navigation) for details.

gregvanl:
[...]

PR #5379, opened 2022-06-02
Update tips-and-tricks.md - add details

ilyachalov:
Add details and corrections in section 'Navigation history'. See Issue #5378 for details.

### DIFF
--- a/docs/getstarted/tips-and-tricks.md
+++ b/docs/getstarted/tips-and-tricks.md
@@ -453,13 +453,23 @@ Keyboard Shortcut: `kb(workbench.action.closeFolder)`
 
 ### Navigation history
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @ilyachalov Thank you for creating a PR and your continued interest in the VS Code documentation.
The goal of the Tips and Tricks topic is to be a very fast, easily scannable set of keyboard shortcuts and links for users who want to ramp up quickly. I agree the wording for the "Navigation history" is thin but I think users who see the keyboard shortcuts and try them out will quickly understand how they work. 
I'd like to keep that section lean so I don't think we want to add a table or more text.

ilyachalov:
@gregvanl Thanks for the explanation! I'll keep that in mind.

gregvanl:
Hi @ilyachalov We have really appreciated your help on our documentation and look forward to more collaboration in the future.

PR #5382, opened 2022-06-03
Add release notes for Jupyter extension

rchiodo:


rebornix (approved):

amunger (approved):

greazer (commented):
Looks good, though isn't there some extra knowledge/actions.involved if trying to connect to https? If so, we could add an item to Discussions that we can reference here.

greazer (commented):
Actually we should probably just have a single Discussion item that indicates some of the limitations for what's here. 

This includes, ipywidget support isn't full yet, variable viewing doesn't support dataframe viewing, interactive window doesn't support debugging. 

Anything else?

rchiodo:
Intellisense doesn't work. Although we're not really providing it anymore.

rchiodo:
Extra actions for https are only when creating a https server on your own with an invalid certificate. The browser won't accept connections to that. There's a wiki page about it that we could mention in the discussion item.

rchiodo:
I'll add a discussion item and link to it in the release notes.

rchiodo:
Oh looks like you did that already.

greazer:
> 

Yeah, and am in the middle of trying to juggle a mac and a pc, so if you want to edit what I started, feel free.

Issue #5141, opened 2022-02-04
Add sample implementation of "Do not show again" option for every notification


hamirmahal:
The [Notification examples](https://github.com/microsoft/vscode-docs/blob/main/api/references/extension-guidelines.md#notification-examples) section of the docs says

> Add a **Do not show again option** for every notification

![image](https://user-images.githubusercontent.com/43425812/152590429-4bb13908-4c1b-4e08-aa25-f722a54663c7.png)

Where can I find a sample implementation of every notification having a functional **Do not show again** checkbox?

If one isn't readily available, can we add this to the documentation?

Also, what are your thoughts on adding a `Discussions` section to this repository for questions like these?

gregvanl:
Closing as only one upvote and looks like the community has lost interest

snehara99:
I would like an example of this as well.

PR #5383, opened 2022-06-04
add example code for "Don't show again" for #5141

hamirmahal:


daviddossett (changes requested):
Thanks for raising a PR @hamirmahal! These guidelines are intended to stay at an experience level (i.e. no explicit code examples). The general format we're following for now is to add links to extension samples, references, and guides to the new "Links" section on each page.

Could you move this to be one of the links in that section?

daviddossett:
Another option for bonus points would be to add an explicit notification sample to our extension samples and add a similar example for "Don't show again" there. 

https://github.com/microsoft/vscode-extension-samples

gregvanl:
Closing as the suggestion is to move the code snippet to an extension sample.
@hamirmahal Thanks for providing the sample code!

PR #5384, opened 2022-06-04
added install Git on local link as part of instruction as is required…

jveethree:
… to function


jveethree:
added bit about installing git as it is required in order for the extension to function.

gregvanl:
Good addition, thanks!

PR #5406, opened 2022-06-14
go.md: update contents and replace go extension's doc links with wiki links

hyangah:
Updates https://github.com/golang/vscode-go/issues/2094


gregvanl:
Thanks!

PR #5407, opened 2022-06-14
[BOYSCOUT] Use a modern key type to generate SSH keys

cloudcosmonaut:
Let's make sure users create keys with a modern key type.


gregvanl:
Thanks!

cloudcosmonaut:
Thanks for the merge!

PR #5411, opened 2022-06-15
Update folder name to be consistent in the example

tangx246:
Just a few lines below, the example references a different folder name in the `omnisharp.json` file. Updating for consistency


gregvanl:
Hi @tangx246 Thank you for submitting a PR and your interest in the VS Code documentation.
I think the folder name in the first paragraph is the parent folder for all NuGet packages, not specific to the Unity analyzer package. We want to recommend that users make the parent folder general.

PR #5412, opened 2022-06-16
Make code example easier to understand

EstebanDalelR:
The negation of `isHigher` is easy to miss.


gregvanl:
Thanks for the suggestion!

PR #5437, opened 2022-07-01
Add enterprise section

sana-ajani:
- Added Group Policy content in new Enterprise doc
- Moved hostnames section from Network to Enterprise doc


joaomoreno (approved):
Did a couple changes:

- Left Network page untouched, since we link to it from many other places
- Created a link from Enterprise to Network regarding the common hostnames
- Rephrased some things

PR #5439, opened 2022-07-01
Add Terminal Shell Integration page

### LINKED ISSUE #5219: Create a user path for setting up shell integration manually
The more I think about it, the more I think we should use 'proprietary' sequences and not rely on the more standard shell integration ones as they could confuse VS Code and we want to add additional capabilities on top. So I imagine there would be a shell integration section in the docs that explains how to source the built-in scripts, and not documenting the sequences we support as we want flexibility to change the protocol.

Tyriar:
Fixes #5219

### DIFF
--- a/docs/editor/terminal-shell-integration.md
+++ b/docs/editor/terminal-shell-integration.md
@@ -81,8 +81,93 @@ The script below contains `<InstallDir>` which must be replaced by VS Code's ins
 
 ## Features
 
-###
+### Command decorations and the overview ruler
+
+One of the things shell integration enables is the ability to get the exit codes of the commands run within the terminal. Using this information, decorations are added to the left of the line to indicate whether the command succeeded of failed. These decorations also show up in the relatively new overview ruler in the scroll bar, just like in the editor.
+
+![Blue circles appear next to successful commands, red circles with crosses appear next to failed commands. The color of the circles appears in the scroll bar](images/terminal-shell-integration/decorations.png)
+
+The decorations can be interacted with to give some contextual actions like re-running the command:
+
+![Clicking a successful command decoration shows a context menu containing items: Copy Output, Copy Output as HTML, Rerun Command and How does this work?]((images/terminal-shell-integration/decoration-menu.png)
+
+The command decorations can be configured with the following settings:
+
+- `terminal.integrated.shellIntegration.decorationIcon`
+- `terminal.integrated.shellIntegration.decorationIconSuccess`
+- `terminal.integrated.shellIntegration.decorationIconError`
+
+### Command navigation
+
+The commands detected by shell integration feed into the command navigation feature (`ctrl/cmd+up`, `ctrl/cmd+down`) to give it more reliable command positions. This feature enables quick navigation between commands and selection of their output.
+
+### Run recent command
+
+The `Terminal: Run Recent Command` command surfaces history from various sources in a quick pick, providing similar functionality to a shell's reverse search (ctrl+r). The sources are the current session's history, previous session history for this shell type and the common shell history file.
+
+![The "run recent command" command shows a quick pick with commands ran previously that can be filtered similar to the go to file command](images/terminal-shell-integration/recent-command.png)
+
+`Alt` can be held to write the text to the terminal without running it. The amount of history stored in the previous session section is determined by the `terminal.integrated.shellIntegration.history` setting.
+
+There is currently no keybinding assigned by default but it can be hooked up to `ctrl+space` for example with the following keybinding:
+
+```json
+{
+    "key": "ctrl+space",
+    "command": "workbench.action.terminal.runRecentCommand",
+    "when": "terminalFocus"
+},
+```
+
+### Go to recent directory
+
+Similar to the run recent command feature, the `Terminal: Go to Recent Directory` command keeps track of directories that have been visited and allows quick filtering and `cd`ing to them.
+
+`Alt` can be held to write the text to the terminal without running it.
+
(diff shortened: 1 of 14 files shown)


docs/editor/terminal-shell-integration.md
```diff
@@ -0,0 +1,173 @@
...
+TOCTitle: Terminal Shell Integration
+ContentId: a6a1652b-c0d8-4054-a2da-feb915eef2cc
+PageTitle: Terminal Shell Integration in Visual Studio Code
+DateApproved: 6/9/2022
+MetaDescription: Visual Studio Code's integrated terminal has the ability to integration with various shells, enhancing the capabilities of the terminal.
+---
+
+# Terminal Shell Integration
+
+Visual Studio Code has the ability to integrate with common shells, allowing the terminal to understand more about what's actually happening inside the shell. This additional information enables some [compelling features](#_features) such as command/directory detection, command decorations and command navigation.
```
meganrogge, line 13:
```suggestion
Visual Studio Code has the ability to integrate with common shells, allowing the terminal to understand more about what's actually happening inside the shell. This additional information enables some [compelling features](#_features) such as command/directory detection, point of interest decorations and command navigation.
```

docs/editor/terminal-shell-integration.md
```diff
@@ -0,0 +1,173 @@
...
+Supported shells:
+
+- Linux/macOS: bash, pwsh, zsh
+- Windows: pwsh
+
+## Installation
+
+### Automatic script injection
+
+The standard way to activate shell integration is to simply set the `terminal.integrated.shellIntegration.enabled` setting to true. Enabling this setting will automatically "inject" the shell integration script into the shell session via shell arguments and/or environment variables when the setting is enabled. This is designed to be an easy way to get started with shell integration but it doesn't work in some more advanced use cases like in sub-shells, through ssh (when not using the [Remote - SSH extension](https://code.visualstudio.com/docs/remote/ssh)) or for some complex shell setups. For those the ability to [manually install the feature](#_manual-installation) is available.
```
meganrogge, line 24:
```suggestion
The standard way to activate shell integration is to set the `terminal.integrated.shellIntegration.enabled` setting to `true`. When enabled, the shell integration script is injected into the shell session via shell arguments and/or environment variables. 

This standard, easy way will not work for some advanced use cases like in sub-shells, through ssh (when not using the [Remote - SSH extension](https://code.visualstudio.com/docs/remote/ssh)) or for some complex shell setups. The recommended way to enable shell integration for those is [manual installation](#_manual-installation).
```

docs/editor/terminal-shell-integration.md
```diff
@@ -0,0 +1,173 @@
...
+
+The command decorations can be configured with the following settings:
+
+- `terminal.integrated.shellIntegration.decorationIcon`
+- `terminal.integrated.shellIntegration.decorationIconSuccess`
+- `terminal.integrated.shellIntegration.decorationIconError`
+
+### Command navigation
+
+The commands detected by shell integration feed into the command navigation feature (`ctrl/cmd+up`, `ctrl/cmd+down`) to give it more reliable command positions. This feature enables quick navigation between commands and selection of their output.
```
meganrogge, line 102:
```suggestion
The commands detected by shell integration feed into the command navigation feature (`ctrl/cmd+up`, `ctrl/cmd+down`) to give it more reliable command positions. This feature allows for quick navigation between commands and selection of their output.
```

docs/editor/terminal-shell-integration.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Run recent command
+
+The `Terminal: Run Recent Command` command surfaces history from various sources in a quick pick, providing similar functionality to a shell's reverse search (ctrl+r). The sources are the current session's history, previous session history for this shell type and the common shell history file.
```
meganrogge, line 106: should we mention that it has additional functionality like opening the output in an editor w a click?

docs/editor/terminal-shell-integration.md
```diff
@@ -0,0 +1,173 @@
...
+
+The current working directory is used to resolve links against, showing the directory a recent command ran within as well as the `"terminal.integrated.splitCwd": "inherited"` feature.
+
+## Supported escape sequences
+
+VS Code supports several custom escape sequences:
+
+### VS Code custom sequences `OSC 633 ; ... ST`
+
+This is VS Code's custom escape sequences that are designed currently to only be triggered from within the shell integration script. We may document these in the future which would make it easier for shells to create their own VS Code shell integration scripts.
```
meganrogge, line 140:
```suggestion
VS Code's custom escape sequences are currently designed to only be triggered from within the shell integration script. We may document these in the future which would make it easier for shells to create their own VS Code shell integration scripts.
```

meganrogge (approved):
Looks great 👍🏼 just a few tweaks

docs/editor/terminal-shell-integration.md
Tyriar: Good idea https://github.com/microsoft/vscode-docs/pull/5439/commits/7b9eebe7ab433ed5a06a838c5d85ff281e7bcb25

Also updated the release notes to include the update

docs/editor/terminal-shell-integration.md
meganrogge: ok how about `such as directory detection and command detection, decorations, and navigation`?

Tyriar: I improved the wording, good idea since it's one of the first things in the page 👍

PR #5444, opened 2022-07-02
docs: a proper command to install autopep8

aiishymanel:
The previous command installed `pep8` package.

Maybe previously it was the same as `autopep8`, but now its a separate package, and the command to install `autopep8` has changed.


fathifesaziza:
hello  fathi medos  read  your   description

Nelkins3 (approved):

PR #5446, opened 2022-07-04
Fixed typo

1amchris:
Fixed a missing quote (perhaps a copy-paste mistake).
See images below.

Current:
Screen Shot 2022-07-04 at 10 35 34 AM

Fix: 
Screen Shot 2022-07-04 at 10 36 27 AM


1amchris:
Another way we could write it is by removing all instances of quotes, since it isn't used anywhere else when referring value, but I'll leave that up to you to tell.

gregvanl:
Thanks!

PR #5450, opened 2022-07-05
Update linux.md

dlech:
Place keyring in `/etc/apt/keyrings/` instead of `/etc/apt/trusted.gpg.d/`. This follows the official recommendation in https://manpages.ubuntu.com/manpages/jammy/en/man8/apt-key.8.html#deprecation.

### DIFF
--- a/docs/setup/linux.md
+++ b/docs/setup/linux.md
@@ -34,7 +34,7 @@ Installing the .deb package will automatically install the apt repository and si
 ```bash
 sudo apt-get install wget gpg
 wget -qO- https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > packages.microsoft.gpg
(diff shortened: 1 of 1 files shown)


Tyriar (commented):
The docs say this only applied from apt >= 2.4, my WSL instance has 2.2 and no `/etc/apt/keyrings` folder so I think this would break there?

Tyriar:
@dlech `mkdir -p` sounds like it should work fine, there wouldn't be an issue with older apt versions using the key in `keyrings/` since the deb points directly at it, right?

dlech:
That is correct (since we specify the directory using `signed-by` in the `.list` file).

dlech:
Since this already used the `install` command , I just added the `-D` option so it is one less line to copy and paste.

Tyriar (approved):
Thanks @dlech 👍

PR #5456, opened 2022-07-06
Add settings for inlay hints with Pylance

luabud:
In the next release, we are adding two types of inlay hints: for variable types and for return types. These two hints are behind feature flags: "python.analysis.inlayHints.functionReturnTypes" and "python.analysis.inlayHints.variableTypes". This PR adds these settings to the docs.


rjmolyneaux:
Thanks @luabud ! Merging this PR. ☺️

PR #5460, opened 2022-07-08
Fix directory command

timdaman:


Tyriar (approved):
Thanks 👍 

FYI coming up in the next release we've made some improvements to installing https://github.com/microsoft/vscode-docs/blob/vnext/docs/editor/terminal-shell-integration.md#manual-installation

PR #5462, opened 2022-07-11
change in CPP Docs

vinaykulk621:
Grammar change in cpp documentation


gregvanl:
Hi @vinaykulk621 Thank you for submitting a PR and your interest in the VS Code documentation.
I think the current version of the sentence reads well and implies that it is talking about a C++ toolset. I don't think we need another article for debugger.

PR #5466, opened 2022-07-12
Fix spelling of inline completions setting name

rzhao271:
Fixes https://github.com/microsoft/vscode-docs/issues/5465


gregvanl:
Thanks!

PR #5467, opened 2022-07-12
Text change

stech11845:
### Changes

- This change just adds "time" to index of API docs and "docs" instead of "doc", which feels a bit nicer (to me, wonder how other people think "time" matters, I have never published extensions before).

### Files changed
`api/index.md`


gregvanl:
Thanks, nice refinements

stech11845:
> Thanks, nice refinements

As always!

PR #5473, opened 2022-07-14
Fix environment variables in PowerShell code

alexandair:
There is no need to call environment variables in PowerShell using ${HOME}.
$HOME is enough.


alexandair (commented):
There is no need to call environment variables in PowerShell using `${HOME}`.
`$HOME` is enough.

TylerLeonhardt (approved):

connor4312 (approved):

PR #5474, opened 2022-07-14
SignatureInformation.activeParameter docs corrected

jbicker:
This corrects the docs regarding `SignatureInformation.activeParameter`. 

Current:
If provided, this is used in place of [SignatureHelp.activeSignature](https://code.visualstudio.com/api/references/vscode-api#SignatureHelp.activeSignature).

Expected:
If provided, this is used in place of [SignatureHelp.activeParameter](https://code.visualstudio.com/api/references/vscode-api#SignatureHelp.activeParameter).


gregvanl:
Hi @jbicker Thanks for spotting this!
vscode-api.md is auto generated from https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.d.ts.
Would you like to open a PR in the main VS Code product repo? Once the comment is fixed there, it will be picked up and published on the website with the next release.

PR #5475, opened 2022-07-14
Add the -y flag in apt-get example in dockerfile

jlian:
This change helps avoid users who update the dockerfile following docs example from running into container creation failures. Without the `-y` flag it hangs at the interactive prompt and the container fails to create.


chrmarti (approved):

gregvanl:
Thanks!

PR #5477, opened 2022-07-15
Clarify how env files are read

rjmolyneaux:
@brettcannon - Getting the wording down was/is a little tricky for this explanation 😂 I added a note before variable substitutions are introduced in the document. Let me know if makes more sense to keep it here or place this after the Variable substitution, H3. I'm also happy to alter the verbiage of the note as well!


docs/python/environments.md
```diff
@@ -252,6 +252,8 @@ An environment variable definitions file is a simple text file containing key-va
 
 By default, the Python extension looks for and loads a file named `.env` in the current workspace folder, then applies those definitions. The file is identified by the default entry `"python.envFile": "${workspaceFolder}/.env"` in your user settings (see [General Python settings](/docs/python/settings-reference.md#general-python-settings)). You can change the `python.envFile` setting at any time to use a different definitions file.
 
+>**Note**: Environment definition files are not read by the Python extension. Unless Visual Studio Code documentation states otherwise, these only affect certain files as per their definition.
```
brettcannon, line 255: They are read by the Python extension, they are only _used_ in certain situations.

```suggestion
>**Note**: Environment definition files are not used in all situations where environment variables are available for use. Unless Visual Studio Code documentation states otherwise, these only affect certain scenarios as per their definition.
```

karrtikr: I think we should add an example:

> For example: Extension does not use it when resolving setting values.

PR #5501, opened 2022-07-29
Added VS build tools are required for omnisharp to function properly

A-R-Lyons:
For intellisense to work with .net5, VS build tools now need to be downloaded and installed manually.  Added a section to that effect


gregvanl:
Thanks!

PR #5506, opened 2022-07-29
Explain better the host process

EstebanDalelR:
This info was on the meta description but not on the page.


gregvanl:
Hi @EstebanDalelR Thanks for opening a PR and your interest in the VS Code documentation.
The extension host can be either Node.js or WebWorker, when running in the browser, so I think leaving the opening statement general is better.

PR #5550, opened 2022-08-05
Document the package manifest sponsors field

remcohaszing:
This is already supported in VS Code and the JSON schema for package.json, but it wasn’t mentioned in the documentation.


gregvanl:
Thanks!

PR #5553, opened 2022-08-07
Add missing gcompat dependency for alpine

agowa338:
Add the gcompat dependency to the list of dependencies required on alpine.
Without it the node binary that is copied to this system does not run.


Chuxel:
`gcompat` shouldn't be strictly required as far as I know.  Some extensions may rely on glibc, however.  @chrmarti Do you know if anything changed here?  I know we auto-install other deps.

chrmarti:
@agowa338 With which version of Alpine are you seeing `gcompat` as required?

agowa338:
I was just trying to set it up and installing gcompat made a lot of errors go away.
But now it also works without it 🤔 Probably some extension...

PR #5557, opened 2022-08-08
Use double quote to prevent globbing and word splitting

Delgan:
Hi!

I just copy/pasted the recipe to persist bash history in my Dockerfile, and [`hadolint`](https://github.com/hadolint/hadolint) (which is a linter for Dockerfile) raised a warning (SC2086).

In this specific case, double-quotes does not matter much, but at least it will avoid the warning to be raised by default for people copy/pasting this documented recipe and using Hadolint. :+1:


gregvanl:
Thanks!

PR #5559, opened 2022-08-09
Add a git bash section to shell integration

Tyriar:
Part of microsoft/vscode#143769, this is new coming in 1.71

### DIFF
--- a/docs/terminal/shell-integration.md
+++ b/docs/terminal/shell-integration.md
@@ -61,7 +61,7 @@ Add the following to your `~/.zshrc` file. Run `code ~/.zshrc` in bash to open t
 
 ⚠️ This is an experimental feature and automatic injection is not supported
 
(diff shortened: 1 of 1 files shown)


meganrogge (approved):

IllusionMH:
> and a minor change will improve it further.

So is it already makes sense to report issues related to Git Bash or it's better to wait for future change that should improve situation as it's mentioned in comment?

Tyriar:
@IllusionMH let's try consolidate issues in #143769 for now to reduce noise as I suspect a lot of git bash problems will also apply to Windows pwsh, we can always fork them into a new issue if needed. For example there will be a few problems with the decorations jumping around, those have the same root cause as https://github.com/microsoft/vscode/issues/153476; conpty may put the sequences in the wrong spot and we try fix up the position after the command has run.

Tyriar:
https://github.com/microsoft/vscode/pull/157697 was the minor change btw 🙂

PR #5563, opened 2022-08-10
Use transpiling over compiling

bennycode:
Throughout the article it is written that `tsc` transpiles TypeScript source code to JavaScript. That's why I find it important to also state that in the beginning of the article.


gregvanl:
Thanks

PR #5564, opened 2022-08-10
Add info on logical lines

KyleMit:
This issue was raised in #26393 and fixed  in #135805. 

It got a shout out in the [October 2021 release notes](https://code.visualstudio.com/updates/v1_62#_thank-you:%7E:text=InsertCursorAbove/Below), but was never documented anywhere on the site.

Since it's not configured via a setting, has even less discoverability that someone would know to add this specific command to the keybind args.

Also asked about on Stack Overflow [How to avoid multi cursor on wrapped lines in VS Code?](https://stackoverflow.com/q/53283020/1366033)

### DIFF
--- a/docs/editor/codebasics.md
+++ b/docs/editor/codebasics.md
@@ -55,6 +55,25 @@ Here's an example of expanding the selection with `kb(editor.action.smartSelect.
 
 ![Expand selection](images/codebasics/expandselection.gif)
 
+### Insert cursors on Logical Lines
+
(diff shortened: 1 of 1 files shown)


ProKashif (approved):

gregvanl:
@KyleMit Apologies on the delay on this PR. I've been wrestling with where this nice example should go given it's a fairly advanced configuration, requiring the reader to know about keybindings, editing JSON, and command arguments.
Having it early on in Basic Editing might be confusing for some people. 
I'll merge your PR and then move it down to the Command Questions section maybe under a question like "How can I avoid extra cursors on wrapped lines?"

gregvanl (approved):
Thanks! And sorry for the wait

PR #5586, opened 2022-08-23
docs: Update the Tree file filtering doc section in getStarted userIn…

MohamedLamineAllal:
…terface

The find control tree filtering was implemented and is the current. And the old way of tree file filtering through focus > type is deprecated. 

This PR updates the Tree file filtering so it reflects the new changes. As it can be confusing if you read the documentation. And you get something that doesn't work.

- updated the section with the release note. As it's complete and clear.
- Added a note about deprecation and what version. With a link to the release note.
- All links and images are mapped internally with relative paths.

### DIFF
--- a/docs/getstarted/userinterface.md
+++ b/docs/getstarted/userinterface.md
@@ -106,11 +106,22 @@ You can select multiple files in the **File Explorer** and **OPEN EDITORS** view
 
 ### Filtering the document tree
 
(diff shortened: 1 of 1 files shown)


MohamedLamineAllal:
The doc was updated after this PR was created. So the PR is obsolete now

gregvanl:
Hi @MohamedLamineAllal, we still appreciate you taking the time to create a PR and your interest in the VS Code documentation. Thank you!

PR #5598, opened 2022-08-28
correct yapf indent width

eltaara:
correct to intended indent


eltaara (commented):
correct to intended indent

gregvanl:
Thanks! 20 columns does seem excessive

PR #5599, opened 2022-08-28
Update mingw toolset instructions

MaxCreatesCode:
Updating the "Using GCC with MinGW" instructions page.  

Accounts for the removal of steps on MSYS2's Getting Started page, which is a dependency that these instructions rely on.
Without this change users following these instructions would be unable to compile C++ when using the default launcher specified by these instructions.  As these instructions acknowledge this is not meant to teach anything about GDB, MinGW, etc. so we can't expect the people using this page to correct this small gap on their own when following the instructions.  

Not sure if VS Code depends on all of these packages.  Might be able to get by on some minimal subset of packages but just including all of them simplifies the instructions.  One issue this resolves is that VS Code expects gdb.exe to be in the bin folder specified by `miDebuggerPath` but just installing `mingw-w64-x86_64-gcc` as [MSYS2's page](https://www.msys2.org/) specifies does not install `mingw-w64-x86_64-gdb`.

(Students are expecting to just download the prompted extension & be able to run C++...they are frustrated that this is not just dynamically bundled together & give up on VS Code when these instructions fail.)

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -29,10 +29,8 @@ To successfully complete this tutorial, you must do the following steps:
 
 1. Follow the **Installation** instructions on the [MSYS2 website](https://www.msys2.org/) to install Mingw-w64. Take care to run each required Start menu and `pacman` command.
 
-1. Install the Mingw-w64 toolset:
-   1. Have an open MSYS2 terminal. (Run "MSYS2 MinGW x64" from the Start menu.)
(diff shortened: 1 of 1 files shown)


ProKashif (approved):

404Nikhil (approved):

gregvanl:
Thanks! Apologies for the delay on getting this reviewed and merged.
The same instructions are also listed https://code.visualstudio.com/docs/languages/cpp#_example-install-mingwx64 and I'll update that topic to follow your instructions

PR #5610, opened 2022-09-01
Fix sentence about using Node.js APIs in web

felipecrs:
If I'm not mistaken, it should be the opposite (as in this PR). Right?


gregvanl:
Yep, confirmed with an engineer. Thanks again!

PR #5616, opened 2022-09-07
Fixed the description above the text sync example

Andarist:
Since the description above this code is this:
> Below is a server implementation that uses the provided simple text document manager that synchronizes text documents by always sending the file's full content from VS Code to the server

I think that this line was misleading here. Or perhaps this description is not accurate


dbaeumer:
Actual the description is not accurate. It should be:

Below is a server implementation that uses the provided text document manager that synchronizes text documents by always sending incremental deltas from VS Code to the server

Andarist:
@dbaeumer thanks for re-checking this. I've corrected the PR with the given suggestion

gregvanl:
@Andarist @dbaeumer Thank you both for this update!

PR #5623, opened 2022-09-09
Updating interpreter drop down images and install packages instructions 

cwebster-99:
In user research conducted on 9/1, we observed users struggle to follow the getting started instructions. When trying to select an interpreter, users were confused as to what they were looking at and if it was correct. We added screenshots to show users that the Select Interpreter drop down should look like both with and without a virtual environment option. Users also got confused when following the install packages instructions. Participants thought that everything after the "for mac" text was only for mac users. We moved the venv prompt note to fall below the creation since it appears after creation and to give more separation in the flow of instructions.


docs/python/python-tutorial.md
```diff
@@ -270,15 +269,19 @@ A best practice among Python developers is to avoid installing packages into a g
    Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope Process
    ```
 
-   **For macOS/Linux**
+   **Virtual Environment Creation for macOS/Linux**
```
luabud, line 272: I'm wondering if we should remove the capitalization from "environment creation"? 🤔

luabud (approved):

docs/python/python-tutorial.md
cwebster-99, line 272:
```suggestion
   **Virtual environment creation for macOS/Linux**
```

docs/python/python-tutorial.md
```diff
@@ -251,11 +253,8 @@ A best practice among Python developers is to avoid installing packages into a g
 > **Note**: For additional information about virtual environments, see [Environments](/docs/python/environments.md#creating-environments).
 
 1. Create and activate the virtual environment
-   >**Note**: When you create a new virtual environment, you should be prompted by VS Code to set it as the default for your workspace folder. If selected, the environment will automatically be activated when you open a new terminal.
 
-   ![Virtual environment dialog](images/tutorial/virtual-env-dialog.png)
-
-   **For Windows**
+   **Virtual Environment Creation for Windows**
```
cwebster-99, line 257:
```suggestion
   **Virtual environment creation for Windows**
```

docs/python/python-tutorial.md
cwebster-99: I like that better! I updated the file to reflect

gregvanl:
Thanks!

PR #5627, opened 2022-09-12
Fix macOs installation instruction for sshfs

KindDragon:
Same as in this guide https://www.telepresence.io/docs/latest/troubleshooting/#volume-mounts-are-not-working-on-macOS


gregvanl:
Thanks!

Issue #4809, opened 2021-09-03
Review performance and update integrated terminal topic


gregvanl:
https://code.visualstudio.com/docs/editor/integrated-terminal
August CSAT FY21Q4 32%
Several product issues and new features during May and June.

Tyriar:
AFAIK the reason for this was the introduction of some links to the page into the settings, using a query param to indicate it came from VS Code would allow us to segment access via troubleshooting links vs organic.

gregvanl:
Next to do pull content out of docs/terminal/basics into advanced and reference

Tyriar:
Current full ToC of basics page:

- Terminal Basics
  - Terminal shells
  - Managing terminals
    - Grouping
    - Customizing Tabs
  - Terminal profiles
    - Configuring profiles
    - Removing built-in profiles
    - Configuring the task/debug profile
  - Working directory
  - Terminal process reconnection
  - Links
  - Local echo
  - Terminal appearance
  - Copy & Paste
  - Using the mouse
    - Right-click behavior
    - Alt click
  - Keybindings and the shell
    - Chord keybindings in the terminal
    - Send text via a keybinding
  - Find
  - Run selected text
  - Automating launching of terminals
  - Next steps
  - Common questions
    - I'm having problems launching the terminal
    - Can I use the integrated terminal with the Windows Subsystem for Linux?
    - Why is VS Code shortcut X not working when the terminal has focus?
    - Can I use Cmder's shell with the terminal on Windows?
    - Why is Cmd+k/Ctrl+k not clearing the terminal?
    - Why is nvm complaining about a prefix option when the integrated terminal is launched?
    - Can I use Powerline fonts in the integrated terminal?
    - How do I configure zsh on macOS to jump words with Ctrl+Left/Right arrow?
    - Why does macOS make a ding sound when I resize terminal split panes?
    - Why is my terminal showing a multi-colored triangle or a completely black rectangle?
    - Why are there duplicate paths in the terminal's $PATH environment variable and/or why are they reversed?
    - I'm having problems with the terminal rendering. What can I do?
    - Git Bash isn't saving history when I close the terminal
    - I see 1~ or [201~ when I paste something
    - Ctrl+A, Ctrl+R output ^A, ^R on zsh
    - Why are the colors in the terminal not correct?

Tyriar:
Proposed new structure below:


## Legend

- :open_hands: Usage
- :gear: Config
- :art: Appearance
- :zap: Advanced
[...]

PR #5630, opened 2022-09-13
Terminal advanced usage page

Tyriar:
Part of #4809

### DIFF
--- a/docs/terminal/advanced.md
+++ b/docs/terminal/advanced.md
@@ -1,0 +1,91 @@
+---
+Order: 4
+Area: terminal
+TOCTitle: Advanced
+ContentId: D458AFDC-C001-43FD-A4BB-9474767B2C04
+PageTitle: Advanced Terminal Usage in Visual Studio Code
+DateApproved: 9/1/2022
+MetaDescription: Visual Studio Code's integrated terminal has several advanced features.
+---
+# Terminal Advanced
+
+## Process reconnection
+
+terminal.integrated.enablePersistentSessions
+terminal.integrated.persistentSessionReviveProcess
+terminal.integrated.persistentSessionScrollback
+
+## Keybinding and the shell
+
+TODO: Explain the conflicts that can occur when hosting a terminal inside another application
+➡️ Why is VS Code shortcut X not working when the terminal has focus?
+
+terminal.integrated.sendKeybindingsToShell
+terminal.integrated.commandsToSkipShell
+
+### Chords
+
+⚡ 👐 ⚙️ Chord keybindings in the terminal
+terminal.integrated.allowChords
+
+Special case of chords:
+➡️ Why is Cmd+k/Ctrl+k not clearing the terminal?
+Merge into keybindings section
+
+### Mnemonics
+
+terminal.integrated.allowMnemonics
+
+### Custom sequence keybindings
+
+⚡ ⚙️ Send text via a keybinding
+workbench.action.terminal.sendSequence
+➡️ ⚙️ How do I configure zsh on macOS to jump words with Ctrl+Left/Right arrow?
+Include example in keybindings section
+
+## Confirmation dialogs
+
+terminal.integrated.confirmOnExit
+terminal.integrated.confirmOnKill
+terminal.integrated.showExitAlert
+
+## Auto replies
+
+terminal.integrated.autoReplies
+
+## Unicode and emoji support
+
+terminal.integrated.unicodeVersion
+
+## Windows and ConPTY
+
+terminal.integrated.windowsEnableConpty
+
+## Process environment
+
+TODO: Brief description of what the shell environment is
+
+### Environment inheritance
+
+terminal.integrated.inheritEnv
+
+### Interaction with `$LANG`
+
+terminal.integrated.detectLocale
+
+### Extension environment contributions
+
+terminal.integrated.environmentChangesRelaunch
+
+## Remote development
+
+### Reducing remote input latency
+
+terminal.integrated.localEchoLatencyThreshold
+terminal.integrated.localEchoEnabled
+terminal.integrated.localEchoExcludePrograms
+terminal.integrated.localEchoStyle
+
+### Local terminals in remote windows
+
+workbench.action.terminal.newLocal
--- a/docs/terminal/advanced.md
+++ b/docs/terminal/advanced.md
@@ -119,7 +119,7 @@ No auto replies are configured by default as this deals with shell input which i
 
 The terminal has both unicode and emoji support, since these are being used in a terminal environment there are some caveats to that support:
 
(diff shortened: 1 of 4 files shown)


docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+The terminal supports two different types of persistent sessions:
+
+- Process reconnection: When reloading a window (eg. after installing an extension), _reconnect_ to the previous process and restore its content.
+- Process revive: When restarting VS Code, a terminal's content is restored and the process is _relaunched_ using its original environment.
+
+Both of these can be disabled by setting `terminal.integrated.enablePersistentSessions` to `false`, and the amount of scrollback restored is determined by `terminal.integrated.persistentSessionScrollback`. Process revive can be configured independently with `terminal.integrated.persistentSessionReviveProcess`.
+
+### Moving terminals between windows
+
+It's possible to drag terminal tabs across different windows to move them under that window. This can also be done manually through the command palette and the `Terminal: Detach Session` and `Terminal: Attach to Session` commands.
```
meganrogge, line 23:
```suggestion
Terminal tabs can be dragged and dropped between VS Code windows. This can also be done manually through the command palette and the `Terminal: Detach Session` and `Terminal: Attach to Session` commands.
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Keybinding and the shell
+
+An issue when embedding a terminal within another application is the question of where to send each keybinding; to the shell hosted in the terminal or to the application? VS Code answers this by having some sane defaults that use a hardcoded list of commands that should always "skip the shell" and be handled by VS Codes keybinding system. This list is fully configurable to fine tune what commands should and shouldn't work:
```
meganrogge, line 27:
```suggestion
As an embedded application, the integrated terminal should intercept some, but not all, keybindings dispatched within VS Code. 

The configurable `"terminal.integrated.commandsToSkipShell"` setting determines which command's keybindings should always "skip the shell" and be handled by VS Codes keybinding system.
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+This feature supports [variable substitution](/docs/editor/variables-reference.md).
+
+Note that the command only works with the `\u0000` format for using characters via their character code (not `\x00`). Read more about these hex codes and the sequences terminals work with on the following resources:
+
+* [XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html)
+* [List of C0 and C1 control codes](https://github.com/xtermjs/xterm.js/blob/0e45909c7e79c83452493d2cd46d99c0a0bb585f/src/common/data/EscapeSequences.ts)
+
+## Confirmation dialogs
+
+In order to avoid additional noise and user actions, the terminal does not show warning dialogs when exiting processes. If warnings are desirable they can be configured with the following settings:
```
meganrogge, line 94:
```suggestion
In order to avoid additional noise and user actions, the terminal does not show warning dialogs when processes exit. If warnings are desirable they can be configured with the following settings:
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+
+- `terminal.integrated.confirmOnExit` - Controls whether to confirm when the window closes if there are active debug sessions.
+- `terminal.integrated.confirmOnKill` - Controls whether to confirm killing terminals when they have child processes.
+- `terminal.integrated.showExitAlert` - Controls whether to show the alert "The terminal process terminated with exit code" when exit code is non-zero.
+
+## Auto replies
+
+The terminal is capable of providing an input response to the shell automatically if an exact sequence of output is received. This could be used for all sorts of things but the most common use case is to automatically reply to the quirk when hitting `kbstyle(Ctrl+C)` in batch scripts that asks whether the user wants to terminate the batch job. To automatically dismiss this message, add this setting:
```
meganrogge, line 102:
```suggestion
The terminal is capable of providing a configurable input response to the shell automatically if an exact sequence of output is received. The most common use case is to automatically reply to the quirk when hitting `kbstyle(Ctrl+C)` in batch scripts that asks whether the user wants to terminate the batch job. To automatically dismiss this message, add this setting:
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+Notice the `\r` here which means `kbstyle(Enter)`, much like [custom sequence keybindings](#_custom-sequence-keybindings) this supports sending escape sequences to the shell.
+
+No auto replies are configured by default as this deals with shell input which is sensitive.
+
+## Unicode and emoji support
+
+The terminal has both unicode and emoji support, since these are being used in a terminal environment there are some caveats to that support:
```
meganrogge, line 118:
```suggestion
The terminal has both unicode and emoji support; since these are being used in a terminal environment there are some caveats to that support:
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Some unicode symbols have ambiguous width that may have changed between unicode versions. Currently we support unicode version 6 and 11 widths which can be configured with the `terminal.integrated.unicodeVersion` setting, this version should match the unicode version used by the shell/operating system, otherwise there could be rendering issues. Note that the unicode version of the shell/OS may not match the font's actual width.
```
meganrogge, line 120:
```suggestion
- Some unicode symbols have ambiguous width that may have changed between unicode versions. Currently we support unicode version 6 and 11 widths which can be configured with the `terminal.integrated.unicodeVersion` setting. That version should match the unicode version used by the shell/operating system, otherwise there could be rendering issues. Note that the unicode version of the shell/OS may not match the font's actual width.
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+- Some emoji that are made up of multiple combined characters may not render correctly, for example skin tone modifiers.
```
meganrogge, line 121:
```suggestion
- Some emojis comprised of multiple characters may not render correctly, for example skin tone modifiers.
```

docs/terminal/advanced.md
```diff
@@ -0,0 +1,182 @@
...
+- Emoji support on Windows is limited on Windows.
+
+## Process environment
+
+The process environment of the application running within the terminal can be influenced by various settings and extensions. Values in the environment is a common reason why output may look different in other terminals.
```
meganrogge, line 126:
```suggestion
The process environment of the application running within the terminal is influenced by various settings and extensions which can explain why output may look different in other terminals. 
```

meganrogge (approved):
👍🏼

docs/terminal/advanced.md
Tyriar: I think `,` might be more correct here? @gregvanl thoughts?

meganrogge: they're both independent clauses, so a semi-colon is correct here

gregvanl: I think I would break it up into two sentences because the second part is a lead to the caveats.
"The terminal has both Unicode and emoji support. When these characters are used in the terminal, there are some caveats to that support:"

gregvanl:
Thanks!

PR #5634, opened 2022-09-15
Fix typo: From dockerFile to dockerfile

krave1986:
The `dockerfile` property in `devcontainer.json` is wrongly cased as `dockerFile`.


gregvanl:
Thanks!

lq-bit (approved):
Thanks

lq-bit (approved):
Thanks

PR #5638, opened 2022-09-17
Fix error in converting probability to binary and update summary.

spicydonkey:
This pull request fixes an error in the `data-science-tutorial.md`, where an incorrect method was used to convert probability to binary outcome.  The correct model accuracy is incompatible with the previous summary.  This PR also fixes the summary.

The image file `images/data-science-tutorial/jupyter-cell-08.png` must be updated.


minsa110 (commented):
Thank you for the suggestion and for the PR! Before we approve, could you please let us know which warning you get for the blue squiggly under `rint`? We'd love to get that fixed before pushing the image into our docs.

minsa110 (approved):

PR #5639, opened 2022-09-17
Add command to run VS Code Insiders from WSL

prryplatypus:
I've ran into this issue a few times before, and every single time I forget why it doesn't work. Hopefully this will save me (and others) some time in the future :)


gregvanl:
Thanks. Good clarification

PR #5646, opened 2022-09-21
Git topic editing

digitarald:
- Added 3 videos for Git, including the intro video right on top.
- Moved SCM Provider content down in favor of Git-related content.
- Added mergetool config


gregvanl (approved):
Looks good. After Reynald's review, we can merge and see how the videos look on the staging server

PR #5650, opened 2022-09-21
Fix language names

rzhao271:
I kept seeing warnings related to invalid language names while building the website, and so I took a look at what language names were being used and changed them to fit ones listed in [linguist/languages.yml](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml).

For plain text, one can just use three backticks without specifying a language.

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -56,7 +56,7 @@ gdb --version
 
 From a Windows command prompt, create an empty folder called `projects` where you can place all your VS Code projects. Then create a sub-folder called `helloworld`, navigate into it, and open VS Code in that folder by entering the following commands:
 
-```cmd
+```bat
 mkdir projects
 cd projects
 mkdir helloworld
--- a/docs/cpp/config-msvc.md
+++ b/docs/cpp/config-msvc.md
@@ -59,7 +59,7 @@ If the Developer Command Prompt is using the BuildTools location as the starting
 
 From the Developer Command Prompt, create an empty folder called "projects" where you can store all your VS Code projects, then create a subfolder called "helloworld", navigate into it, and open VS Code (`code`) in that folder (`.`) by entering the following commands:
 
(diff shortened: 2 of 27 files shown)


docs/cpp/config-mingw.md
```diff
@@ -56,7 +56,7 @@ gdb --version
 
 From a Windows command prompt, create an empty folder called `projects` where you can place all your VS Code projects. Then create a sub-folder called `helloworld`, navigate into it, and open VS Code in that folder by entering the following commands:
 
-```cmd
+```bat
```
gregvanl, line 59: shouldn't this be `bash` like in the file above rather than `bat`?

docs/cpp/config-msvc.md
```diff
@@ -59,7 +59,7 @@ If the Developer Command Prompt is using the BuildTools location as the starting
 
 From the Developer Command Prompt, create an empty folder called "projects" where you can store all your VS Code projects, then create a subfolder called "helloworld", navigate into it, and open VS Code (`code`) in that folder (`.`) by entering the following commands:
 
-```cmd
+```bat
```
gregvanl, line 62: `bash` or `bat`?

gregvanl:
Thanks for doing this. Looks good, just wondering if we should be using `bash` or `bat` for command prompt code snippets

docs/cpp/config-mingw.md
rzhao271: The step says "from the windows command prompt"

docs/cpp/config-msvc.md
rzhao271: The step says "From the Developer Command Prompt", which also uses batch

docs/cpp/config-mingw.md
gregvanl: OK, got it. You were being precise in the context of the article.

PR #5651, opened 2022-09-21
Round two of fixing language names

rzhao271:
We use shiki as the highlighter, so the actual reference list is at [shiki/languages.md](https://github.com/shikijs/shiki/blob/main/docs/languages.md).
One thing to note is that the ssh-config name currently isn't recognized locally, but I'm hoping that updating shiki would help.

### DIFF
--- a/docs/remote/ssh.md
+++ b/docs/remote/ssh.md
@@ -143,7 +143,7 @@ Finally, you'll be asked to pick a config file to use. You can also set the `"re
 
 For example, entering `ssh -i ~/.ssh/id_rsa-remote-ssh yourname@remotehost.yourcompany.com` in the input box would generate this entry:
 
-```sshconfig
+```ssh-config
 Host remotehost.yourcompany.com
     User yourname
     HostName another-host-fqdn-or-ip-goes-here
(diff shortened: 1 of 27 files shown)


gregvanl (approved):

docs/remote/ssh.md
```diff
@@ -143,7 +143,7 @@ Finally, you'll be asked to pick a config file to use. You can also set the `"re
 
 For example, entering `ssh -i ~/.ssh/id_rsa-remote-ssh yourname@remotehost.yourcompany.com` in the input box would generate this entry:
 
-```sshconfig
+```ssh-config
```
gregvanl, line 146: This might not work looking at the vscode-docs CI build log.
"WARNING -- WILL NOT HIGHLIGHT CODE BLOCK (```) DUE TO ERROR :::No language registration for ssh-config"

rzhao271:
The newer shiki renames `dockerfile` to `docker`.
It adds support for `ssh-config` as well.
This PR should be merged at the same time the backend is updated.

gregvanl (approved):

PR #5652, opened 2022-09-22
Fix return type from vscode.executeDefinitionProvider

JQChan:
According to issue https://github.com/microsoft/vscode-docs/pull/4098 ，but not fix this file yet.

### DIFF
--- a/api/extension-guides/command.md
+++ b/api/extension-guides/command.md
@@ -40,7 +40,7 @@ async function printDefinitionsForActiveEditor() {
     return;
   }
 
-  const definitions = await vscode.commands.executeCommand<vscode.Location[]>(
+  const definitions = await vscode.commands.executeCommand<vscode.LocationLink[]>(
     'vscode.executeDefinitionProvider',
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @JQChan Thank you for your PR and your interest in the VS Code API documentation.
`executeDefinitionProvider` can return either an array of Location or LocationLink so the current code snippet is correct.

In https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostApiCommands.ts:

```TypeScript
	// -- go to symbol (definition, type definition, declaration, impl, references)
	new ApiCommand(
		'vscode.executeDefinitionProvider', '_executeDefinitionProvider', 'Execute all definition providers.',
		[ApiCommandArgument.Uri, ApiCommandArgument.Position],
		new ApiCommandResult<(languages.Location | languages.LocationLink)[], (types.Location | vscode.LocationLink)[] | undefined>('A promise that resolves to an array of Location or LocationLink instances.', mapLocationOrLocationLink)
```

PR #5657, opened 2022-09-24
Should reference engines.vscode not vscode.engine

golf1052:
The extension manifest shows `engines.vscode` https://code.visualstudio.com/api/get-started/extension-anatomy#extension-manifest. `vscode.engine` does not exist.


gregvanl:
Thanks!

PR #5675, opened 2022-10-03
Fix instructions to show builtin extensions

Naitreey:
On macOS, there isn't **Show Built-in Extensions** button. I need to search for `@builtin` to show them. So I added a separate instruction for the OS.

I can't verify whether the button is still there on Windows and Linux versions (no environment sorry).


gregvanl:
Hi @Naitreey The instructions are stale and should mention the **Filter Extensions** button, which will bring up the dropdown menu with **Built-in** as an option.

image

Would you like to update your PR to call out this flow?

Naitreey:
Hi @gregvanl, thanks for your feedback. I've updated PR accordingly, also with a new screenshot for the workflow. Please review if it's appropriate.

gregvanl:
Thanks! Looks great

PR #5680, opened 2022-10-04
simplify yeoman generation instructions

pburrows:
remove the instructions to globally install yeoman and use npm init instead


gregvanl:
Hi @pburrows Thank you for submitting a PR and your interest in the VS Code documentation.

I tried your suggestion after uninstalling yo and generator-code globally and it failed with "Error: Cannot find module 'libnpx'".

Using the existing tutorial instructions worked fine for me. I also notice that the office Yeoman site https://yeoman.io/ recommends installing Yeoman globally.

Closing this PR as the current documentation works well. Perhaps there are some drawbacks to the global installation that I'm unaware of but seems a common workflow.

PR #5690, opened 2022-10-06
docs: update Apple Silicon reference

mikeirvingweb:
changed M1 references to Apple Silicon now that there are new generations


gregvanl:
Thanks!

PR #5693, opened 2022-10-07
Update screenshots for Remote extension rename

gregvanl:


bamurtaugh (approved):
LGTM, thanks so much for taking the time, @gregvanl!

PR #5705, opened 2022-10-11
Change Python + Azure App Service tutorial link

gkulin:
@spboyer and I noticed that the tutorial to deploy Python web apps to Azure App Service was really more of a resource page for what to do (instead of a more standard tutorial). So, we want to change the tutorial link to a different tutorial more aligned with the goal of the page.


docs/python/python-on-azure.md
```diff
@@ -20,6 +20,6 @@ The following tutorials on the [Python Azure Developer's Center](https://learn.m
 
 Tutorial | Description
 --- | ---
-[Deploy Python web apps to Azure App Service](https://learn.microsoft.com/azure/python/tutorial-deploy-app-service-on-linux-01) | Deploy a web app using the [Azure App Service](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice) extension.
+[Deploy Python web apps to Azure App Service](https://learn.microsoft.com/en-us/azure/app-service/tutorial-python-postgresql-app?toc=%2Fazure%2Fdeveloper%2Fpython%2Ftoc.json&bc=%2Fazure%2Fdeveloper%2Fbreadcrumb%2Ftoc.json&tabs=flask%2Cwindows%2Cazure-portal%2Cterminal-bash%2Cazure-portal-access%2Cvscode-aztools-deploy%2Cdeploy-instructions-azportal%2Cdeploy-instructions--zip-azcli%2Cdeploy-instructions-curl-bash) | Deploy a web app with PostgreSQL database using the [Azure App Service](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice) extension.
```
gregvanl, line 23: That is a scary looking URL. Does it need to have so many query parameters? Maybe they are required for the appropriate tabs to be displayed?
Nit: Remove 'en-us' from URLs so that readers will get the localized version if non-English

gregvanl:
@gkulin I wonder if we should point people to https://learn.microsoft.com/azure/app-service/quickstart-python and not involve a PostgreSQL database?

gkulin:
@gregvanl what do you think of these changes? 

Instead of replacing the tutorial with the PostgreSQL database, I added the simpler quickstart tutorial you mentioned. Shayne also had the idea of adding a column to the table to call out & link related tools there. I realize this strays from the format of all the tutorial tables on the site, but if you like this I could go through and update the other tables to ensure consistency

gregvanl:
I think adding the related tools column is a good suggestion.
Rather than use a semi-colon delimited list, you could seperate with `<br>` tags for vertical list in the column. 
For an example, have a look at https://code.visualstudio.com/api/extension-guides/overview#guides-samples
You can see how to add the `<br>` tags in api\extension-guides\overview.md

I would also list the general framework and technologies first, like you did in the second row, and put the Azure specific tools later in the list.

I think you can leave out the Python extension since all the tutorials would use it and all the topics in the Python TOC mention that extension

gkulin:
> Rather than use a semi-colon delimited list, you could seperate with `<br>` tags for a vertical list in the colum

thanks! I didn't like the look of the semi-colon list but wasn't sure what else to do. The breaks are much better. I think the changes are all set now.

If I should go back and update the other deployment tables to match this format, would you prefer for that to all be in this PR or should it be separate for each content topic?

docs/python/python-on-azure.md
```diff
@@ -18,8 +18,9 @@ The [Azure Tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.
...
+[Deploy Python containers to Azure App Service](https://learn.microsoft.com/azure/developer/python/tutorial-deploy-containers-01) | Deploy a Docker container |  [Docker](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) <br> [Azure App Service](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice) <br> [Azure Tools extension pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-node-azure-pack)
+[Deploy Python to Azure Functions](https://learn.microsoft.com/azure/azure-functions/create-first-function-vs-code-python) | Deploy serverless code with Azure Functions | [Azure Functions Core Tools](https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local#install-the-azure-functions-core-tools) <br> [Azure Functions extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python) <br> [Azure Tools extension pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-node-azure-pack)
```
gregvanl, line 26: I think you can leave out "extension" for "Azure Functions extension" since you aren't using "extension" in the other links.
Also this link is pointing to the Python extension rather than the Azure Functions extension.

gregvanl:
@gkulin Let's do just the Python tutorial page for this PR. Once it's published you can create separate PRs for other tables you want to update and we can bring in the other teams such as the Java extension team for review.

spboyer (approved):

gregvanl (approved):
Looks good! 
I'll go ahead and merge and fix the extension link in a follow up PR this afternoon

gregvanl:
I'll also fix the Flask links to be https://flask.palletsprojects.com/

gregvanl:
Published

PR #5713, opened 2022-10-17
Changed grammar error on line 549

tonykipkemboi:
Before: `...you can you open folders inside WSL...` 
After:  `...you can use it to open folders inside WSL...`


gregvanl (approved):
Thanks!

tonykipkemboi:
I love OS!

PR #5731, opened 2022-10-25
Update jupyter-notebooks.md

IqraMahmood:


gregvanl:
Hi @IqraMahmood Thank you for taking the time to submit a PR and your interest in the VS Code documentation.

I don't think your suggestion is required as the VS Code website will render the Command Palette command appropriately for the readers platform (macOS, Windows, or Linux).

Here is what you should see on Windows:

image

gregvanl:
No response from author so closing the PR.

PR #5748, opened 2022-10-29
Add more details on auto imports and update default setting value

luabud:


luabud:
cc @heejaechang @bschnurr @debonte @rchiodo

hediet (dismissed):

docs/python/editing.md
```diff
@@ -22,45 +22,46 @@ IntelliSense is a general term for code editing features that relate to code com
...
 
-> **Tip**: Check out the [IntelliCode extension for VS Code (preview)](https://go.microsoft.com/fwlink/?linkid=2006060). IntelliCode provides a set of AI-assisted capabilities for IntelliSense in Python, such as inferring the most relevant auto-completions based on the current code context. For more information, see the [IntelliCode for VS Code FAQ](https://learn.microsoft.com/visualstudio/intellicode/intellicode-visual-studio-code).
+By default, only top-level symbols/packages are suggested, but you can customize this behavior through the `python.analysis.packageIndexDepths` setting. User defined symbols (i.e. not coming from installed packages or libraries) are only automatically imported if they have already been used in files that were previously opened in the editor. Otherwise, they will only be available through the [add imports Quick Fix](/docs/python/editing.md#quick-fixes).
```
rchiodo, line 38: Would it be worth it to add an example of package index depths setting? I guess reading this I'm not entirely sure how it works, so I was thinking an example would help me at least. 

Like what's a top-level symbol/package? First level in a directory? What if I didn't realize python libs are laid down as directory trees?

luabud: I added a link to the settings reference so they can see an example of how to use it, and added an example of what a top-level package import suggestion would look like, but I'm not sure this is a good one. @heejaechang do you have a better example in mind by any chance?

rchiodo (approved):

gregvanl (approved):
Thanks!

PR #5750, opened 2022-10-31
Moved image of Run and Debug view under the section describing the view. 

kev1010:
Added a parent Header (Run and Debug view and Run Menu)  and text to highlight the fact that there are two areas that allow controlling debugging. This was confusing previously.

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -12,8 +12,6 @@ MetaSocialImage: images/debugging/Debugging.png
 
 One of the key features of Visual Studio Code is its great debugging support. VS Code's built-in debugger helps accelerate your edit, compile, and debug loop.
 
-![Debugging diagram](images/debugging/debugging_hero.png)
-
 ## Debugger extensions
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @kev1010 Thank you for submitting a PR and your interest in the VS Code documentation.
We'd like to keep the overview image of the debugger UI at the beginning of the topic and prefer to keep the two parts of the UI separate; the Run and Debug view and the Run menu.

kev1010:
The image doesn't make sense unless the context is placed around it because there are two paths to start debugging and one of them doesn't even you the form in the image. So having the image at the top of the debugging section is misleading and confusing. This goes into the second issue of the text changes. The wording sounds like a contradiction to itself and makes the article illegible. The reader does not know there are two completely different paths to debugging because it is not intuitive that someone would want to do this.

Microsoft's dismissive attitude as shown in this issue negates the open-source philosophy. Even though my solution may not have been the best. It was better by at least highlighting the contradiction of two debugging paths. I was not encouraged to spend much time initially on this because of previous interactions with Microsoft and now it will be the coldest day before spending even this much time. Microsoft's open-source commitment is a falsehood.

Issue #5584, opened 2022-08-23
Clarify extension unique name restriction


isidorn:
Extension authors are confused about duplicate name errors. This is due to the fact that the Marketplace decided to make extensions unique on the name, as opposed to the publisher.name until recently. It's not clear for ext authors that (1) this change was made and (2) what the error is really about.

We should create a central documentation location we can direct them to, which makes it clear that the rules have changed? And, could we have the MP return a better error message?

The docs should be on the VS Code side and a disclaimer in vsce's README.md.

Examples of issues:
https://github.com/microsoft/vscode-vsce/issues/682#issuecomment-1198351782
https://github.com/microsoft/vscode-vsce/issues/761

@prashantvc agreed to drive this :)

Once we have better docs we should comment on those GitHub issues and clarify the confusion?

fyi @joaomoreno

prashantvc:
@isidorn I think we can close this

isidorn:
Thanks for updating the docs to make this more transparent!
Let's go ahead and close this, and I commented in your PR with some polish items. @prashantvc can you please also tackle them? Thanks!

PR #5759, opened 2022-11-03
Clarify extension unique name restriction issue #5584

prashantvc:
Fixed the issue #5584


gregvanl (approved):
Thanks!

isidorn:
@prashantvc do we really need to specify the error in the docs? The point of the error is that it is self-explanatory?
Also we should clarify that this is a new restriction.
For example "not allow" should be "no longer allow"

PR #5771, opened 2022-11-08
Update launch.json per current Flask launch.json

pamelafox:
When I asked VS Code to auto-create a Flask launch.json configuration, it did not quite match what was in the tutorial. This PR updates it to match what was generated.

I believe the changes make sense, too, since:
1) FLASK_ENV is deprecated in favor of FLASK_DEBUG
2) justMyCode was a recently added feature to VS Code?


luabud (approved):
Great catch, thank you! 
`justMyCode` was supported for a long time already, but it started being added by default into `launch.json` by the Python extension to increase its discoverability.

PR #5772, opened 2022-11-08
Make flask sample code style more Pythonic

pamelafox:
I noticed the imports were out of order, so I ran [shed](https://github.com/Zac-HD/shed) on the sample code. It fixed the imports and also added the standard 2 lines between definitions at the global layer.
I think it's nice when sample code follows Pythonic conventions so that people don't get warnings from linters running in their editor when they try to use a sample.

Perhaps the ordering is intentional, however, given you later describe how to sort imports? That's on views.py file though, so that learning goal could still be achieved. Up to you!


luabud (approved):

gregvanl:
Thanks!

PR #5779, opened 2022-11-10
Remove outdated info about the VSCode server private preview

helpimnotdrowning:
Server docs still contain paragraph with the link to the old sign up form to the private preview, this is no longer needed.


gregvanl (approved):
Thanks!

bhagatparwinder:
I noticed this, too, while setting up the server. I was curious about how often changes are deployed to production.

PR #5787, opened 2022-11-18
fix(docs): authorized_key -> authorized_keys

kongmoumou:
according [doc](https://github.com/PowerShell/Win32-OpenSSH/wiki/Security-protection-of-various-files-in-Win32-OpenSSH#authorized_keys), the file name should be `authorized_keys` 🤣


gregvanl:
Hi @kongmoumou Thanks for noticing the error and submitting a PR.
It looks like there is another instance of "`authorized_key` file" later in the topic. Would you also like to fix that mistake in this same PR? 
The second instance is in the paragraph just above "Improving your security with a dedicated key".

kongmoumou:
> Hi @kongmoumou Thanks for noticing the error and submitting a PR. It looks like there is another instance of "`authorized_key` file" later in the topic. Would you also like to fix that mistake in this same PR? The second instance is in the paragraph just above "Improving your security with a dedicated key".

done~

gregvanl (approved):
Thanks!

PR #5788, opened 2022-11-18
Typo Docker Contexts: Use

michielbdejong:
In my instance of VSCode, the command category is called "Docker Context**s**", plural.


gregvanl (approved):
Thanks for catching this!

PR #5791, opened 2022-11-19
Update activation-events.md

7flash:


7flash:

gregvanl:
Hi @7flash, thank you for submitting a PR and your interest in the VS Code API documentation.
While your change will fix navigation on GitHub, it isn't needed for in-product Markdown Preview and will break navigation on the live code.visualstudio.com site.
The preprocessing of relative Markdown links for topics under /api can't handle a Markdown file extension. 
Topics under /docs use a different preprocessor and so you might have seen relative Markdown links with the complete file extension. I admit it is a bit confusing.

PR #5799, opened 2022-11-23
Document overwriting a key binding with an empty command

justanotheranonymoususer:
It's mentioned here and it's very useful:
https://stackoverflow.com/a/43192243/2604492

Document it to help people find about it easier.


gregvanl (approved):
Thanks!

PR #5803, opened 2022-11-25
Update create-dev-container.md

JulianTurner:
messed up bracket


gregvanl:
Hi @JulianTurner Thank you for submitting a PR and your interest in the VS Code documentation.

The current example is correct for specifying a badge, it is a link `[]()` with an embedded image `` so that is why there is a leading bracket. 

`[](link path)`

You can see this example used in the https://github.com/microsoft/vscode-remote-try-java README.md

PR #5818, opened 2022-12-01
Git - unsafe repositories

lszomoru:


digitarald (approved):
LGTM, one thought I had was to mention the git CLI command which most other resources on this topic do; but we do link to the git config page already and the docs are about our commands.

docs/sourcecontrol/faq.md
```diff
@@ -65,6 +65,12 @@ One way to avoid these prompts is to set up a [credential helper](https://help.g
 
 Another option is to disable the auto fetch feature by changing the following setting: `"git.autofetch": false`.
 
+### Why is VS Code telling me that the git repository is potentially unsafe?
```
gregvanl, line 68: Did we want to go with:
"Why is VS Code warning that the git repository is potentially unsafe?"

"telling" is a little anthropomorphic and also not inclusive to deaf users

lszomoru: @gregvanl, no specific reason, and you are absolutely correct! 
Happy to rephrase if you have a better suggestion. Thanks!

gregvanl: Cool, I'll merge and update the wording when I do an edit pass on vnext.

PR #5819, opened 2022-12-01
update jupyter command

KyleMit:
You can see the command here without the `Jupyter:` category prefix:

image

But if you type in the command as it appears in the docs, the command palette will return empty:

image


onlinebanking11 (approved):

rebornix (approved):

gregvanl:
Thanks!

PR #5844, opened 2022-12-09
Fix some links that are broken when published

gjsjohnmurray:
Fix 3 broken links on https://code.visualstudio.com/docs/remote/vscode-server


gregvanl (approved):
Thanks

PR #5845, opened 2022-12-09
Clarify that closing remote VS Code closes tunnel; fix typo too

gjsjohnmurray:
Fixes a minor typo, and also adds a note to clarify that if the tunnel is enabled from the VS Code UI on the remote it will only remain available while VS Code is running there.


gregvanl:
Hi @gjsjohnmurray Thanks for pointing this out, I'm discussing with an engineer whether we need more documentation around this. I was surprised.
I've fixed the typo in another follow up PR.

connor4312 (approved):

connor4312:
This is currently the case, though this might change in the future (by having the UI register the tunnel as a system service)

PR #5856, opened 2022-12-11
Case python tutorial - python command works too

denisecase:
After installing Python (with conda), they may not need to use 

`python3 --version`

We find that the simpler

`python --version`

Works as well.


brettcannon:
> We find that the simpler
> 
> `python --version`
> 
> Works as well.

But you're not suggesting `python3 --version` doesn't work in all scenarios where Python 3 is installed? I'm trying to understand why listing two options is simpler to you than suggesting the use of `python3`?

denisecase:
Most of the people I introduce to Python have new installations, and the plain 'python' command is more consistent with other languages.  As the breaking changes from Python 2 days get further back in the rearview mirror, it seems less critical to keep the python3. Most only need python3 if they are also dealing with a legacy of version 2. Hopefully, someday, those will be in the past. 

I teach students to rely on official documentation when ever possible. This is a rare case where most of us don't do it the way the official documents indicate.

brettcannon:
Unfortunately, [Apple shipped Python 2 in macOS until 12.3](https://www.imore.com/apple-wont-be-including-python-2-its-macos-123-update). And with [macOS 12 still supported](https://en.wikipedia.org/wiki/MacOS_version_history#Releases) and not even two years old yet, I think Python 2 is still visible in the review mirror for users.

So while I appreciate the sentiment and effort put in to create this PR, I'm afraid for keeping the instructions simple and universally applicable, I am going to close this PR. But once macOS 12 is no longer supported I would support all of our docs to just say `python`.

PR #5858, opened 2022-12-12
Section on custom style sheets

aeschli:


api/extension-capabilities/overview.md
```diff
@@ -123,3 +123,10 @@ There are certain restrictions we impose upon extensions. Here are the restricti
...
 
 At VS Code, we're continually trying to optimize use of the underlying web technologies to deliver an always available, highly responsive editor and we will continue to tune our use of the DOM as these technologies and our product evolve. To ensure that extensions cannot interfere with the stability and performance of VS Code, and that we can continue to improve the DOM of VS Code without breaking existing extensions, we run extensions in an [Extension Host](/api/advanced-topics/extension-host) process and prevent direct access to the DOM.
+
+### No Custom Style Sheets
+
+A custom style sheet provided by users or extensions would work against the DOM structure and class names. These are not documented as we consider them internal. To evolve, refactor or improve VS Code, we need the freedom to make changes to the user interface. Each change tp the DOM can break existing custom style sheets. This would result in frustration with style sheet providers and also a bad user experience when user are impacted by UI glitches coming from the broken style sheet.
```
isidorn, line 129: typo - "each change tp the DOM"

isidorn (changes requested):
Can we also add something like this in the above Restriction section:

"These restrictions help guide extensions to not go against the design philosophy of VS Code. "

api/extension-capabilities/overview.md
isidorn: Also "when user are" should be "when users are"

PR #5866, opened 2022-12-14
Docker content refresh

ghogen:


docs/containers/docker-compose.md
```diff
@@ -167,7 +167,7 @@ For debugging Python with Docker Compose, follow these steps:
 
 ### .NET
 
-1. On the **Debug** tab, choose the **Configuration** dropdown, choose **New Configuration** and select the `Docker Attach` configuration template **.NET Core Docker Attach (Preview)**.
+1. On the **Debug** tab, choose the **Configuration** dropdown, choose **New Configuration** and select the `Docker Attach` configuration template **.NET Core Docker Attach**.
```
bwateratmsft, line 170: We haven't removed the "(Preview)" tag from this (though I don't have a good reason why). We are going to remove the "Core" beginning in 1.24.0 to align with branding of .NET, now that .NET Core 3.1 is EOL.

bwateratmsft: From my understanding though, ASP.NET Core is still called ASP.NET Core, so we haven't removed "Core" there. Or at least, that's what VS is doing, so I wanted to follow it.

ghogen, line 170:
```suggestion
1. On the **Debug** tab, choose the **Configuration** dropdown, choose **New Configuration** and select the `Docker Attach` configuration template **.NET Core Docker Attach (Preview)**.
```

ghogen: Reverted this one. @bwateratmsft  As this was the only comment, would you mark this as Approved?  We'll do some other files in a separate PR.

bwateratmsft (approved):

gregvanl (approved):

PR #5877, opened 2022-12-19
Terminal: Fix inputrc command for paste issues

orgads:


Tyriar (approved):
Thanks for spotting @orgads 👍

orgads:
Why did you remove the language?

gregvanl:
The Markdown code block renderer only supports a limited number of language IDs and `inputrc` is not supported and `bash` rendering would be incorrect.

PR #5880, opened 2022-12-21
add applocker note when running as admin

bpasero:
@deepak1556 I suggest to make an FAQ entry for running as admin on AppLocker that we can link to from the error message in the command line. Could you see if there are more details to fill in?

//cc @gregvanl

### DIFF
--- a/docs/setup/windows.md
+++ b/docs/setup/windows.md
@@ -103,3 +103,9 @@ Using the Registry Editor (regedit):
 3. Find the `.svg` key.
 4. Set its `Content Type` Data value to `image/svg+xml`.
 5. Exit `regedit`.
+
+### Unable to run as admin when AppLocker is enabled
+
+With the introduction of process sandboxing (see our [blog](https://code.visualstudio.com/blogs/2022/11/28/vscode-sandbox)) running as administrator is currently unsupported when AppLocker is configured on the system. Please run with your normal user account if possible.
+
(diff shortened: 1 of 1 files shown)


docs/setup/windows.md
```diff
@@ -103,3 +103,9 @@ Using the Registry Editor (regedit):
 3. Find the `.svg` key.
 4. Set its `Content Type` Data value to `image/svg+xml`.
 5. Exit `regedit`.
+
+### Unable to run as admin when AppLocker is enabled
+
+With the introduction of process sandboxing (see our [blog](https://code.visualstudio.com/blogs/2022/11/28/vscode-sandbox)) running as administrator is currently unsupported when AppLocker is configured on the system. Please run with your normal user account if possible.
```
deepak1556, line 109:
```suggestion
With the introduction of process sandboxing (see our [blog](https://code.visualstudio.com/blogs/2022/11/28/vscode-sandbox)) running as administrator is currently unsupported when AppLocker is configured on the system due to a limitation of the runtime sandbox, please refer to https://bugs.chromium.org/p/chromium/issues/detail?id=740132 for additional context. If your work requires to run the application from a elevated terminal, then you can launch with `--no-sandbox --disable-gpu-sandbox` as a workaround.
```

deepak1556 (commented):
This is a good idea 👍 , left minor improvement to the entry.

PR #5885, opened 2022-12-22
Fix missing reportFunctionMemberAccess

243f6a8885a308d313198a2e037:
There is a duplicate `"reportPropertyTypeMismatch"` in Value. `"reportFunctionMemberAccess"` was not in the documentation, but seems to fit there.


Bwerth048 (approved):

gregvanl (approved):
Thanks!

PR #5900, opened 2023-01-01
c-cpp-properties-schema-reference.md: configurationProvider: mention vscode-cpptools-api

### LINKED ISSUE #5816: Document the interface for an extension to be used as a `configurationProvider`
The current documentation only state that the `configurationProvider` attribute in `c_cpp_properties.json` to be "The ID of a VS Code extension that can provide IntelliSense configuration information for source files", and there's no way for a user to check if an extension supports that, nor can an extension developer get any information about the interface.

https://github.com/microsoft/vscode-docs/blob/eb97c3dc6fa7e6de8bbec096aaecc75b91e3f6a4/docs/cpp/c-cpp-properties-schema-reference.md?plain=1#L94-L95

Is it implemented through certain "commands" the target extension provides? It would be great to see it document.

ShamrockLee:
This PR enhances the documentation about the `configurationProvider` option specified inside `c-cpp-properties.json` by introducing the requirement for an extension to become a `configurationProvider` candidate.

Not sure how to name https://github.com/microsoft/vscode-cpptools-api properly, so I just call it vscode-cpptools-api.

Fixes #5816

### DIFF
--- a/docs/cpp/c-cpp-properties-schema-reference.md
+++ b/docs/cpp/c-cpp-properties-schema-reference.md
@@ -94,6 +94,8 @@ For more information about changing these settings, see [Customizing Default Set
 - `configurationProvider`
   The ID of a VS Code extension that can provide IntelliSense configuration information for source files. For example, use the VS Code extension ID `ms-vscode.cmake-tools` to provide configuration information from the CMake Tools extension.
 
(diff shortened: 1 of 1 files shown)


gregvanl (approved):

gregvanl:
Thanks!

PR #5902, opened 2023-01-03
Add when clause view IDs

gregvanl:
Fixes https://github.com/microsoft/vscode/issues/166647

Also call out the difference between view containers (viewlets) and views for the when clause conditionals.


Tyriar (approved):

sandy081:
Apologies for getting to it late. I have some suggestions and created a separate PR for those -  https://github.com/microsoft/vscode-docs/pull/5913

PR #5903, opened 2023-01-04
Update C/C++ extension FAQ.

sean-mcmanus:
Part of https://github.com/microsoft/vscode-cpptools/issues/10235 .

### DIFF
--- a/docs/cpp/faq-cpp.md
+++ b/docs/cpp/faq-cpp.md
@@ -9,11 +9,11 @@ MetaDescription: Frequently asked questions about the C/C++ extension in Visual 
 ---
 # Frequently asked questions
 
-- [Why are my files corrupted on format?](#why-are-my-files-corrupted-on-format)
 - [How do I get IntelliSense to work correctly?](#how-do-i-get-intellisense-to-work-correctly)
 - [Why do I see red squiggles under Standard Library types?](#why-do-i-see-red-squiggles-under-standard-library-types)
 - [How do I get the new IntelliSense to work with MinGW on Windows?](#how-do-i-get-the-new-intellisense-to-work-with-mingw-on-windows)
 - [How do I get the new IntelliSense to work with the Windows Subsystem for Linux?](#how-do-i-get-the-new-intellisense-to-work-with-the-windows-subsystem-for-linux)
+- [Why are my files corrupted on format?](#why-are-my-files-corrupted-on-format)
 - [What is the difference between includePath and browse.path in c_cpp_properties.json?](#what-is-the-difference-between-includepath-and-browsepath)
 - [How do I recreate the IntelliSense database?](#how-do-i-recreate-the-intellisense-database)
 - [What is the ipch folder?](#what-is-the-ipch-folder)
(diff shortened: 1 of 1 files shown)


sean-mcmanus:
Let me know if anyone doesn't understand why I made a particular change and I can explain.

sean-mcmanus:
@WardenGnaw Is there anything in the Debugger section that should be updated?

gregvanl (approved):
Thanks!

docs/cpp/faq-cpp.md
```diff
@@ -9,11 +9,11 @@ MetaDescription: Frequently asked questions about the C/C++ extension in Visual
 ---
```
AlexandraKemperMS, line 9: Could you update the Date so we know we already have done the freshness pass?

docs/cpp/faq-cpp.md
```diff
@@ -9,11 +9,11 @@ MetaDescription: Frequently asked questions about the C/C++ extension in Visual
...
 # Frequently asked questions
 
-- [Why are my files corrupted on format?](#why-are-my-files-corrupted-on-format)
 - [How do I get IntelliSense to work correctly?](#how-do-i-get-intellisense-to-work-correctly)
 - [Why do I see red squiggles under Standard Library types?](#why-do-i-see-red-squiggles-under-standard-library-types)
 - [How do I get the new IntelliSense to work with MinGW on Windows?](#how-do-i-get-the-new-intellisense-to-work-with-mingw-on-windows)
 - [How do I get the new IntelliSense to work with the Windows Subsystem for Linux?](#how-do-i-get-the-new-intellisense-to-work-with-the-windows-subsystem-for-linux)
+- [Why are my files corrupted on format?](#why-are-my-files-corrupted-on-format)
```
AlexandraKemperMS, line 16: Why did you decide to move this? I am going to see which ones of these links are clicked the most, so we could sort it most to least clicked, but I if you have another specific reason let me know.

sean-mcmanus: @AlexandraKemperMS I moved it because it was strange to be the 1st question since it's actually not a common issue that people frequently ask, i.e. the IntelliSense issues are more frequently asked.

docs/cpp/faq-cpp.md
sean-mcmanus: Okay, I wasn't sure who is supposed to update that date.

AlexandraKemperMS: No problem, as far as I am aware it can be anyone on the team, does not have to be Greg or Tyler, so you are good to go.

docs/cpp/faq-cpp.md
AlexandraKemperMS: That sounds good - my instinct is also that people ask around IS etc. more often. We can leave this for now and if I find anything in the data (migth take a bit to find) we can potentially switch the order in a future PR.

docs/cpp/cpp-ide.md
```diff
@@ -4,7 +4,7 @@ Area: cpp
 TOCTitle: Editing
 ContentId: 61D63E54-67E2-4743-B5CB-C6E7F582982A
 PageTitle: Edit and navigate C++ code in Visual Studio Code
-DateApproved: 7/25/2019
+DateApproved: 1/17/2023
```
gregvanl, line 7: Was the wrong file DateApproved updated?
If cpp-ide.md has been reviewed and doesn't need an update, you can just bump the DateApproved

sean-mcmanus: oh, sorry about that...

docs/cpp/faq-cpp.md
```diff
@@ -139,6 +133,6 @@ If your debugger is showing a grey stack trace, won't stop at a breakpoint, or t
 
 ## What do I do if I suspect a C/C++ extension problem
 
-If you are experiencing a problem with the extension that we can't diagnose based on information in your issue report, we might ask you to enable logging and send us your logs. See [C/C++ extension logging](/docs/cpp/enable-logging-cpp.md) for how to get C/C++ extension logs.
+If you have any other questions, please start a discussion at [GitHub discussions](https://github.com/microsoft/vscode-cpptools/discussions), or if you believe your issue may require a fix, file an issue at [GitHub issues](https://github.com/microsoft/vscode-cpptools/issues).
```
AlexandraKemperMS, line 136: it would be better to rephrase
"or if you believe your issue may require a fix", file an issue at" -> "or for any issues, file an issue at"

sean-mcmanus: I don't understand. "for any issues" seems too vague since their issue could just be a question such as how to get their IntelliSense working. The current wording differentiates between issues that are questions that are not expected to require a fix (best handled in "discussinos") and issues that require a fix (best handled in "issues").

sean-mcmanus: Did you have another alternative wording?

AlexandraKemperMS: The format I wrote the fix in might be confusing - I was looking to make the phrasing more clear because to a user, any issue they find they will believe will require a fix, so using the word "believe" is redundant. I agree that we should mention the discussions first if they are confused on how to use the product, and issues second. Here is the full sentence of what I meant:
"If you have any other questions, please start a discussion at [GitHub discussions](https://github.com/microsoft/vscode-cpptools/discussions), or if you find an issue, file an issue at [GitHub issues](https://github.com/microsoft/vscode-cpptools/issues)."

sean-mcmanus: Ok, that sounds good to me.

gregvanl (approved):
Thanks!

PR #5918, opened 2023-01-06
It's on keypress message with javascript

Ismailirfan:


Ismailirfan (commented):
It's my javascript on keydown events with sent message. I'm working on it

Issue #5738, opened 2022-10-26
Documentation states variable substitution is both supported and not supported in environment definition files


isaacwaldron:
Variable substitution including environment variables appears to be documented as supported in [environments.md on line 251](https://github.com/microsoft/vscode-docs/blob/68a719a8c561861006ce8cb8ba064ecd3a7dabef/docs/python/environments.md#L251).  However, there is a note on [line 219](https://github.com/microsoft/vscode-docs/blob/68a719a8c561861006ce8cb8ba064ecd3a7dabef/docs/python/environments.md#L219) that indicates this is not supported.  The actual behavior appears to be that it is not supported.

luabud:
Thanks for bringing this up -- you're correct, we currently do not support it. We need to update the variable substitution section accordingly.

PR #5925, opened 2023-01-06
Clarify variable substitution is not supported in .env files

luabud:
For https://github.com/microsoft/vscode-docs/issues/5738


gregvanl (approved):

karrtikr (commented):
Variable substitution is actually supported, atleast partially: https://github.com/microsoft/vscode-python/blob/26b614b739745a620bec661e4e48b1d27392ea48/src/client/common/variables/environment.ts#L158

So if it's not working it is a bug: https://github.com/microsoft/vscode-python/issues/20211. Should we temporarily remove this until that is fixed?

luabud:
Yeah I think so. We do have this issue open to track it: https://github.com/microsoft/vscode-python/issues/14950

PR #5934, opened 2023-01-10
Document systemd support in WSL (microsoft/vscode-remote-release#7821)

chrmarti:


bamurtaugh (approved):
Looks great, thanks Christof!

PR #5937, opened 2023-01-11
Updates Mac menu path for Settings

shawnthompson:
Preferences is now titled Settings...

### DIFF
--- a/docs/getstarted/settings.md
+++ b/docs/getstarted/settings.md
@@ -25,7 +25,7 @@ To modify user settings, you'll use the Settings editor to review and change VS 
 To open the Settings editor, use the following VS Code menu command:
 
 * On Windows/Linux - **File** > **Preferences** > **Settings**
-* On macOS - **Code** > **Preferences** > **Settings**
+* On macOS - **Code** > **Settings...** > **Settings**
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @shawnthompson Thanks for pointing this out. What version of macOS are you using?
I'm on Monterey and I still see Code > Preferences.
I wonder if this change started with the Ventura update?

shawnthompson:
Hi @gregvanl, 

I am using Ventura 13.1 so it much be something that changed in this version of macOS.

Now that I think of it, I heard that Apple changed **preferences** to **settings** to make is more similar to wording in iOS.

Would this mean you'd probably have to a note in your documentation for the different versions of macOS?

gregvanl:
Yep, there are about 35 instances in the docs where we call out **Code** > **Preferences**.
I'm reluctant to make the changes quite yet, given that the majority of Mac users won't be on Ventura for some time and it will get unwieldy to put both menu paths.
We could do something like **Code** > **Preferences/Settings** but that may be confusing

gregvanl:
Closing this PR as we've implemented platform and version detection on code.visualstudio.com so that **File** > **Preferences** is rendered appropriately for readers on macOS, including those on version Ventura (13) and above.

Window/Linux: **File** > **Preferences** > blah
macOS: **Code** > **Preferences** > blah
macOS (13+): **Code** > **Settings** > blah

PR #5949, opened 2023-01-19
Update Rust Linting Check on Save Command

archcorsair:
# Summary
- The old command: `rust-analyzer.checkOnSave.command` is no longer a valid command with the current version of the rust-analyzer vscode extension.
- To achieve the desired result without a invalid type error the command used must be: `rust-analyzer.check.command`


gregvanl:
Hi @archcorsair, thank you for spotting the setting change. 
Can you also update the setting name text along with the setting ID? It is now called Rust-analyzer > Check: Command.

image

archcorsair:
@gregvanl Good catch, fixed 👍 I updated the text, when you mention the setting ID is there another change required?

gregvanl:
Hi @archcorsair I'm not too familiar with the rust-analyzer settings so not sure if there were other setting changes. 
For example, there is a setting to enable/disable check on save `rust-analyzer.checkOnSave` (default true), which uses `rust-analyzer.check.command`. I don't know if this is a new setting or was there before. Unfortunately, I don't see a CHANGELOG for rust-analyzer in their repo.
You could add a mention of `rust-analyzer.checkOnSave` to that section in case people want to disable check on save.

archcorsair:
@gregvanl Had a peek around and it appears that `rust-analyzer.checkOnSave` is set to `true` by default for the default linting via `check` (instead of `clippy`). It does not appear that any other configurations needs to be adjusted. It might not hurt to add a statement to verify that `rust-analyzer.checkOnSave` is checked in the settings UI or set to `true` in the `settings.json` file.

gregvanl (approved):

PR #5978, opened 2023-01-25
added the extension remove steps

prashantvc:
Added the instruction to remove/unpublish an extension from the marketplace


gregvanl:
Hi @prashantvc Thank you for submitting this update. 
To add images and .gifs to the vscode-docs repo, we ask that people enable Git LFS in their local repo so that large binary files are stored in Git LFS rather than directly in the vscode-docs repo.
You can read how to set up Git LFS https://github.com/microsoft/vscode-docs/blob/main/README.md#cloning
I'll create a new PR based on this one with your new images.

gregvanl:
Here's the new PR https://github.com/microsoft/vscode-docs/pull/5982 with some edits and using Git LFS.

You can always check if a PR will add an image or .gif to Git LFS by looking at the image with "View file" and checking the top banner:
image

PR #5984, opened 2023-01-26
Fix Indentation Rule

namelessGonbai:
I just realized my mistake about four years ago.(#2297)


rebornix (approved):

gregvanl:
Thanks for coming back to this!

PR #5990, opened 2023-01-27
Avoid gerund and shorten header

gregvanl:
This helps avoid a line break in the pages right navigation list and also recommended writing practice to avoid gerunds in headers


bamurtaugh (approved):
Thank you!

PR #5995, opened 2023-01-28
AI doc feedback

bamurtaugh:
Address feedback from @digitarald today:

- Reduce size of notification images
- Combine sign in and sign up sections


digitarald (approved):
🎉

PR #6007, opened 2023-02-01
Explain new terminal a11y features

### LINKED ISSUE #5948: describe terminal accessibility help and terminal accessibility mode
Navigation mode is currently the first mention in the docs for accessibility. We should move that down to be below these new things

Tyriar:
Fixes #5948

### DIFF
--- a/docs/editor/accessibility.md
+++ b/docs/editor/accessibility.md
@@ -160,7 +160,7 @@ Certain features such as folding and minimap (code overview) are disabled when i
 
 Terminal accessibility help can be accessed via `kb(workbench.action.terminal.showAccessibilityHelp)` which describes some useful tips when using a screen reader. One of the tips described is to use `kb(workbench.action.terminal.enterAccessibilityMode)` to access the terminal's buffer, this will enter the screen reader's browse mode automatically (depending on your screen reader) and provide an accessible view of the entire terminal's buffer.
 
-The legacy "navigation mode" allows navigating through terminal rows, press `kbstyle(F1)` to open the Command Palette and search for "terminal navigation mode" to find these command. This mode will be removed soon in favor of the new accessibility mode.
+The legacy "navigation mode" allows navigating through terminal rows, press `kbstyle(F1)` to open the Command Palette and search for "terminal navigation mode" to find these commands. This mode will be removed soon in favor of the new accessibility mode.
 
 ### Shell integration
 


docs/editor/accessibility.md
```diff
@@ -158,7 +158,9 @@ Certain features such as folding and minimap (code overview) are disabled when i
 
 ## Terminal accessibility
 
-Output in the Integrated Terminal can be navigated through by using the "navigation mode" commands available in the Command Palette (press `kbstyle(F1)` and search for "terminal navigation mode").
+Terminal accessibility help can be toggled via `kbstyle(Alt+F1)` which describes some useful tips when using a screen reader. One of the tips described is to use `kbstyle(Shift+Tab)` to access the terminal's buffer, this will enter the screen reader's browse mode automatically (depending on your screen reader) and provide an accessible view of the entire terminal's buffer.
```
meganrogge, line 161: maybe this is a bug, but terminal accessibility help cannot be toggled with that keybinding at the moment - only turned on.

meganrogge: I don't think that's a bug as the editor does the same and we say how it can be hidden (focus elsewhere or via escape)

meganrogge, line 161:
```suggestion
Terminal accessibility help can be accessed via `kbstyle(Alt+F1)` which describes some useful tips when using a screen reader. One of the tips described is to use `kbstyle(Shift+Tab)` to access the terminal's buffer, this will enter the screen reader's browse mode automatically (depending on your screen reader) and provide an accessible view of the entire terminal's buffer.
```

meganrogge (changes requested):
one small adjustment

docs/editor/accessibility.md
gregvanl: Prefer to use `kb(workbench.action.terminal.showAccessibilityHelp)` than hard-coding the keyboard shortcut so if there are platform-specific keybindings, users will see the appropriate keybindings.
In this case, all platforms are now "Alt+F1" but earlier in milestone Windows had been "Shift+F1"

gregvanl: Similar for `kb(workbench.action.terminal.enterAccessibilityMode)` even though now all platforms use "Shift+Tab" it could change in the future

docs/editor/accessibility.md
```diff
@@ -158,7 +158,9 @@ Certain features such as folding and minimap (code overview) are disabled when i
...
+
+The legacy "navigation mode" allows navigating through terminal rows, press `kbstyle(F1)` and search for "terminal navigation mode" to find these command. This mode will be removed soon in favor of the new accessibility mode.
```
gregvanl, line 163: "press `kbstyle(F1)` to open the Command Palette and search for ...."

docs/editor/accessibility.md
Tyriar: It's alt+f1 and shouldn't change, but yeah that's better :+1:

docs/editor/accessibility.md
```diff
@@ -158,7 +158,9 @@ Certain features such as folding and minimap (code overview) are disabled when i
 
 ## Terminal accessibility
 
-Output in the Integrated Terminal can be navigated through by using the "navigation mode" commands available in the Command Palette (press `kbstyle(F1)` and search for "terminal navigation mode").
+Terminal accessibility help can be accessed via `kb(workbench.action.terminal.showAccessibilityHelp)` which describes some useful tips when using a screen reader. One of the tips described is to use `kb(workbench.action.terminal.enterAccessibilityMode)` to access the terminal's buffer, this will enter the screen reader's browse mode automatically (depending on your screen reader) and provide an accessible view of the entire terminal's buffer.
+
+The legacy "navigation mode" allows navigating through terminal rows, press `kbstyle(F1)` to open the Command Palette and search for "terminal navigation mode" to find these command. This mode will be removed soon in favor of the new accessibility mode.
```
gregvanl, line 163: Oops, missed a small nit earlier: "to find these commands."

meganrogge (approved):

gregvanl (approved):
Thanks!

PR #6008, opened 2023-02-02
chore: rm chromium bug info from applocker section

deepak1556:
Remove link to upstream bug since it is not relevant to end users, we already have link VSCode issue that users can follow for updates.

/cc @bpasero


bpasero (approved):

gregvanl:
@deepak1556 Missed that this was targeting vnext. I'll merge to main and publish

deepak1556:
Thanks @gregvanl

PR #6022, opened 2023-02-05
Extension display name and publisher display name new restrictions

amit-kulkarni-in:
Updating the extension documentation to communicate the policy changes and mention new restrictions on the extension manifest file and verified domain. https://github.com/microsoft/vscode-discussions/discussions/433


gregvanl (approved):
Thanks!

PR #6023, opened 2023-02-06
Improve description of the `contributes.viewContainers` id property

NouemanKHAL:
Describing the `id` field as `the name of the new view container` is imprecise. Instead, it should be `the id of the new view container`.


gregvanl (approved):
Thanks for the clarification

PR #6062, opened 2023-02-22
Add small section about editing a profile.

laundmo:
Editing profiles by using the normal means of changing settings, modifying extensions etc might see obvious to some but still should be explained somewhere.

This PR is just a small section about this fact, and a note about editing the settings.json directly while switching profiles.


gregvanl (approved):
Thanks. 
The final version of this section may differ slightly after an edit pass.

isidorn:
@laundmo thank you for the PR and thanks @gregvanl for the review and for pinging me.

@sandy081 can you also review these changes please :) Since the `"Note:"` part feels like a bug that we can fix?

sandy081:
`"Note:"` makes sense to me. We do not want to auto close and reopen files on switching profiles. This can be unexpected.

PR #6071, opened 2023-02-23
Describe remote indicator position correctly

gjsjohnmurray:
Remote indicator is a LEFT end of Status Bar, not the right end.


gregvanl (approved):
Thanks!

PR #6078, opened 2023-02-24
Update remote indicator images

joyceerhl:
Based on endgame feedback I made a change to the remote indicator to only show the `Install Additional Remote Extensions` command if there are additional recommended remote extensions to be installed. 

In our docs screenshots we always show all 6 extensions' commands, so I'm updating them again to reflect the latest (one less item in the remote indicator).


bamurtaugh (approved):
LGTM!

PR #6083, opened 2023-02-26
Replace python with py for windows

ForsakenRei:
Found the upper one updated one but this one is not.


luabud (approved):
Thanks for the PR!

PR #6084, opened 2023-02-27
Profiles topic

### LINKED ISSUE #6020: Profiles documentation
Profiles was released to stable in the 1.75 release.
Should either have a new topic (under User Guide) or add a section to the Settings topic explaining Profiles

gregvanl:
Fixes #6020

### DIFF
--- a/docs/editor/command-line.md
+++ b/docs/editor/command-line.md
@@ -1,5 +1,5 @@
 ---
-Order: 18
+Order: 19
 Area: editor
 TOCTitle: Command Line Interface
 ContentId: 8faef870-7a5f-4070-ad17-8ba791006912
--- a/docs/editor/command-line.md
+++ b/docs/editor/command-line.md
@@ -73,6 +73,14 @@ Argument|Description
 
 ![go to line and column](images/command-line/goto-line-column.png)
 
+## Select a profile
+
+You can launch VS Code with a specific [profile](/docs/editor/profiles.md) via the `--profile` command-line interface option. You pass the name of the profile after the `--profile` argument and open a folder or a workspace using that profile. The command line below opens the `web-sample` folder with the "Web Development" profile:
+
+`code ~/projects/web-sample --profile "Web Development"`
+
+If the profile specified does not exist, a new empty profile with the given name is created.
+
 ## Working with extensions
 
 You can install and manage VS Code [extensions](/docs/editor/extension-marketplace.md) from the command line.
--- a/docs/editor/profiles.md
+++ b/docs/editor/profiles.md
@@ -98,28 +98,15 @@ Profiles view will show current profile details (with Export) and imported profi
 
 ## Uses for Profiles
 
-todo@isidorn
+Profiles are a great way to customize VS Code to better fit your needs. In this section we'll look at some common use cases for profiles.
 
-### Programming Languages
+Since profiles are rememberd per workspace they are a great way to customize VS Code for a specific programming language. For example, you can create a JavaScript frontend profile that includes the extensions, settings and customisations you use for JavaScript development in one workspace, and have a Python backend profile that includes the extensions, settings and customisations you use for Python development active in another workspace. Using this approach you can easily switch between workspaces and always have VS Code configured the right way.
 
-Per workspace
-Also web development (front end versus back end)
-
-### Demos
-
-Simple without extra extensions or customizations
-Back to a default layout
-Also good for documentation screenshots
-Can use larger font size for presenting
+When doing a demo, you can use a profile to set up a specific configuration for your demo. For example, you can create a profile with a specific set of extensions and settings like zoom level, font size, and color theme. By doing this, a demo will not mess up your normal VS Code setup.
 
 ### Education
 
-Simplified layout and settings (autosave enabled) and just the extensions needed for a class
-
-### Testing
-
-Set up a specific repro or configuration without modifying your work setup
-"Oops" profile to put your setup back to a known state
+Profiles can be used to customise VS Code for students to ease the use in a classroom setting. Profiles allow educators to quickly share a customised VS Code setup with students. For example, educators can create a profile with a specific set of extensions and settings needed for a computer science class and then share that profile with students.
 
 ## Command line
 
--- a/docs/editor/profiles.md
+++ b/docs/editor/profiles.md
@@ -13,13 +13,13 @@ Visual Studio Code has 100s of [settings](/docs/getstarted/settings.md), 1000s o
 
 ## Create a Profile
 
-VS Code treats your current configuration as the **Default Profile**. As you modify settings, install extensions, or open and close views and panels, these customizations are tracked in the Default Profile.
+VS Code treats your current configuration as the **Default Profile**. As you modify settings, install extensions, or change UI layout by moving views, these customizations are tracked in the Default Profile.
 
-To create a new profile, you can use the **File** > **Preferences** > **Profiles** > **Create Profile** menu item (**Code** > > **Preferences** > **Profiles** > **Create Profile** on macOS).
+To create a new profile, you can use the **File** > **Preferences** > **Profiles** > **Create Profile** menu item (**Code** > **Preferences** > **Profiles** > **Create Profile** on macOS).
 
 ![Create Profile command](images/profiles/create-profile.png)
 
-You can create a new profile based on the current profile (**Profiles: Create from Current Profiles**) or create an Empty Profile. An Empty Profile includes no user customizations (settings, extensions, snippets, etc.) but with the UI state (view layouts and sizing) preserved.
+You can create a new profile based on the current profile (**Profiles: Create from Current Profiles**) or create an Empty Profile. An Empty Profile includes no user customizations (settings, extensions, snippets, etc.).
 
 You can also access the Profile command menu via the **Manage** gear button in the lower right of the Activity bar.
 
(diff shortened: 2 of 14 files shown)


docs/editor/command-line.md
```diff
@@ -73,6 +73,14 @@ Argument|Description
 
 ![go to line and column](images/command-line/goto-line-column.png)
 
+## Select a profile
```
isidorn, line 76: Great that we remembered to also add this!

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+---
+# Profiles in Visual Studio Code
+
+Visual Studio Code has 100s of [settings](/docs/getstarted/settings.md), 1000s of [extensions](/docs/editor/extension-marketplace.md), and unnumerable ways to adjust the UI layout to customize the editor. VS Code **Profiles** let you create sets of customizations and quickly switch between them or share them with others. This topic explains how to create, modify, export, and import profiles.
+
+## Create a Profile
+
+VS Code treats your current configuration as the **Default Profile**. As you modify settings, install extensions, or open and close views and panels, these customizations are tracked in the Default Profile.
+
+To create a new profile, you can use the **File** > **Preferences** > **Profiles** > **Create Profile** menu item (**Code** > > **Preferences** > **Profiles** > **Create Profile** on macOS).
```
isidorn, line 18: Is `**Code** > >` a typo? Should there be just one `>`

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+* Title bar
+* **File** > **Preferences** > **Profiles**
+* **Manage** gear button hover
+
+If you are still using the Default Profile, no profile name is displayed.
+
+The **Manage** gear button displays a badge with the first two letters of the active profile so you can quickly check which profile you are running.
+
+![Manage gear displaying "DE' to indicate that the user's 'Demo' profile is active](images/profiles/profile-gear-two-letters.png)
```
isidorn, line 44: The `DE` short name can easily be confused with the `DEfeault` profile or `DEutsch`. So let's use a different profile name.

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Edit a profile
+
+You can edit a profile just as you would normally change any VS Code configuration. You can install/uninstall/disable extensions, change settings, and adjust snippets like normal. These changes are stored in your currently active profile.
+
+### Workspace associations
+
+When you select create or select an existing profile, it is associated with the current workspace and whenever you open that folder, the workspace's profile is active. If you open another folder, the profile either goes back to the Default Profile if no profile is set, or switches to that folder's associated profile.
```
isidorn, line 52: "If you open another folder, the profile either goes back to the Default Profile if no profile is set"

This is not correct, we will use the profile that you last used in the previous folder. Right? @sandy081

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+Setting a workspace's profile back to the Default Profile clears the workspace's profile association.
+
+## Managing profiles
+
+### Switch profiles
+
+You can quickly switch between profiles with the **Profiles: Switch Profile** command in the Command Palette, which presents a dropdown listing your available profiles.
```
isidorn, line 60: Should we mention here "through the profiles menu or ...." ?

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+The **Profiles: Show Contents** command (available in the Command Palette or Profiles menus) brings up the Profiles view, where you can review the customizations for the profile.
+
+![Profile view displaying contents of a Doc Work profile](images/profiles/profile-show-contents.png)
+
+A profile can include:
+
+* Settings - In a profile-specific `settings.json` file.
+* Extensions - You can remove extensions from the profile by unchecking their entry in the Profiles view.
+* UI state - Currently open views and their positioning.
```
isidorn, line 80: UI state is a bit more than that. @sandy081 do you have good ideas how to generalise this sentence?
Maybe "Currently open views, their positioning and any other layout state."

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+* Keybindings - In a profile-specific `keybindings.json` file.
+* Snippets - In a profile-specific `{language}.json` files.
+* User Tasks - In a profile-specific `tasks.json` file.
+
+When you create a new profile based on the Default Profile, the profile-specific configuration files are populated from your user configuration files. Workspace-specific settings are not automatically included in a new profile.
```
isidorn, line 85: Here we should mention what happens if you do not want something to be a part of a profile. In this case, we actually use the content from the default profile.
Example: if you import a profile, and uncheck the keybindnigs, the profile you import will instead use keybindings from the default profile.

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Share Profiles
+
+### Export
+
+You can export a profile in order to save it or share it with others. The **Export Profile** command displays the Profiles view with the contents of the active profile and an **Export** button. You can unselect various elements of the profile such as extensions or configuration files before you export the profile.
+
+When you select **Export**, you are prompted for the profile name and whether you want to export to a [GitHub gist](https://docs.github.com/get-started/writing-on-github/editing-and-sharing-content-with-gists/creating-gists) or your local file system.
+
+#### Save as a GitHub gist
```
isidorn, line 95: Do we use #### in our docs? This does not get rendered nicely in vscode. Should we at least bold this small title for clarity?

docs/editor/profiles.md
isidorn: We can also mention this in the import flow. Whatever you think works better.

isidorn (changes requested):

isidorn:
fyi @sandy081 @burkeholland

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+DateApproved: 2/2/2023
+MetaDescription: Expand your development workflow with task integration in Visual Studio Code.
+---
+# Profiles in Visual Studio Code
+
+Visual Studio Code has 100s of [settings](/docs/getstarted/settings.md), 1000s of [extensions](/docs/editor/extension-marketplace.md), and unnumerable ways to adjust the UI layout to customize the editor. VS Code **Profiles** let you create sets of customizations and quickly switch between them or share them with others. This topic explains how to create, modify, export, and import profiles.
+
+## Create a Profile
+
+VS Code treats your current configuration as the **Default Profile**. As you modify settings, install extensions, or open and close views and panels, these customizations are tracked in the Default Profile.
```
sandy081, line 16: I would not say open and close views and panels instead say change views layout like moving views. Because opening and closing views is not part of the profile.

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+To create a new profile, you can use the **File** > **Preferences** > **Profiles** > **Create Profile** menu item (**Code** > > **Preferences** > **Profiles** > **Create Profile** on macOS).
+
+![Create Profile command](images/profiles/create-profile.png)
+
+You can create a new profile based on the current profile (**Profiles: Create from Current Profiles**) or create an Empty Profile. An Empty Profile includes no user customizations (settings, extensions, snippets, etc.) but with the UI state (view layouts and sizing) preserved.
```
sandy081, line 22: UI state is also not included in empty profile. For eg all my view layouts will be reset to default. There is some workspace specific UI state which is not preserved like views size, views opening and closing.

docs/editor/profiles.md
```diff
@@ -0,0 +1,169 @@
...
+
+If you are still using the Default Profile, no profile name is displayed.
+
+The **Manage** gear button displays a badge with the first two letters of the active profile so you can quickly check which profile you are running.
+
+![Manage gear displaying "DE' to indicate that the user's 'Demo' profile is active](images/profiles/profile-gear-two-letters.png)
+
+### Edit a profile
+
+You can edit a profile just as you would normally change any VS Code configuration. You can install/uninstall/disable extensions, change settings, and adjust snippets like normal. These changes are stored in your currently active profile.
```
sandy081, line 48: Also mention about UI layout customizations like moving views around, say moving view to secondary side bar, hiding views, view containers or actions.

docs/editor/profiles.md
sandy081: Yes, thats correct. A newly opened folders takes the profile of the previous window.

docs/editor/profiles.md
sandy081: UI Layout includes - Views layout (positions), visibility of views and actions.

sandy081 (changes requested):

docs/editor/profiles.md
gregvanl: Hmm, sometimes works that way but not if you've explicitly switched back to the Default Profile, which I thought cleared the profile association.
I'll update the topic and play around with this a bit more

docs/editor/profiles.md
gregvanl: I figured people had already discovered the profile list in the profile menus and the check UI. This section is for discovery of the quick switching command.

docs/editor/profiles.md
gregvanl: Can users hide actions?

docs/editor/profiles.md
gregvanl: We recently tuned the H4 styling on the website to be a reasonable font size and bold. Here's what it will look like live:

image

docs/editor/profiles.md
isidorn: Makes sense. Should we then say "You can also switch"?

docs/editor/profiles.md
isidorn: Yes.

docs/editor/profiles.md
gregvanl: Hmm, this is subtle and somewhat confusing behavior. I don't see this behavior for example with extensions. If I have an extension in a profile and I unselect the extension before exporting, when I import I don't see my Default Profile extensions.
Let's leave this out for now and we can iterate on the best way to describe this behavior.

docs/editor/profiles.md
gregvanl: Updated image to use "Primary Works" which shows as "PR"

docs/editor/profiles.md
gregvanl: Added a paragraph

sandy081 (approved):

isidorn (approved):

PR #6091, opened 2023-02-28
Update images for Azure Container Apps

ghogen:


gregvanl:
Hi @ghogen Since you're in the file, do you think the H1 title should be shortened to avoid the line break:

image

Maybe it could just be "Deploy to Azure" or "Deploy a container image to Azure"?

You could also think about shortening the H2 "Deploy the image to Azure App Service or Azure Container Apps", which has a line break and also a three link right navigation link.

image

Maybe shorten to "Deploy to Azure App Service or Azure Container Apps"?

ghogen:
@gregvanl Ready to merge.

gregvanl (approved):

PR #6097, opened 2023-03-01
Update codebasics.md

impressivewebs:
Added "CSS" to the list of default formatters under the "Formatting" heading.


gregvanl (approved):
Thanks!

impressivewebs:
@gregvanl No problem. FYI - I don't know if there are other languages with built in formatting, but that was the only one I noticed was missing from that page in the docs.

gregvanl:
I think that covers it since LESS and SCSS are just variants of CSS.

image

PR #6103, opened 2023-03-01
Add clarification for debugger extension per OS

modkaffes:
While following the tutorial on macOS, I installed the first listed extension and had to look up why the debugger wasn’t working (giving the "MIDebuggerPath" error). A small addition like that could probably help other noobs in compiled languages like me :)


gregvanl (approved):
Thanks! Good clarification

Issue #6094, opened 2023-02-28
Make path to dev container more obvious in documentation


chrmarti:
The "Developing inside a Container" page misses a short path to the template picker and a running dev container:
- The "devcontainer.json file" link in the first paragraph of the page links to section that suggests manually adding a devcontainer.json.
- The "Quick start: Open an existing folder in a container" is further down.

Could we start the page with a paragraph that skips the details and tries to direct the user to the right "Quick start" section below? E.g., with wording like: "If you already have Docker and the Dev Containers extension installed, you can skip ahead to `try a sample`, `open an existing folder in a container` or `clone a repository in a Docker volume`.

Could we add the template picker to the "Create a devcontainer.json file" section? Alternatively we could mention it and link to the "Quick start" section showing the template picker, but there is a chance that readers will skim the text and miss that link.

/cc @hediet who provided feedback, thanks!

PR #6105, opened 2023-03-01
Update dev container doc based on feedback

bamurtaugh:
Fixes https://github.com/microsoft/vscode-docs/issues/6094.

@chrmarti @hediet thanks for sharing this feedback, it makes a lot of sense to me too. 

I agree just listing 3 quick starts may be confusing or overwhelming, i.e. which do I pick? Do I need to complete them all in order? I added a `## Picking your quick start` section at the top to provide further context. I also numbered each quick start as that feels a bit easier to parse (I can quickly see there are 3 quick starts, rather than not knowing if there are 1, 5, 10, or more).

I removed the [Working with Containers](https://code.visualstudio.com/docs/devcontainers/containers#_working-with-containers) header as it felt confusing to have this so early and feels like it may direct users to leave the page. I moved its couple of sentences to the top of the doc in the existing top intro section.

I also added a screenshot of the template picker to the "Create a devcontainer.json file" section, and reordered that section so that the template picker comes first, rather than it appearing that users should be hand creating their devcontainer.json.

Let me know if you have other feedback!

cc @Chuxel

### DIFF
--- a/docs/devcontainers/containers.md
+++ b/docs/devcontainers/containers.md
@@ -17,6 +17,11 @@ Workspace files are mounted from the local file system or copied or cloned into 
 
 This lets VS Code provide a **local-quality development experience** including full IntelliSense (completions), code navigation, and debugging **regardless of where your tools (or code) are located**.
 
+The Dev Containers extension supports two primary operating models:
+
+* You can use a container as your full-time development environment
+* You can [attach to a running container](/docs/devcontainers/attach-container.md) to inspect it.
+
 ## Getting started
 
 **Note**: You can learn how to get up-and-running quickly with dev containers in the introductory [Dev Containers tutorial](/docs/devcontainers/tutorial.md).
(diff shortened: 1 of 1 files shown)


Chuxel (approved):

bamurtaugh:
@chrmarti @hediet please just let me know if you have any feedback, thanks so much!

chrmarti (approved):
I wonder if it would make sense to move the architecture diagram further down because that is not what potential new users are looking for, they are usually more interested in the Quick Starts and the prerequisites. Maybe something to think about first.

LGTM!

bamurtaugh:
That's great feedback @chrmarti - I'll make a note of that since it's how we generally structure our remote docs.

@gregvanl this is ready for your review and merge whenever you have a chance, thanks!

gregvanl (approved):
Looks good

PR #6106, opened 2023-03-02
Add identifier of Docker Compose into identifiers.md

GOUKOU007:


gregvanl:
Hi @GOUKOU007 Thanks for noticing that the `dockercompose` identifier is missing from the document. The friendly name displayed in the Status bar indicator and the Set Language Mode drop down is "Compose" for the `dockercompose` identifier.

image

Can you use "Compose" and move the entry up the list into the C's so the identifier list remains alphabetically sorted?

gregvanl:
Maybe you could also move "CUDA C++" down below "CSS" to be better sorted?

GOUKOU007:
Of course, I've commit the corresponding modification.

gregvanl (approved):
Thanks!

PR #6110, opened 2023-03-03
Update FAQ for Win 8

ChaseKnowlden:
Electron 23 will not support Windows 8 and 8.1.


gregvanl:
Hi @ChaseKnowlden Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
Closing this PR for now as it is probably too early to call this out in the FAQ as VS Code is not yet on Electron 22.
We should wait until the corresponding deprecation changes to make into VS Code core and then we can either reopen this PR or create a new one.

PR #6005, opened 2023-01-31
Fix and update the doc for java-webapp to the latest

silenceJialuo:
This is Jialuo Gan here and I am the PM at Microsoft for Java on Azure Tooling. I have fixed all the issues to keep it up to date. Please wait for Java 17 runtime issues to be fixed and then merge this doc to the main. I will update it later.

PR #6112, opened 2023-03-06
Fix and update the doc for java-webapp to the latest (#6005)

sulliven9:
Fix #6005


docs/remote/vscode-server.md
```diff
@@ -0,0 +1,125 @@
+---
+Order: 6
```
Stellaspada, line 2: 9

PR #6125, opened 2023-03-08
Fix typo in task-provider.md

adriancostin6:
changes  "it would be be better to for VS Code to call" into "it would be be better for VS Code to call" at line 79


gregvanl:
Hi @adriancostin6 Thanks for catching this and submitting a PR!

Would you also like to remove the duplicated "be" in "it would be be better..." in this PR?

adriancostin6:
> Hi @adriancostin6 Thanks for catching this and submitting a PR!
> 
> Would you also like to remove the duplicated "be" in "it would be be better..." in this PR?

it's done, can't believe I've missed the duplicate. maybe I was tired 😄

gregvanl (approved):
Thanks!

PR #6142, opened 2023-03-14
Remove sections from Python's "editing" page which do not make sense there

brettcannon:
Most have their own sections already.


gregvanl:
Hi @brettcannon Can you check for links to the H2 headers you are removing? I see several links from places like settings-reference.md to /docs/python/editing.md#formatting.
You can either remove those header links or update them to somewhere else.
Do you think anyone externally might have bookmarked or linked to those headers?

brettcannon:
Totally slipped my mind we don't have a separate "Formatting" page.

What I'm going to do is close this PR for now and think about what we want to do here and formatting support is being moved out to their own extensions.

PR #6150, opened 2023-03-19
adding missing file's type extension

groovy-sky:


gregvanl:
Hi @groovy-sky Thanks for your interest in the VS Code documentation and for taking the time to submit a PR.
Unfortunately, topics under /api don't have the same topic link preprocessing as topics under /docs so they can't use the .md file extension. If you search around, you'll notice that other topic links under /api don't have the .md file extension.
Closing this PR as it will result in broken topic links when rendered on code.visualstudio.com

PR #6154, opened 2023-03-22
reorganize when-clauses-doc

ulugbekna:


aiday-mar (approved):

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
                    "when": "debuggersAvailable && !inDebugMode" },
 ```
 
-Above the built-in **Start Debugging** command has the keyboard shortcut `kb(workbench.action.debug.start)`, which is only enabled when there is an appropriate debugger available (context `debuggersAvailable` is true) and the editor isn't in debug mode (context `inDebugMode` is false).
+Above, the built-in **Start Debugging** command has the keyboard shortcut `kb(workbench.action.debug.start)`, which is only enabled when there is an appropriate debugger available (context key `debuggersAvailable` is true) and the editor isn't in debug mode (context key `inDebugMode` is false).
 
-## Conditional operators
+## Composing when-clauses with operators
```
gregvanl, line 21: Changing this H2 header will break any header links. We can update our links in vscode-docs such as in https://code.visualstudio.com/docs/getstarted/keybindings#_conditional-operators but this will break any external links or bookmarks. 
Are you sure it is necessary to update this H2 header?

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
 
-For conditional expressions, you can use the following conditional operators:
+A when-clause can consist of a single context key, e.g., `inDebugMode`, or can use various operators to express more nuanced editor state.
```
gregvanl, line 23: Not sure I see the need to change "when clause" to "when-clause". Why do you think we need to add the hyphen?
Also we prefer to not use latin abbreviations, use "for example" instead

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
+
+_Note_: When-clauses, starting from VS Code version 1.77, can include parentheses. Extensions that use parentheses in when-clauses should be specified as compatible with VS Code version 1.77 and up only.
```
gregvanl, line 25: Prefer to use `> **Note**:` for indented Note sections

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
+<code>!foo \|\| bar </code>            | `(!foo) \|\| bar`
+<code>foo \|\| bar && baz </code>      | <code>foo \|\| (bar && baz)</code>
+<code>!foo && bar \|\| baz </code>     | <code>(!foo && bar) \|\| baz</code>
+<code>!(foo \|\| bar) && baz </code>   | <code>(remains same) !(foo \|\| bar) && baz</code>
 
-### key-value when clause operator
+### Equality operators
 
-There is a key-value pair match operator for `when` clauses. The expression `key =~ value` treats the right-hand side as a regular expression to match against the left-hand side. For example, to contribute context menu items for all Docker files, one could use:
+You can check for equality a context key's value in the current editor context against a value. Note that the right-hand side is a value and not interpreted as a context key, i.e., it is not looked up in the context.
```
gregvanl, line 49: Not sure this sentence reads correctly. 
Maybe "You can check for equality of a context key's value against a specified value."
Avoid Latin abbreviations so "Note that the right-hand side is a value and not interpreted as a context key, meaning it is not looked up in the context."

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
+
+Operator   | Symbol | Example
+--------   | ------ | -----
+Equality   | `==`   | `"editorLangId == typescript"` or `"editorLangId == 'typescript'"`
+Inequality | `!=`   | `"resourceExtname != .js"` or `"resourceExtname != '.js'"`
+
+Notes:
+
+- If the value on the right-hand side is a string containing whitespace, it must be wrapped in single-quotes, e.g., `"resourceFilename == 'My New File.md'"`
```
gregvanl, line 58: Avoid Latin abbreviations so "for example, ...."

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
+- `===` has the same behavior as `==`, and `!==` has the same behavior as `!=`
+
+### Comparison operators
+
+You can compare a context key's value against a number. Note that left- and right-hand side of the operator must be separated by whitespace, i.e., `foo < 1` not `foo<1`.
```
gregvanl, line 63: Avoid Latin abbreviations. I think you can just put the example in parentheses.

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
 
-## Available contexts
+Notes:
+
+- The right-hand side of the `=~` operator follows _almost_ the same rules as JavaScript regular expression literals ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#creating_a_regular_expression)). It is "almost the same rules" because when clauses are written within a JSON string, so the value needs to follow both encoding of JSON strings and regular expressions. For example, `/file:\/\/\//` should become `file:\\/\\/\\/`. In other words, if one wants to use a slash within their regular expression, they need to escape it with two backslash characters as in the example.
+- There does _not_ exist an operator `!=~`, but you can negate the match expression `!(foo =~ /baz/)`
+
+#### Regular expression flags
+
+It is possible to use flags with the regular expression literals, e.g., `resourceFilename =~ /json/i` or `myContextKey =~ /baz/si`.
```
gregvanl, line 91: for example

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,130 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
 
 ```json
    "when": "resourceFilename =~ /docker/"
 ```
 
-## Available contexts
+Notes:
+
+- The right-hand side of the `=~` operator follows _almost_ the same rules as JavaScript regular expression literals ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#creating_a_regular_expression)). It is "almost the same rules" because when clauses are written within a JSON string, so the value needs to follow both encoding of JSON strings and regular expressions. For example, `/file:\/\/\//` should become `file:\\/\\/\\/`. In other words, if one wants to use a slash within their regular expression, they need to escape it with two backslash characters as in the example.
+- There does _not_ exist an operator `!=~`, but you can negate the match expression `!(foo =~ /baz/)`
```
gregvanl, line 87: I don't think you need emphasis here

api/references/when-clause-contexts.md
```diff
@@ -171,108 +264,76 @@ focusedView | True when specific view is focused.<br>Example: `"focusedView == '
...
-* workbench.panel.markers.view - Problems
-* workbench.panel.output - Output
-* workbench.panel.repl.view - Debug Console
-* terminal - Integrated Terminal
-* workbench.panel.comments - Comments
+* `workbench.explorer.fileView` - File Explorer
+* `workbench.explorer.openEditorsView` - Open Editors
+* `outline` - Outline view
+* `timeline` - Timeline view
+* `workbench.scm - Source Contr`ol
```
gregvanl, line 271: Typo

api/references/when-clause-contexts.md
ulugbekna: Good point. How about we separate the exact header from the link target using this technique - https://developers.google.com/style/headings-targets#add-a-custom-anchor (the "Recommended" one) ?

gregvanl: We try to keep to strict Markdown syntax and only embed HTML tags in rare cases.
Also not sure how this would work with Matt's Markdown header link verification, which I think would still flag that link as broken

api/references/when-clause-contexts.md
gregvanl: Initially I thought using `"when" clause` might be a good idea since it matches what's in the JSON keybindings file but searching for `"when" clause`, I only found one instance in vscode-docs so I'm reluctant to go through all the docs for this.
I think leaving it as `when clause` is fine.

gregvanl: I do see that we sometimes use "`when` clause", using backtick for an inline code block styling but that is also rare and when we are talking about the attribute in a JSON example

api/references/when-clause-contexts.md
```diff
@@ -16,37 +16,128 @@ For example, VS Code uses when clauses to enable or disable command keybindings,
...
+There is a match operator (`=~`) for when clauses. The expression `key =~ regularExpressionLiteral` treats the right-hand side as a regular expression literal to match against the left-hand side. For example, to contribute context menu items for all Docker files, one could use:
 
 ```json
    "when": "resourceFilename =~ /docker/"
 ```
 
-## Available contexts
+Notes:
+
+- The right-hand side of the `=~` operator follows the same rules as JavaScript regular expression literals ([reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#creating_a_regular_expression)), except characters need to follow escaping rules both of JSON strings and regular expressions. Hence, using a slash within the regular expression requires double escaping. For example, what is written as `/file:\/\//` in JavaScript needs to become `/file:\\/\\//` when used within a when clause.
```
gregvanl, line 84: You can leave of the `/en-US` locale in the URL so that non-US non-English readers can hopefully see a localized version of the page.

ulugbekna:
@gregvanl thanks heaps for reviewing - very helpful! 

There's a bit of problem with merging it now: the PR contains 1.77 changes. 

Should we revert? 

Should have I created the PR against `vnext`?

gregvanl:
Oops, yes, vscode-docs vnext would have been better if the text calls out 1.77 features.
It's just the mention of using parentheses, correct? If so, then I think it's fine to be live a couple of days early

ulugbekna:
> It's just the mention of using parentheses, correct? If so, then I think it's fine to be live a couple of days early

parentheses, regex flags, and `===`/`!==` ops 

Shouldn't be too big of a deal since release is in a day or two, but I can send a revert PR if you'd like

> vscode-docs vnext would have been better if the text calls out 1.77 features.

thanks! I sent [6177](https://github.com/microsoft/vscode-docs/pull/6177) against vnext.

gregvanl:
I think it's fine to leave those updates live. 
I don't see examples of these yet in the Default Keybindings, which is probably where many people are learning about when clauses.

PR #6155, opened 2023-03-22
update terminal accessibility docs

meganrogge:


Tyriar (approved):
lgtm, just not sure about "**Command** (kb(command))" @gregvanl?

PR #6161, opened 2023-03-24
Remove unnecessary content about static files for tutorial-django

imba-tjd:
There is no need to mention `ALLOWED_HOSTS` at this point.

For *staticfiles_urlpatterns* and *Issue 13*, see https://github.com/microsoft/python-sample-vscode-django-tutorial/issues/13#issuecomment-1482231972
If you use django's runserver, you don't need them. If you set DEBUG=False, they take no effect anyway.

With that, most of them become unnecessary, even though <code>When switching to production, navigate to `settings.py`, set `DEBUG=False`</code> is correct.

### DIFF
--- a/docs/python/tutorial-django.md
+++ b/docs/python/tutorial-django.md
@@ -410,22 +410,6 @@ Static files are pieces of content that your web app returns as-is for certain r
 
 Serving static files in Django is something of an art, especially when deploying to production. What's shown here is a simple approach that works with the Django development server and also a production server like Gunicorn. A full treatment of static files, however, is beyond the scope of this tutorial, so for more information, see [Managing static files](https://docs.djangoproject.com/en/3.1/howto/static-files/) in the Django documentation.
 
-When switching to production, navigate to `settings.py`, set `DEBUG=False`, and change `ALLOWED_HOSTS = ['*']` to allow specific hosts. This may result in additional work when using containers. For details, see [Issue 13](https://github.com/microsoft/python-sample-vscode-django-tutorial/issues/13).
-
-### Ready the app for static files
-
-1. In the project's `web_project/urls.py`, add the following `import` statement:
-
-    ```python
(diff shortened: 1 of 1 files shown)


luabud:
Thanks for your suggestion! After talking to my colleague who has a ton of Django experience, we decided to leave this section in our docs giving it's not technically incorrect and provides info that could be helpful -- for example, it leads into more explanation about how allowed hosts works in production. ALLOWED_HOSTS is a field that is pre-populated by django-management command in a brand new project, and it needs to be altered in order to use it in production. 

If you would be open to help us improve the paragraph by explaining what ALLOWED_HOSTS are (for example with the paragraph below), we'd be more than happy to take a contribution!
Otherwise we can do that as a separate PR 😊

> ..., and change `ALLOWED_HOSTS = ['*']`, which uses a wildcard to allow any host, to a comma-separated list of specific hosts for a more secure deployment.

imba-tjd:
🆗How about staticfiles_urlpatterns()? Like I said, if you use `python manage.py runserver`, you don't need this. And if you set `DEBUG=False`, you also don't need this.

luabud:
Even though it may not be _needed_, I'm not convinced it's not helpful to leave it there for additional scenarios. @dawnwages -- would appreciate your guidance/opinions here!

dawnwages:
Agreed with @Luciana 

re: https://code.visualstudio.com/docs/python/tutorial-django#_serve-static-files

> When switching to production, navigate to settings.py, set DEBUG=False, and change ALLOWED_HOSTS = ['*'] to allow specific hosts. This may result in additional work when using containers. For details, see [Issue 13](https://github.com/microsoft/python-sample-vscode-django-tutorial/issues/13).
 
I would say keep it. This is the only point in the tutorial that ALLOWED_HOSTS is mentioned. It is not technically incorrect. It leads into more explanation about how allowed hosts works in production and ALLOWED_HOSTS is a field that is pre-populated by django-management command in a brand new project.

PR #6164, opened 2023-03-26
fix: Corrected to Ruby LSP

lolychee:


gregvanl (approved):
Thanks! 
I've got Git LFS on the brain

Issue #6141, opened 2023-03-14
Update Environments section in Python language doc


karrtikr:
https://code.visualstudio.com/docs/languages/python#_environments

- [x] Update screenshot to not use Anaconda, and show interpreters on right hand side instead
- [x] Update the Invalid interpreter status bar item
- [x] Correct debugging line here:
   ```
   The selected environment is used for IntelliSense, auto-completions, linting, formatting, and any other language-related feature other than debugging.
   ```
- [x] Maybe teach how to open command palette instead of screenshot?
- [x] Take section out on installing packages

PR #6176, opened 2023-03-27
Use correct link to images in Python environment doc

karrtikr:
For https://github.com/microsoft/vscode-docs/issues/6141


gregvanl (approved):
Thanks for tidying this up!

PR #6179, opened 2023-03-28
Add quotes around excludes value

rzhao271:
I'm not sure whether this will work, but I think it's worth trying.

This PR attempts to resolve some recent codeSignValidation issues that appeared in S360.


rzhao271:
Closing for now.

PR #6181, opened 2023-03-29
Update nodejs-debugging.md

imba-tjd:
I didn't use this settings before, but it seems that it's not default now

图片


gregvanl (approved):
Thanks for noticing this change!

PR #6188, opened 2023-03-30
give credit where credit is due

chrisdias:
Minor change to be more explicit about collaborating with GitHub on Chat View


gregvanl (approved):
Awesome!

PR #6215, opened 2023-04-10
Fix name of extension status bar

mromanch:


gregvanl:
Hi @mromanch Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
We generally refer to the Remote indicator to the left of the Status bar as the "Remote" Status bar item so we'll leave the tutorial text as is

mromanch:
Yes, I see it called "Remote Host" inside Visual Code! Specififcally, I had installed the extension several months ago, but got. sidetracked by real life. Along the way, I disabled the statusbar to get back some real estate. When I came back to this tutorial, I couldn't find the status bar, and had forgotten I had turned it off, and when looking in the possible options, "Remote Host" was available, but "Remote" was not. If I look at only the enabled extensions I have filtered by "remote", I find I have 9, so precission seems like a good idea.

PR #6240, opened 2023-04-21
Added example about how to use vscode to see changes after git fecth

vcrobe:


gregvanl:
Hi @vcrobe Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
The purpose of the section you want to update is just to call out various Git scenarios where VS Code can be used as the Git editor and not really a place for specific Git workflow steps.

PR #6241, opened 2023-04-23
Update path to snippets configuration

kamilturek:
The path to  "Configure User Snippets" button specified in docs was wrong for macOS.

image


kamilturek:
Ah, I see it's a duplicate of https://github.com/microsoft/vscode-docs/pull/5937. Closing then.

PR #6250, opened 2023-04-28
Run imageoptim on updates up to 1.60

digitarald:
0_3_0 to 1_60.

Since this didn't touch animated gifs, it only won about 188 Mb:

`Saved 187.7 MB out of 932.4 MB. 35.9% per file on average (up to 79.9%)`


gregvanl (approved):
The preview is too much for github.com

PR #6259, opened 2023-05-01
Join sshfs options

joseluisbf-kpsr:


gregvanl:
Hi @joseluisbf-kpsr Thank you for your interest in the VS Code documentation and for submitting a PR.
Is this change required for the command to work or just a shorthand preference? 
I'd prefer to leave the switches separated to make it clear that each switch is an option. 
Someone copying the command can tune command line as they wish for their own use

PR #6262, opened 2023-05-01
Fix vscode-docs pipeline triggers

rzhao271:
Renames some website docs pipelines for clarity and to avoid a name collision with the vscode-website-prod name.


.pipelines/onebranch.buddy.yml
```diff
@@ -19,7 +19,7 @@ resources:
       ref: refs/heads/main
 
   pipelines:
-    - pipeline: vscode-website-main
```
sbatten, line 22: I'm not sure I understand the point of this rename. the pipeline name is a nickname for the pipeline that we ingest. we are ingesting the vscode-website/main and vscode-website/release/prod so the names match their contents today.

rzhao271: vscode-website-main is fine as a name, but there is another pipeline named vscode-website-prod already in the website repo. I noticed that the website buddy build triggers the docs buddy build, but that the website official build doesn't trigger the docs official build.

.pipelines/onebranch.official.yml
```diff
@@ -19,11 +19,12 @@ resources:
       ref: refs/heads/main
 
   pipelines:
-    - pipeline: vscode-website-prod
+    - pipeline: vscode-website-docs-main
       source: vscode-website (official)
       trigger:
         branches:
-          - release/prod
+          - main
```
rzhao271, line 26: I need to revert this line

rzhao271:
It turns out `trigger > branches` filters by branches, whereas `branches` selects which branch to pull the artifact from. I wonder whether the trigger was waiting for something from `vscode-docs/release/prod` instead of `vscode-website`.

sbatten (approved):

PR #6266, opened 2023-05-02
Fixing typo with `Introduction to Git in VS Code` hint

SpencerKaiser:
The current version of the docs reference `Select Default Shell`, which does not exist


gregvanl (approved):
Thanks!

PR #6276, opened 2023-05-05
Update MSYS2 default installed package to it's recommendation.

wy-luke:
In MSYS2 official doc, the default installed package became `mingw-w64-ucrt-x86_64-gcc`. It says that `ucrt` is a better solution.

>UCRT (Universal C Runtime) is a newer version which is also used by Microsoft Visual Studio by default. It should work and behave as if the code was compiled with MSVC.

So I thought maybe it's better to change our doc according to this change.

I changed the `mingw-w64-x86_64-toolchain` to `mingw-w64-ucrt-x86_64-toolchain` and changed the path from `C:\msys64\mingw64\bin` to `C:\msys64\ucrt64\bin` and so on.

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -148,7 +148,7 @@ Your new `tasks.json` file should look similar to the JSON below:
         {
             "type": "cppbuild",
             "label": "C/C++: g++.exe build active file",
-            "command": "C:\\msys64\\mingw64\\bin\\g++.exe",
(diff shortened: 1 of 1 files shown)


gregvanl:
Adding `don't merge` label while the C++ extension engineering team has a review.

gregvanl:
Thank you again @wy-luke for this suggestion.
Now that the C++ extension supports UCRT, the team has submitted https://github.com/microsoft/vscode-docs/pull/6631 which covers this PR

PR #6300, opened 2023-05-11
Update remove keybinding instructions

Sophismata:
See https://github.com/microsoft/vscode/issues/181654.

Replaced empty string advice with `noop` and `^noop`.

Please feel free to make improvements.

### DIFF
--- a/docs/getstarted/keybindings.md
+++ b/docs/getstarted/keybindings.md
@@ -260,11 +260,16 @@ Here is an example:
 
 ```
 
-To override a specific key binding rule with an empty action, you can specify an empty command:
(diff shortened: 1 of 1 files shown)


Bysea0402 (approved):

gregvanl:
@Sophismata Thanks for this PR and your interest in the VS Code documentation.
Marking as `don't merge` since the fix will be available in the next 1.79 release.
Leaving this PR open until then in case anyone else needs the workaround.

gregvanl:
Correction: The fix went out in the 1.78.2 release so closing this PR now.

af4jm:
Agree with @Sophismata. The change in 1.78 had no accompanying documentation change, even with the rollback in 1.78.2, this change documents 2 distinct options for key disabling, where the existing documentation only offers 1 of the options, and doesn't specify which of the 2 it is.

So why would this be closed without merging???

@gregvanl

PR #6332, opened 2023-05-19
Enable a few more markdown language features for repo

mjbvz:
- Added `markdown.copyFiles.destination` so VS Code creates files dropped into release notes in expected path (requires VS Code 1.79)

- Turned on `organizeLinkDefinitions` on save. This removes unused link definitions and sorts the remaining definitions


gregvanl (approved):
Thanks!

PR #6334, opened 2023-05-20
Remove apostrophe from "it's" for possessive use

omccully:
"it's" is short for "it is", which is not right. "its" is correct here because that's the possessive word.


gregvanl (approved):
Thanks

PR #6340, opened 2023-05-24
Fix name of the macOS menu item

scottamain:


gregvanl:
Hi @scottamain Thanks for submitting a PR and your interest in the VS Code documentation.
This update will be handled with we merge https://github.com/microsoft/vscode-docs/pull/5937 and do a pass over all the documentation.
Still waiting until Ventura has more adoption and also thinking about how we might include both File menu paths, without it looking too clunky :-)

PR #6345, opened 2023-05-25
Mention Toggle Screen Accessibility Mode command

### LINKED ISSUE #6320: Toggling off Screen Reader Optimized, is not optimized for non-sighted users
In the following link https://code.visualstudio.com/docs/editor/accessibility#_screen-reader-mode

This sentence feels odd 
"The Status Bar displays Screen Reader Optimized in the lower right and **you can exit screen reader mode by clicking on the display text.**"

Given that we are talking about screen readers should we also mention that they could access that setting and toggle it on and off from the Command Pallet?

Screenshot 2023-05-16 at 4 25 17 PM

gregvanl:
Fixes #6320

### DIFF
--- a/docs/editor/accessibility.md
+++ b/docs/editor/accessibility.md
@@ -154,7 +154,7 @@ You can press `kb(editor.action.showAccessibilityHelp)` to trigger the **Show Ac
 
 ## Screen reader mode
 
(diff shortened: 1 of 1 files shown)


meganrogge (approved):
Thanks!

PR #6346, opened 2023-05-25
mask the remote Jupyter server URI

### LINKED ISSUE #5806: Jupyter support doc: Documentation incorrectly mentions blurring not present in image (anymore)
Page: https://code.visualstudio.com/docs/python/jupyter-support-py
Section: https://code.visualstudio.com/docs/python/jupyter-support-py#_connect-to-a-remote-jupyter-server

The ordered list of connection steps claims, in step 4:

> The Python Interactive window indicates where code is run by displaying the URI (which is blurred out in the image below):

But the current version of the image in this repo doesn't have a blurred URL. Nor does it really appear to indicate that jupyter is running remotely, since the visible URL points to `http://localhost:8888/...`.

(The image shown on the docs site, which is also `docs/python/jupyter/jupyter-running-remotely.png` in this repo)

image

Previous revisions of the image did show a blurred, presumably remote, URL, but that was changed by #4729 is no longer the case.

I'm not sure a localhost HTTP URL is a particularly good illustration of a "remote" Jupyter server, since the docs even discuss authentication and configuring SSL. But if the current image continues to be used, then the text about the blurred URL is outdated and should be removed.

gregvanl:
Fixes #5806
Opted to just mask the URI since so much of the section talks about remote Jupyter servers on other computers.

### DIFF
--- a/docs/python/images/jupyter/jupyter-running-remotely.png
+++ b/docs/python/images/jupyter/jupyter-running-remotely.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:f129d353143f1395091c385f11d3f48ab9bacc4052bf1303aaace5c0980358d9
-size 149789
+oid sha256:67a04e343377d509add7aa114807a5ab8631f167721cb15fd7f7c867fb6e350c
+size 140543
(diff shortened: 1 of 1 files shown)


amunger:
the token value should probably be blurred as well - that's much more sensitive information, so people might wonder why we're blurring one but not the other

amunger (approved):

andres395 (commented):
Bloqueo

PR #6352, opened 2023-05-28
Adds Hungarian to Available locales

dmalan:
Available from @microsoft in Marketplace at https://marketplace.visualstudio.com/items?itemName=MS-CEINTL.vscode-language-pack-hu, but seems to be missing from https://code.visualstudio.com/docs/getstarted/locales#_available-locales.


gregvanl (approved):
Thanks!

PR #6391, opened 2023-06-08
Remove mentions of VS Code Insiders for Copilot

bamurtaugh:
Based on the latest https://code.visualstudio.com/updates/v1_79#_use-copilot-chat-in-stable-vs-code.


gregvanl (approved):
Thanks!

PR #6394, opened 2023-06-10
fix: add `not` to testing API introduction

abdul-alhasany:
It seems that `not` should be added given the structure of the sentence.

### DIFF
--- a/api/extension-guides/testing.md
+++ b/api/extension-guides/testing.md
@@ -9,7 +9,7 @@ MetaDescription: Testing APIs in VS Code allow users to discover and run unit te
 
 # Testing API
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @abdul-alhasany Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
The current text reads correctly. Perhaps you read the preceding "than" as "that", in which case you suggestion would be correct?

abdul-alhasany:
@gregvanl You are right. I had to go back and read it few times to realize it. Perhaps we can change the structure of the whole sentence, because it is not easily understood, particularly for non-native English speakers.

PR #6457, opened 2023-07-05
typo in settings.md

DeadlyMissile:
I believe this is the correct command


gregvanl:
Hi @DeadlyMissile I think there are several other mentions of the old command name "Preferences: Open Settings (JSON)" in settings.md. Would you like to fix those as well with this PR?

gregvanl (approved):
Thanks!
I'll follow up and change other instances.

PR #6527, opened 2023-07-21
Add to Yum/DNF the same comment about code-insiders than the other methods

DianaOlympos:
Checked locally that `code-insiders` is present in the repository.

As such, just adding it to the documentation so that the next to read it know it is there :smile: Like it is available for other methods to install.


Tyriar (approved):
Thanks!

PR #6564, opened 2023-08-03
Vscode logo-image link bug fixed: overview.md

Usmanzahidcode:
The image in the learn/overview.md had a broken link, which is now resolved.


gregvanl:
Hi @Usmanzahidcode Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
The image renders correctly on the live code.visualstudio.com site and we need to use the live site relative path since the image is embedded in an img tag.
image

PR #6565, opened 2023-08-03
Update Unity topic for new extension

gregvanl:
Not sure what is going on with the images and Git LFS and why they don't show up in the GitHub preview.
Will double-check this on the staging server after this PR is merged and built


LiamDobbelaere:
What's the motivation behind this change? I think the old info is still relevant because I just set up VSCode on my laptop for Unity with the guide and it wasn't working. To my frustration, I remember there being more steps to it and couldn't find them until I used the Wayback machine.

Now I have no code completions despite having followed the guide and also having C# Dev Kit (which seems to autoinstall), as well as having regenerated project files :\

On my main PC, where I followed the guide before this change, it -does- properly recognize MonoBehaviour as a class and all that, it works fine.

jbevain:
The motivation behind this change is two fold:

* Unity used to support a Visual Studio Code integration which is based on old style `csproj`,
* The new `C#` extension only supports SDK style `csproj`.

We recently shipped a preview of a new experience for Unity developers so they can use the new `C#` extension.

If you're running intro troubles with this preview we'd love to help, please head to https://github.com/microsoft/vscode-dotnettools/ to look at existing issues or file a new one. Thanks!

PR #6579, opened 2023-08-08
Add update note

bamurtaugh:
We've found folks are still referencing https://code.visualstudio.com/blogs/2022/07/07/vscode-server for the latest on the server and tunneling. I propose we leave the guidance in-tact as we consider blogs a point-in-time reference, and we add a link to our latest blog and/or docs.


gregvanl (approved):
Thanks

PR #6587, opened 2023-08-11
fix: add missing pricing parameter to manifest

bandantonio:


gregvanl (commented):
Thanks!

gregvanl (approved):

gregvanl:
Hi @bandantonio While reviewing this PR, I noticed that the `pricing` attribute is not included in the package.json schema and so does not show up in IntelliSense.
Would you like to submit a PR to the vscode product repo to add this attribute to the schema?
You would update IJSONSchema in https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/extensions/common/extensionsRegistry.ts

bandantonio:
@gregvanl Yes, I definitely would like to. Should I create an issue for that?

gregvanl:
@bandantonio That is a good idea for tracking and context and then link the PR to the issue.

You could call out the missing attribute and add an image like:

image

PR #6588, opened 2023-08-11
Clarify macOS support policy

Tyriar:
We primarily want to support what Apple supports. Note that this doesn't mean we won't ship fixes for older versions of macOS, just we may not necessarily and they are typically lower priority than the supported versions.

Related microsoft/vscode#184713

### DIFF
--- a/docs/supporting/requirements.md
+++ b/docs/supporting/requirements.md
@@ -22,7 +22,7 @@ We recommend:
 VS Code is supported on the following platforms:
 
 * Windows 10 and 11 (32-bit and 64-bit)
-* macOS versions receiving security fixes from Apple, this typically includes the latest release and the 2 versions before it
+* macOS versions with Apple security update support. This is typically the latest release and the two previous versions.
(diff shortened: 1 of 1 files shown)


docs/supporting/requirements.md
```diff
@@ -22,7 +22,7 @@ We recommend:
 VS Code is supported on the following platforms:
 
 * Windows 10 and 11 (32-bit and 64-bit)
-* OS X High Sierra (10.13+)
+* macOS versions receiving security fixes from Apple, this typically includes the latest release and the 2 versions before it
```
gregvanl, line 25: How about: 
"macOS versions with Apple security update support. This is typically the latest release and the two previous versions."

Tyriar: Sounds good 👍 

```suggestion
* macOS versions with Apple security update support. This is typically the latest release and the two previous versions.
```

gregvanl (approved):
Thanks!

PR #6589, opened 2023-08-11
Refer to the correct names of the default color themes.

lperson:


gregvanl:
Hi @lperson The stored color theme setting values include the "Default" prefix to indicate that the themes are part of the VS Code default color themes.
You can see this by reviewing `defaultSettings.json` via Preferences: Open Default Settings (JSON)":

image

image

The default color theme has changed from "Default Dark+/Light+" to "Default Dark/Light Modern".
Would you like to update this PR to make those changes?

SaliPasali (approved):

lperson:
Thanks for the review @gregvanl. I'll revise.

docs/getstarted/themes.md
```diff
@@ -26,7 +26,7 @@ The active color theme is stored in your user [settings](/docs/getstarted/settin
 
 ```json
   // Specifies the color theme used in the workbench.
-  "workbench.colorTheme": "Default Dark+"
+  "workbench.colorTheme": "Dark Modern"
```
gregvanl, line 29: The setting value is "Default Dark Modern"

docs/getstarted/themes.md
```diff
@@ -47,10 +47,10 @@ Windows and macOS support light and dark color schemes. There is a setting, `win
 
 To customize the themes that are used when a color scheme changes, you can set the preferred light, dark, and high contrast themes with the settings:
 
-* `workbench.preferredLightColorTheme` - defaults to "Default Light+"
-* `workbench.preferredDarkColorTheme` - defaults to "Default Dark+"
-* `workbench.preferredHighContrastColorTheme` - defaults to "Default High Contrast"
-* `workbench.preferredHighContrastLightColorTheme` - defaults to "Default High Contrast Light"
+* `workbench.preferredLightColorTheme` - defaults to "Light Modern"
```
gregvanl, line 50: The setting values still include "Default" in the string.

lperson: Yeah, I see that. This is the confusion that inspired me to make the PR. The documentation is meant to use the underlying names, but as a user, I use the names in the UI, and they're different.

So thanks for your patience. I'll revise (again). What do you think about noting in the document that the names are different in the UI?"

docs/getstarted/themes.md
gregvanl: This section is calling out the underlying settings as a JSON snippet so it should reflect the actual setting JSON value.

docs/getstarted/themes.md
gregvanl: This section could be changed to use the Settings editor names instead of the settings IDs and in that case, using the Settings editor visible name would be correct.

To customize the themes that are used when a color scheme changes, you can set the preferred light, dark, and high contrast themes in the Settings editor:

* Workbench: Preferred Dark Color Theme - defaults to "Dark Modern"
* Workbench: Preferred High Contrast Color Theme - defaults to "Dark High Contrast"
* etc

gregvanl: Nice to call out the Settings editor here as the previous section used the "Select Color Theme" picker UI

gregvanl:
@lperson Thank you for pointing out the stale color themes.
Closing as covered by https://github.com/microsoft/vscode-docs/pull/6663, which also updates the Color Theme drop down image and removes a few other stale color theme references

PR #6592, opened 2023-08-14
fix typo in extension-capabilities overview

adriancostin6:
Plural is used in the title. Changed it to singular :)


gregvanl (approved):
Thanks!

PR #6604, opened 2023-08-19
Fix broken mermaid link

neshkeev:


gregvanl (approved):
Thanks for spotting this!

PR #6605, opened 2023-08-20
I think that port 3000 should be forwarded.

ecormaksin:
When I tried this tutorial, I could not open 'http://localhost:3000' from a host web browser.
I noticed that the port 3000 of the container was not published by executing 'docker ps' on the host.
Therefore, I think that port 3000 should be forwarded at 'devcontainer.json' (and '9000' at "portsAttributes" section may be wrong...)
And the port 3000 should be opend at the image of container, shouldn't it?


bamurtaugh:
Thanks for opening! The example intends to show https://github.com/microsoft/vscode-remote-try-node/blob/main/.devcontainer/devcontainer.json, so I believe the port should be updated to 3000 instead of 9000, but there shouldn't be a need to uncomment `forwardPorts`.

bamurtaugh:
Port 3000 should be auto-forwarded; there's additional info on https://containers.dev/implementors/json_reference/:

> An array of port numbers or `"host:port"` values (e.g. [`3000, "db:5432"`]) that should always be forwarded from inside the primary container to the local machine (including on the web). The property is most useful for forwarding ports that cannot be auto-forwarded because the related process that starts before the `devcontainer.json` supporting service / tool connects or for forwarding a service not in the primary container in Docker Compose scenarios (e.g. `"db:5432"`). Defaults to `[]`. 

I just tested out using this repo (https://github.com/microsoft/vscode-remote-try-node) in VS Code with the Dev Containers extension, and it was all forwarded properly for me as-is 😄. Are you experiencing an issue where when you run that repo, the port isn't auto-forwarded?

ecormaksin:
@bamurtaugh 
At the first time when I tried [this tutorial](https://code.visualstudio.com/docs/devcontainers/tutorial), I could not open 'http://localhost:3000' from my host browser.

However somehow, I stoped the Docker container(`doker stop <container_id>`), removed the container(`docker rm <container_id>`), removed the base image(`docker rmi <base_image_id>`), and when I tried at the second time, it worked!

I wonder why when I execute `docker ps`, the contaier's port is not port forwarded, I can open the page from my host.

Could you tell me the mechanism?

bamurtaugh:
Glad it's working now!

Since we're discussing specific issue(s) with the tech itself, could you please open an issue with your repro steps and output vs expected behavior in https://github.com/microsoft/vscode-remote-release/issues? And then update this PR so that it's ready for re-review? Thanks so much!

ecormaksin:
I posted another issuse #6616 as question.

gregvanl (approved):
Thanks

PR #6631, opened 2023-09-01
Adding UCRT support for MinGW

AlexandraKemperMS:
With UCRT being added in our extension (among other MSYS versions) as noted in our [changelog](https://github.com/microsoft/vscode-cpptools/releases/tag/v1.17.2) we now need to update our docs to support UCRT as the default minGW installation version.

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -38,7 +38,7 @@ Get the latest version of MinGW-w64 via [MSYS2](https://www.msys2.org/), which p
 1. In this terminal, install the MinGW-w64 toolchain by running the following command:
 
     ```MSYS2
-    pacman -S --needed base-devel mingw-w64-x86_64-toolchain
+   pacman -S --needed base-devel mingw-w64-ucrt-x86_64-toolchain
     ```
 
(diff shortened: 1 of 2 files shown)


docs/cpp/config-mingw.md
```diff
@@ -425,6 +425,16 @@ If you have Visual Studio or WSL installed, you may need to change `compilerPath
 
 You must follow the steps on the [MSYS2 website](https://www.msys2.org/) and use the MSYS CLI to install MinGW-w64, which contains those tools. You will also need to install the full MinGW-w64 toolchain (`pacman -S --needed base-devel mingw-w64-x86_64-toolchain`) to get the `gdb` debugger.
```
gregvanl, line 426: Should the package name be updated for this section as well?

gregvanl:
Should https://code.visualstudio.com/docs/languages/cpp#_example-install-mingwx64-on-windows be updated to UCRT as well?

docs/cpp/config-mingw.md
AlexandraKemperMS: Yes, I updated it

AlexandraKemperMS:
> @gregvanl

This is a good point, I fixed this in my last commit

gregvanl (approved):
Thanks!

PR #6639, opened 2023-09-05
Run Python file in new terminal after running line.

breyed:
For more on the pitfall that these instructions guide away from, see: https://stackoverflow.com/q/61628896/145173

### DIFF
--- a/docs/languages/python.md
+++ b/docs/languages/python.md
@@ -39,7 +39,7 @@ The Python extension then provides shortcuts to run Python code using the curren
 
 ![Using the run python file in terminal button](images/python/run-python-file-in-terminal-button.png)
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @breyed Thanks for the suggestion and creating a PR.
Looking over the SO comments and reading the PR, I don't think the terminal is incompatible so much as still in REPL mode, which expects Python code rather than file output.
Talking with the Python extension team, we've come up with:

"The same terminal is also used for **Run Python File in Terminal**. If that terminal is still running the REPL, you should exit the REPL (`exit()`) or switch to a different terminal before running a Python file."

Would you like to update your PR with this text? Also note that we like to show command text in bold.

breyed:
@gregvanl Good suggestion. I updated the PR accordingly.

gregvanl (approved):
Thanks!

PR #6640, opened 2023-09-05
Suggesting fixes on the launch.json and tasks.json sample files

Ramin2c:
The problem with the current tasks.json and launch.json files is as follows:

1. "${file}" Unlike what this doc suggests, points to the current file which can mean tasks.json and not any cpp file 
2. "${fileDirname}" Points to the directy of the current active file (i.e. tasks.json) which is <path>/.vscode
3. "${fileDirname}/${fileBasenameNoExtension}" Will eventually point to to "<path>/.vscode/tasks"

The combination of the two files is not helping with the debug of any cpp file.
My environment is as follows:

- VS Code 1.18.1
- Windows 10
- C++ environment: running in devconatiner "mcr.microsoft.com/devcontainers/cpp:ubuntu"

### DIFF
--- a/docs/cpp/config-linux.md
+++ b/docs/cpp/config-linux.md
@@ -142,9 +142,9 @@ Your new `tasks.json` file should look similar to the JSON below:
         "command": "/usr/bin/g++",
         "args": [
             "-g",
-            "${file}",
+            "${workspaceFolder}/*.cpp",
             "-o",
-            "${fileDirname}/${fileBasenameNoExtension}"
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @Ramin2c Thank you for your interest in the VS Code and C++ extension documentation and for taking the time to submit a PR.
The goal of this Hello World tutorial is to simply build and debug the active single C++ file. While your approach is more general, we'd like to keep the tutorial scoped to a single Hello World example file.

PR #6644, opened 2023-09-06
fix "Remote/Linux" page not showing up in TOC

Xernisfy:
Recently, I googled VSCode integration with Alpine and found this page [https://code.visualstudio.com/docs/remote/linux](https://code.visualstudio.com/docs/remote/linux).

It doesn't appear in the TOC under "REMOTE" though.
So it cannot be openend from within the docs.
Looking at the other files, it appears that the `Order` property is not set and may cause the issue.
Entry no. 12 seems to be free so I set it to this.

If this is caused by something else, please let me know.
But I dont't know how to preview the whole repo, and in "single file preview" there is no TOC.

### DIFF
--- a/docs/remote/linux.md
+++ b/docs/remote/linux.md
@@ -1,5 +1,5 @@
 ---
-Order:
+Order: 12
 Area: remote
 TOCTitle: Linux Prerequisites
 PageTitle: Linux Prerequisites for Visual Studio Code Remote Development


gregvanl:
Hi @Xernisfy Thank you for your interest in the VS Code documentation and taking the time to submit a PR.
This is by design in order to keep the TOC list from getting too long. Sometimes we'll have reference or very specific pages that we don't surface in the TOC but do link to from other topics (for example there are 9 cross links to docs/remote/linux) and make visible to search.
You are correct that leaving the Order value empty leaves a topic out of the TOC.
Closing now as we'd like to continue to leave this out of the Remote TOC

PR #6645, opened 2023-09-06
Fix highlight lang names

rzhao271:
Fixes the syntax names used for highlighting so that shiki can highlight the blocks properly.

### DIFF
--- a/docs/cpp/config-mingw.md
+++ b/docs/cpp/config-mingw.md
@@ -37,7 +37,7 @@ Get the latest version of MinGW-w64 via [MSYS2](https://www.msys2.org/), which p
 
 1. In this terminal, install the MinGW-w64 toolchain by running the following command:
 
-    ```MSYS2
+    ```sh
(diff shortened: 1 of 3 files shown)


gregvanl:
For command line examples, should we be using `sh` or `bash`? 
In vscode-docs right now we have over 200 instances of ```bash and only 10 of ```sh

rzhao271:
I'm thinking it can just be `sh` in this case considering there's no bash-isms in the code blocks.

gregvanl:
Is there any rendering difference between sh or bash?
I could try experimenting locally to see what changes if we update the use of bash to sh

rzhao271:
Looking at shiki 0.10 docs and the latest shiki docs, sh and bash should produce the same result.
Ref https://github.com/shikijs/shiki/blob/d21acb5fd408c69b750b5d6f41a62eac03f607d2/docs/languages.md

gregvanl (approved):

PR #6657, opened 2023-09-07
Document `editor/lineNumber/context` menu contribution point and its context key

gjsjohnmurray:
Available since April 2023 (see https://github.com/microsoft/vscode/issues/180730)


gregvanl (approved):
Thanks!

PR #6670, opened 2023-09-14
Small edits and formatting

gregvanl:
Also add line break to Custom URL table to adjust column widths more equitably


gregvanl:
After adding the line break to the Marketplace URL, the table should look like:
image

bamurtaugh (approved):
Thank you!

PR #6683, opened 2023-09-21
faq: blank vscode delete GPUCache after update

muhlemmer:
This updates the FAQ to help users that have a blank screen after update.

Having experienced similar issues with Chrome on this hardware, it seems deleting the GPU cache is the best way to solve it.

I've only added the Linux path here, because I don;t have a windows system and I wouldn't know where the GPUCache is stored,


gregvanl (approved):
Thanks!

PR #6688, opened 2023-09-21
Edit pass on C# Dev Kit FAQ updates

gregvanl:
This is a basic edit/formatting pass
I did convert the settings.json screenshots to code blocks

@mikadumont @webreidi let me know if you'd like anything else updated


mikadumont (commented):
I don't see the existing questions can we add those back that were under General? 
Can we add a TOC similar to C++

roblourens (approved):

PR #6689, opened 2023-09-22
Clarify use of postCreateCommand vs. Dockerfile

chrmarti:


bamurtaugh (approved):
Thank you, looks great!

PR #6716, opened 2023-10-05
Properly use ed25519 throughout SSH keys guide

Snaptraks:
The Quick start section has commands to set up key based authentication where most of the commands used a ed25519 key, but the ssh-keygen command uses rsa. Someone following the commands exactly will encounter a problem when trying to copy the id_ed25519 key when it does not exist (but id_rsa will).


gregvanl (approved):
Thanks!

PR #6719, opened 2023-10-05
Clarify language about `.pth` files and import hooks

debonte:
Fix https://github.com/microsoft/pylance-release/issues/4855

The current documentation around editable installs tells users to use `.pth` files instead of import hooks. A user recently pointed out that this language is confusing because installing an import hook requires a `.pth` file. So, the existence of a `.pth` file doesn't indicate that an editable install is Pylance-friendly.

The Pyright docs were [already updated](https://github.com/microsoft/pyright/pull/6079).


debonte:
@luabud

docs/python/editing.md
```diff
@@ -171,7 +171,7 @@ This error happens when Pylance is unable to find the package or module you're i
...
 - If you are importing a package that is not installed, you can install it by running the following command in an activated terminal: `python -m pip install {package_name}`.
 - If you are importing a package that is already installed in a different interpreter or kernel, [select the correct interpreter](/docs/python/environments.md#select-and-activate-an-environment).
-- If you are working with an editable install and it is currently set up to use import hooks, consider switching to using `.pth` files to enhance compatibility and ensure smoother import behavior. Learn more in the [Pyright documentation](https://microsoft.github.io/pyright/#/import-resolution?id=editable-installs).
+- If you are working with an editable install and it is currently set up to use import hooks, consider switching to using `.pth` files that simply contain file paths instead, to enhance compatibility and ensure smoother import behavior. Learn more in the [Pyright documentation](https://microsoft.github.io/pyright/#/import-resolution?id=editable-installs).
```
gregvanl, line 174: We try to avoid using "simply", do you mean to say "only"?
Should the .pth file just contain file paths?

debonte: Yes, only file paths.

gregvanl (approved):
Thanks

PR #6728, opened 2023-10-09
format menu contribution points

gregvanl:
Format with identifier first and description next similar to other sections of Contribution Points document.

### DIFF
--- a/api/references/contribution-points.md
+++ b/api/references/contribution-points.md
@@ -767,42 +767,42 @@ Last, a `group` property defines sorting and grouping of menu items. The `naviga
 Currently extension writers can contribute to:
 
 - `commandPalette` - global Command Palette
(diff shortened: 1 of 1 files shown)


Tyriar (changes requested):
Can these be alphabetical? Not sure what the order is meant to be if not

gregvanl:
The order is kinda sorta UI and use-based; top level UI like the Command Palette and editor first, less common or more specific UI like Timeline and notebooks down lower, which I like.
I'll try reordering alphabetically and see what that looks like. For example, Comments will float up but is only used currently in the GHPRI extension

gregvanl:
@Tyriar Ordering looks fine alphabetically. Ready for another review

Tyriar (approved):

PR #6753, opened 2023-10-18
Use generic name for program path example

### LINKED ISSUE #6722: Bot for PokemonGO as example
Hello,
I was reading the docs on debugging python with vscode when I saw that as an example, a Pokemon Go bot was used...
https://github.com/microsoft/vscode-docs/blob/6560339b629dba7ea5d65f0bf78da0fb46102b1c/docs/python/debugging.md?plain=1#L348-L356

I think we should change that so we don't have a problem one day

alexwkleung:
Resolves #6722.

Let me know if I need to make any changes.


gregvanl (approved):
Thanks!

PR #6754, opened 2023-10-20
Clarify where the `scm/sourceControl` menu appears

gjsjohnmurray:
The screenshot immediately after this change would benefit from an update. It shows an outdated view title; SOURCE CONTROL PROVIDERS is now SOURCE CONTROL REPOSITORIES (noted in the updated text) as well as old Activity Bar icons, no blue Commit button and capitalized CHANGES header. Nor does it even show an open `scm/sourceControl` context menu on one of the four repositories in that view.


gregvanl (approved):
Thanks!

PR #6760, opened 2023-10-24
Add FastAPI tutorial to Python docs 

luabud:
For https://github.com/microsoft/vscode-python/issues/21586


docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+
+Once you're done, you can continue with the [Start Coding](#start-coding) section below.
+
+### Locally in VS Code
+To successfully complete this tutorial in [VS Code](https://code.visualstudio.com/), you need to first setup your Python development environment. Specifically, this tutorial requires:
+- Python 3 (check the [installation guide](python-tutorial.md#install-a-python-interpreter) if you don’t have it installed)
+- [Python extension for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-python.python) (For additional details on installing extensions, see [Extension Marketplace](https://code.visualstudio.com/docs/editor/extension-marketplace))
+
+In this section, we will create a folder to be opened as a workspace in VS Code, set up a Python virtual environment, install the project’s dependencies and write the first lines of our FastAPI app.
```
lszomoru, line 37: Don't want to be pedantic but this section does not include writing any code as that is in the next section so I would change this to: 

> In this section, we will create a folder to be opened as a workspace in VS Code, set up a Python virtual environment, and install the project’s dependencies.

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+7. When asked by the environment type, select **Venv**:
+![Drop down with "Venv" or "Conda" as options for environments that can be created with the Python: Create Environment command.](images/environments/create_environment_dropdown.png)
+
+8. Then select the latest version of Python available on your machine:
+![List of available global environments that can be used to create a virtual environment.](images/fastapi-tutorial/create_environment_interpreters_list.png)
+
+9. Select the requirements.txt file from the drop-down list, so the dependencies are automatically installed, and then select "OK":
+![Check box selected to install dependencies from requirements.txt file](images/fastapi-tutorial/create_environment_select_requirements.png)
+
+The virtual environment will be created, the dependencies will be automatically installed and the environment will be selected for your workspace to be used by the Python extension.
```
lszomoru, line 67: It would be nice to add steps (if any exist) on how to confirm that the virtual environment has been created successfully and it has been selected in the workbench.

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+
+### Locally in VS Code
+To successfully complete this tutorial in [VS Code](https://code.visualstudio.com/), you need to first setup your Python development environment. Specifically, this tutorial requires:
+- Python 3 (check the [installation guide](python-tutorial.md#install-a-python-interpreter) if you don’t have it installed)
+- [Python extension for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-python.python) (For additional details on installing extensions, see [Extension Marketplace](https://code.visualstudio.com/docs/editor/extension-marketplace))
+
+In this section, we will create a folder to be opened as a workspace in VS Code, set up a Python virtual environment, install the project’s dependencies and write the first lines of our FastAPI app.
+
+1.	In your file system, create a project folder for this tutorial, such as `groceries-plugin`.
+2.	Open this new folder in VS Code (**File** > **Open Folder…** or `kb(workbench.action.files.openFolder)`).
```
lszomoru, line 40: Should we add details on whether users should trust/not trust the workspace? I was able to create the virtual environment but other sections might only work in a trusted workspace. I think that we should probably explicitly call out that the user should trust the workspace.

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+
+    class ItemPayload(BaseModel):
+        item_id: Optional[int]
+        item_name: str
+        quantity: int
+    ```
+
+## Create Routes
+Now we need a place to store the grocery list items. For simplicity, let’s start with an empty dictionary.
+1. First, let's import all the packages we will need for the sample. Open the main.py file and replace the first import line with the following ones:
```
lszomoru, line 122: main.py should be `main.py`

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+5. From the drop-down menu, select the **FastAPI** configuration option from the drop-down list:
+![Drop down with debugger configuration options, with FastAPI being highlighted.](images/fastapi-tutorial/fastapi_debug_config_option.png)
+
+    This will invoke uvicorn to start the application server through the debugger, which will allow us to step through the code to inspect its behavior. You will see something like the following in the terminal:
+
+    ![Uvicorn server running message displayed in the terminal, with an URL to access the app.](images/fastapi-tutorial/fastapi_debug_terminal.png)
+
+6. `Ctrl` + click the http://127.0.0.1:8000/ URL in the terminal to open your default browser to that address:
+![Hello World message displayed in the browser.](images/fastapi-tutorial/helloworld_browser.png)
+7. Stop the debugger by clicking on the stop icon in the debug toolbar, or through `kb(workbench.action.debug.stop)`.
```
isidorn, line 94: Could we between 6 and 7 say something like "congrats your FastAPI is running! You can set a breakpoint..."
Or something like that. I would like a sense of achievement here :)

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,601 @@
...
+Pydantic is a data validation and parsing library that integrates seamlessly with FastAPI. It allows you to define data models using Python classes with [type hints](https://docs.python.org/3/library/typing.html) for automatic validation and parsing of incoming data (which are called payloads) in API requests.
+
+Pylance, the default language server for Python in VS Code, supports type hinting features that can be helpful for working with Pydantic models and FastAPI. Let's enable a few of them:
+1. Open the Settings UI page (`kb(workbench.action.openSettings)`)
+2. Search for "python type checking mode" and set it to "basic" to enable basic type checking. This will enable Pylance to show diagnostics and warnings to catch simple type-related errors. Alternetivaly, you can set it to "strict" to enable more advanced [type checking rules](https://microsoft.github.io/pyright/#/configuration?id=diagnostic-rule-defaults).
+    ![Python Analysis Type Checking Mode options (off, basic and strict) in settings UI page.](images/fastapi-tutorial/type_checking_mode_setting.png)
+3. Next, search for "Python inlay type hints", and enable inlay hints for Variable Types and Function Return Types:
```
isidorn, line 104: The inlay hints part sounds to me like an optional part.
I suggest to either have it at the end of the document or to mention this is optional goodness.
In the tutorial I wold like to have the minimal steps to get into a good state, and then additional I can do more stuff to make my life easier.

On a related note, why don't we turn these on by default?

docs/python/tutorial-fastapi.md
luabud: Great catch, thank you!

docs/python/tutorial-fastapi.md
luabud: Good point. Added a screenshot as well as a note with a workaround in case the Python extension fails to auto select the venv for some reason.

docs/python/tutorial-fastapi.md
luabud: That's a good point. I made some edits to make it clear this is optional, but also added a bit more context around the value of using Pylance's typing features with FastAPI projects. 

We have plans to turn the inlay hints on by default for files we see explicit type information cause we agree it'd be helpful. However we don't have plans to add it by default in all Python files given a lot of people in the Python community would be highly annoyed by that 😅 Typing in Python in usually seen as essential for production-ready code, but for data analysis, prototyping and basic scripting, a lot of folks in the community find types to be too constrictive and even distracting.

docs/python/tutorial-fastapi.md
luabud: I love that! This inspired me to add a few more congrats on success points, as well as a thank you note at the end. Thanks for calling that out 😊

docs/python/tutorial-fastapi.md
luabud: Great point, added a step calling out trusting the workspace. Thanks!

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+
+The 3 steps below are optional, but given FastAPI leverages type hints extensively to improve code readability and validation, we can take advantage of Pylance's type checking features to catch errors early on:
+
+1. Open the Settings UI page (`kb(workbench.action.openSettings)`)
+2. Search for "python type checking mode" and set it to "basic" to enable basic type checking. This will enable Pylance to show diagnostics and warnings to catch simple type-related errors. Alternatively, you can set it to "strict" to enable more advanced [type checking rules](https://microsoft.github.io/pyright/#/configuration?id=diagnostic-rule-defaults).
+    ![Python Analysis Type Checking Mode options (off, basic and strict) in settings UI page.](images/fastapi-tutorial/type_checking_mode_setting.png)
+3. Next, search for "Python inlay type hints", and enable inlay hints for Variable Types and Function Return Types:
+    ![Two Python Analysis Type Hints settings being enabled in the settings UI page: for Function Return Types and for Variable Types.](images/fastapi-tutorial/function_and_variable_return_type_hint_settings.png)
```
cwebster-99, line 123: Since this is marked as optional, do you think we should put the steps for adding the model before this section? My thinking is that it allows the user to get to the crucial information quicker but I also see that enabling these settings makes sense before adding the model for the users that do this step. Or maybe this could be formatted as a 💡Tip? +0 but would love to hear your thoughts.

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+            # generate an id for the item based on the highest ID in the grocery_list
+            item_id = max(grocery_list.keys()) + 1 if grocery_list else 0
+            grocery_list[item_id] = ItemPayload(
+                item_id=item_id, item_name=item_name, quantity=quantity
+            )
+
+        return {"item": grocery_list[item_id]}
+    ```
+
+    If you have enabled type hints, you might notice Pylance adds inlay hints with the function return type, as well as the types for `item_ids` and `item_id`. You can optionally double click on each suggestion to insert them into the code:
```
cwebster-99, line 182:
```suggestion
    If you have enabled type hints in the previous section, you might notice Pylance adds inlay hints with the function return type, as well as the types for `item_ids` and `item_id`. You can optionally double-click on each suggestion to insert them into the code:
```

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+    ![Inlay function return and variable type hints being displayed by Pylance throughout the sample code.](images/fastapi-tutorial/pylance_inlay_hints.png)
+
+Now let's see if this route is working as expected. The fastest way to do so is to leverage both VS Code's debugger as well as FastAPI's `/docs` endpoint, which provides information about all the available API routes and allows you to interact with the API to explore their parameters and responses. This documentation is generated dynamically based on the metadata and type hints defined in the FastAPI application.
+
+4. Add a breakpoint next to the if quantity <= 0 statement, by clicking on the left margin of the line number (or `kb(editor.debug.action.toggleBreakpoint)`). This will make the debugger stop prior to the execution of that line, allowing us to inspect the code line by line.
```
cwebster-99, line 187:
```suggestion
4. Add a breakpoint next to the if `quantity <= 0` statement, by clicking on the left margin of the line number (or `kb(editor.debug.action.toggleBreakpoint)`). This will make the debugger stop prior to the execution of that line, allowing us to inspect the code line by line.
```

cwebster-99 (approved):
This is looking great with all of the feedback incorporated! Awesome work!

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+
+If you are already working on **GitHub Codespaces** with our existing template, you can skip directly to the ["Replace the database"](#replace-the-database) section.
+
+If you are on Windows, you can work with Redis by setting up either a [Docker container](https://www.docker.com/products/docker-desktop/), a [GitHub Codespace](https://github.com/features/codespaces) or through [WSL](https://learn.microsoft.com/en-us/windows/wsl/) (Windows Subsystem for Linux). In this tutorial we will use a Docker container, but you can refer to the [section above](#github-codespaces) for instructions on how to set up a GitHub Codespace.
+
+Otherwise, if you are on a Linux or a macOS machine, you can install Redis by following the [instructions on their website](https://redis.io/docs/getting-started/), and then skip to the ["Replace the database"](#replace-the-database) section.
+
+### Setting up a Docker Container on Windows
+The Dev Container extension for VS Code offers a simple way to configure a Linux-based workspace in your Windows machine, putting your entire project and its dependencies as well as tools you need for your project into a tidy container.
```
bamurtaugh, line 317: Should be "Dev Containers" (plural), and would recommend linking to the actual extension!

bamurtaugh: Also recommend wording as "The VS Code Dev Containers extension"

bamurtaugh, line 317: I love you're bringing in the Dev Containers extension! Though I'm not sure I would introduce it as a way to set up a Linux-based environment. Would recommend pulling more from the intro in https://code.visualstudio.com/docs/devcontainers/containers and focusing on the full-featured environment part

bamurtaugh: Could also mention the WSL extension, if you really want to focus on the Linux-based part at some point

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+For the steps below, make sure you have the following requirements installed on your machine:
+
+#### Requirements
+-	[Docker for Windows](https://www.docker.com/)
+-	[Dev Containers extension for VS Code](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
```
bamurtaugh, line 322:
```suggestion
-	[VS Code Dev Containers extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers)
```

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+
+#### Create the Dev Container configuration
```
bamurtaugh, line 324: "Dev container" should be lowercase, unless referring to the "Dev Containers extension" or "Dev Container Spec"

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+1. Open the Command Palette and run the “Dev Container: Add Dev Container Configuration Files…"
```
bamurtaugh, line 325: Extension name and its commands should always be plural "containers"

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+2. Select "Python 3":
+
+    <img src="images/fastapi-tutorial/devcontainers_python3.png" alt="Python 3 option selected in the Dev Container configuration files list." width="500"/>
+3. Select the default version.
+
+We can optionally install features to be included in the container. For this tutorial, we will install [Redis Server](https://github.com/itsmechlark/features/tree/main/src/redis-server), which is a community contributed feature that installs and adds the proper dev container set up for Redis.
```
bamurtaugh, line 331: "Features" should be capitalized here and throughout the rest of the doc. Would also recommend providing some more info on what Features are and linking to their docs (as I'd expect many users aren't familiar with them)

```suggestion
We can optionally install Features to be included in the container. For this tutorial, we will install [Redis Server](https://github.com/itsmechlark/features/tree/main/src/redis-server), which is a community contributed Feature that installs and adds the proper dev container set up for Redis.
```

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,652 @@
...
+
+4. Select "Redis Server" as an additional feature to be installed, press "OK", and then select "Keep Defaults".
+    <img src="images/fastapi-tutorial/devcontainers_redis_server_feature.png" alt="Redis Server option selected in the Dev Container configuration files list." width="500"/>
+
+This will create a `.devcontainer` folder in your workspace, with a `devcontainer.json` file. Let's make some edits to this file so the container setup will include steps such as installing the VS Code extensions we'll need as well as the project dependencies.
+
+5. Open the `devcontainer.json` file.
+6. Add a "," after the `"features" : { ... }` entry, so we can add more settings to the file.
+7. Locate the content below and remove the comment (`//`) from that line, so the dependencies can be installed once the container is created:
```
bamurtaugh, line 340: Recommend explaining a bit more about `postCreateCommand` and maybe link to https://containers.dev/implementors/json_reference/#lifecycle-scripts

bamurtaugh (commented):
Thanks so much for mentioning dev containers, love this cross-linking! I think the steps are great, mostly commented on some formatting or cross-linking suggestions.

docs/python/tutorial-fastapi.md
luabud: Yeah that's a great point. I swapped these two instruction blocks within this section!

docs/python/tutorial-fastapi.md
luabud: Great point! I tweaked it to frame it more around the "full featured dev environment" instead of "Linux-based". I mention WSL in the previous section too 😊

docs/python/tutorial-fastapi.md
luabud: This all makes tons of sense, thank you! I _think_ I fixed all occurrences 😅

docs/python/tutorial-fastapi.md
```diff
@@ -0,0 +1,660 @@
+---
+Order: 10
+Area: python
+TOCTitle: FastAPI Tutorial
+ContentId: 0d32bced91aa5c2ee5696fc7995370ae
+PageTitle: Python and FastAPI Tutorial in Visual Studio Code
+DateApproved:
+MetaDescription: Python FastAPI tutorial showing IntelliSense and debugging support in Visual Studio Code, the best Python IDE.
```
gregvanl, line 8: True but we should be humble

luabud: I copied this from the [Django](https://github.com/microsoft/vscode-docs/blob/main/docs/python/tutorial-django.md) and [Flask ](https://github.com/microsoft/vscode-docs/blob/main/docs/python/tutorial-flask.md) tutorials FWIW 😅 I think it might have been added for SEO ranking?

gregvanl: Too funny, I never noticed. No one has complained and you may be right about SEO juice. Leave it in.

gregvanl (approved):

gregvanl:
Thanks!

PR #6766, opened 2023-10-27
Fix screenshots and shorten last header

luabud:
Making some final tweaks :)


gregvanl (approved):
Awesome.
Thanks for fixing the double F5!

PR #6769, opened 2023-10-27
Remove non-existent keybindings

gregvanl:
Open Folder not available on macOS.
Couple of typos


luabud (approved):
Great catch! thanks :)

PR #6791, opened 2023-11-03
Fix language and list in v1_84.md

JohnVeness:
Fixed a couple of things in the release notes, namely the awkward repeat of "to the" regarding the Activity Bar, and changed the order of the codicons list to match the image.


gregvanl (approved):
Thanks! Good changes

PR #6815, opened 2023-11-10
Clarify that Panel can only be maximized when center-aligned

gjsjohnmurray:
Prompted by https://github.com/microsoft/vscode/issues/197873


gregvanl (approved):
Thanks!

PR #6822, opened 2023-11-13
Describe word wrap behaviour

hickford:


gregvanl (approved):
Thanks. Nice clarification

PR #6824, opened 2023-11-13
C#: Add debugger exception settings section

gregg-miskelly:
The C# debugger docs were missing information on how to configure exception settings. This also removes usage of `https://aka.ms/VSCode-CS-LaunchJson`, as this repo should be the source of truth about debugger docs.


gregvanl (approved):
Thanks!

PR #6829, opened 2023-11-15
Removed yarn parameter in deploying from GitHub

dangreene0:
## Summary
Specifying `--yarn` will prevent people using npm from `building` or `deploying` the package.

## Examples
- [Example of a this breaking an Action](https://github.com/dangreene0/tf2-vscript-snippets/actions/runs/6871551981)
- [Example of workflow without yarn parameter](https://github.com/dangreene0/tf2-vscript-snippets/actions/runs/6871636061/job/18688806034)

## Solutions
- Remove the `--yarn` instruction since it should detect what is necessary. (In this PR)
- Specify to use `--yarn` when they install with yarn.
- Separate `--yarn` and `--no-yarn` sections.

### DIFF
--- a/api/working-with-extensions/continuous-integration.md
+++ b/api/working-with-extensions/continuous-integration.md
@@ -87,7 +87,7 @@ You can enable the build to run continuously when pushing to a branch and even o
 
 ```json
 "scripts": {
-  "deploy": "vsce publish --yarn"
+  "deploy": "vsce publish"
 }
 ```
 


gregvanl:
Hi @dangreene0 Thank you for your interest in the VS Code documentation and taking the time to create a PR.
Currently the topic uses yarn throughout and mentions this under Common Questions https://code.visualstudio.com/api/working-with-extensions/continuous-integration#do-i-need-to-use-yarn-for-continuous-integration so even though `vsce publish` should detect that someone is using yarn in most cases, we'd like to keep the instructions consistent.
I've created https://github.com/microsoft/vscode-docs/issues/6832 to look into moving the examples off yarn to npm.

dangreene0:
> Hi @dangreene0 Thank you for your interest in the VS Code documentation and taking the time to create a PR. Currently the topic uses yarn throughout and mentions this under Common Questions https://code.visualstudio.com/api/working-with-extensions/continuous-integration#do-i-need-to-use-yarn-for-continuous-integration so even though `vsce publish` should detect that someone is using yarn in most cases, we'd like to keep the instructions consistent. I've created #6832 to look into moving the examples off yarn to npm.

Thank you very much for your response and explanation. I appreciate your time and agree with your suggested solution.

PR #6831, opened 2023-11-15
Clarify that terminal view maximize requires Center panel alignment

gjsjohnmurray:
Arose from https://github.com/microsoft/vscode/issues/198275


gregvanl (approved):
Thanks!

PR #6836, opened 2023-11-16
Update jsconfig.md

kazuma0129:
Hi team. I have updated the documentation as I found that the content was out of date.

I referred here to the content that should be updated.
ref: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/jsconfig.json

### DIFF
--- a/docs/languages/jsconfig.md
+++ b/docs/languages/jsconfig.md
@@ -86,9 +86,9 @@ Below are `jsconfig` `"compilerOptions"` to configure the JavaScript language su
 Option  | Description
 ----------------|-----
 `noLib` | Do not include the default library file (lib.d.ts)
-`target`| Specifies which default library (lib.d.ts) to use. The values are "es3", "es5", "es6", "es2015", "es2016", "es2017", "es2018", "es2019", "es2020", "esnext".
-`module` | Specifies the module system, when generating module code. The values are "amd", "commonJS", "es2015", "es6", "esnext", "none", "system", "umd".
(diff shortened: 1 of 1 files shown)


kazuma0129:
By the way, off the topic, there seems to be a difference in definition between the content of the document and the json schema (which is what is suggested by the VSCode).
For example, the value of `module` is set to `es2015` in the documentation, but the value defined in the json schema is actually `ES2015`. ref: https://github.com/SchemaStore/schemastore/blob/c0e78776f848b8ea866f379f159109adf4d62ad5/src/schemas/json/jsconfig.json#L352-L371

I don't understand and would like to know, does this affect the working of the system?
If possible, we would like to match the documentation and schema definitions to avoid confusion.

gregvanl (approved):
Thanks!
I'll investigate whether we should use upper case for the setting values

kazuma0129:
@gregvanl Thanks for the review. Please feel free to let me know if I can be of any help.

gregvanl:
@kazuma0129 Here's the follow up PR to update the setting value casing to match IntelliSense.
The casing doesn't matter in the JSON but can be confusing so thank you for pointing out the discrepancy!
https://github.com/microsoft/vscode-docs/pull/6839

kazuma0129:
> Here's the follow up PR to update the setting value casing to match IntelliSense.
The casing doesn't matter in the JSON but can be confusing so thank you for pointing out the discrepancy!
https://github.com/microsoft/vscode-docs/pull/6839

Looks great! :+1: Thank you.

PR #6837, opened 2023-11-16
Update typescript-tutorial 404 not found urls

seungwonleee:
Modified the 404 not found page to work properly.


gregvanl (approved):
Thanks!

PR #6838, opened 2023-11-16
Update vscode://file paths in the docs

denis-sokolov:
The syntax VS Code desires was recently clarified in a comment: https://github.com/microsoft/vscode/issues/197319#issuecomment-1793663562

This commit aligns the documentation with that comment.

To be clear, I am not in favor of this change, the new syntax looks misleading and there is a needless dichotomy between Windows and other versions. But as per the answer in the other issue, this is currently the right way to invoke VS Code.

### DIFF
--- a/docs/editor/command-line.md
+++ b/docs/editor/command-line.md
@@ -142,25 +142,37 @@ You can also open projects and files using the platform's URL handling mechanism
 Open a project
 
 ```bash
+# Windows
 vscode://file/{full path to project}/
-
 vscode://file/c:/myProject/
+
+# Other
+vscode://file{full path to project}/
+vscode://file/home/user/myProject/
 ```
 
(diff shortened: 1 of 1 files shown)


andrewnicols:
Pinging @bpasero for this in relation to https://github.com/microsoft/vscode/issues/197319

Shivamax679 (approved):

bpasero:
How about just removing the generic links and only provide examples so that its clear how to use the links?

denis-sokolov:
It seems like with [the latest change](https://github.com/microsoft/vscode/issues/198509#issuecomment-1816205881) we don’t necessarily need any change in the docs. I’ll close this for cleanup, but if anyone decides to iterate, you can freely use my original suggestion if need be.

PR #6852, opened 2023-11-22
remove an unnecessary whitespace

xia0hj:
Hi I found an unnecessary whitespace here, no whitespace in start region `//#region` that end region should also be consistent.


gregvanl (approved):
Thanks

PR #6859, opened 2023-11-25
Merged bullet points on Copilot Chat requirements

Balastrong:
With #6685 the second sub-bullet point has been removed, I think we can now inline it on a single bullet point.


gregvanl (approved):
Thanks!

PR #6867, opened 2023-11-30
Azure Account -> Azure Resources and GitLab support removal, GitHub addition

bwateratmsft:
A couple of things in this PR.

- Azure Account is being slowly deprecated and replaced with the built-in Microsoft authentication provider. I could use some help with the verbiage here. The Azure Resources extension is required to sign in to some tenants (e.g., those with extra MFA after the primary sign-in) and is recommended (though not strictly required) if a sovereign cloud is being used.
- GitLab container registry support has been removed from the Docker extension (see https://github.com/microsoft/vscode-docker/issues/4132)
- GitHub container registry support has been added to the Docker extension (see https://github.com/microsoft/vscode-docker/issues/869)

Closes https://github.com/microsoft/vscode-docker/issues/4147

### DIFF
--- a/docs/containers/images/registries/github.png
+++ b/docs/containers/images/registries/github.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a564c23da454dffa723d909244b1cc63ffd5a55d1c5386fb06595c828a07b307
+size 10569
--- a/docs/containers/images/registries/github.png
+++ b/docs/containers/images/registries/github.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:a564c23da454dffa723d909244b1cc63ffd5a55d1c5386fb06595c828a07b307
-size 10569
+oid sha256:8292094c9741331f1cc6dd9e35fbec441e478e3efcb31d10068990ccdfde59ac
+size 11214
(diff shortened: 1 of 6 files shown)


fiveisprime:
The changes look good. The Python and Java articles are pretty intense - it's not something to change in this PR, but I wonder how helpful those are.

> Alex, we may want to add some specific documentation somewhere on the docs site about the Azure Resources extension, with particular focus on sign in (sovereign clouds, MFA-requiring tenants, etc.)

I like how the Docker section has an overview. Maybe the Azure section could use an overview rather than the "extensions" page that goes over the Azure Resources extension. That "extensions" landing page was built when it was more difficult to discover the various service extensions, which isn't so much an issue now that every service shows up by default.

ghogen:
@bwateratmsft  
Thanks, this looks great, I just have one question about that first image that shows your GitHub id. I don't know the VS Code policies as well, but on the Learn platform, we have a PII policy that would ask us to adjust the image in some way to not show any PII.

bwateratmsft:
> @bwateratmsft Thanks, this looks great, I just have one question about that first image that shows your GitHub id. I don't know the VS Code policies as well, but on the Learn platform, we have a PII policy that would ask us to adjust the image in some way to not show any identifiable information.

I'll take a new screenshot without. But the old GitLab image had a GitLab username in it so I didn't think it'd matter.

ghogen (approved):

gregvanl:
published

PR #6876, opened 2023-12-04
Replace quickFixWidgetVisible with codeActionMenuVisible

majutsushi:
`quickFixWidgetVisible` seems to have been superseded by `codeActionMenuVisible`. The former doesn't exist in the current code base, while the latter was introduced about a year ago.

References:
https://github.com/microsoft/vscode/issues/55111
https://github.com/microsoft/vscode/blob/f0d15e28789098cbb6390f793d27e71a85fa5537/src/vs/platform/actionWidget/browser/actionWidget.ts


gregvanl (approved):
Thanks!

PR #6896, opened 2023-12-09
Fix typo.

IEvangelist:
I believe you're referring to "thenable", not "themable".


gregvanl:
Hi @IEvangelist Thanks for your interest in the VS Code documentation and taking the time to submit a PR.
The current text is correct as extensions that contribute custom UI such as webviews should make sure they can adjust to the user's selected Color Theme (light, dark, high contrast, etc.)

PR #6928, opened 2023-12-19
Clarify snippet escaping rules

DanTup:
The current text reads like you can escape $ and } where not strictly necessary, but according to VS Code's behaviour and https://github.com/microsoft/vscode/issues/201059 this is not the case - you may only escape the characters that are required to be escaped, otherwise you'll see backslashes in the output.

LSP PR: https://github.com/microsoft/language-server-protocol/pull/1868


gregvanl (approved):
Thanks!

PR #6930, opened 2023-12-20
Update config-linux.md

max-hill-4:
Documentation incorrectly states to replace ${file} with "${workspaceFolder/*.cpp" which would lead to two quotes surrounding the arg.

I have removed the incorrect characters.

### DIFF
--- a/docs/cpp/config-linux.md
+++ b/docs/cpp/config-linux.md
@@ -190,7 +190,7 @@ with this:
 
 ### Modifying tasks.json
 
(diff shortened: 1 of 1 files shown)


gregvanl:
Hi @max-hill-4 Thank you for your PR and your interest in the VS Code documentation.
I see your point and it is confusing that double quotes aren't being used consistently.
Perhaps a better changes would be to put double quotes around `${file}` so that readers know to replace the entire quoted string:
You can modify your `tasks.json` to build multiple C++ files by using an argument like "${workspaceFolder}/*.cpp" instead of "${file}".
This sentence also is repeated in the other 5 tutorials (Windows, WSL, and macOS). Would you like to update those instances as well in this PR? Thanks

gregvanl:
Closing in favor of updates in #6934.
@max-hill-4 Thanks again for pointing out the discrepancy!

max-hill-4:
Will do soon :D !

gregvanl:
Hi @max-hill-4 I made the edits (putting double quotes around `"${file}"`) to the C++ tutorials in the PR #6934 mentioned above.
Thanks again for noticing the mismatch!

PR #6940, opened 2023-12-30
Follow up Vue.js tutorial for new port and flow

gregvanl:
After the recent update for Vite and Vue 3, a couple of updates were missed.
The localhost port is now 5173 and the command to run the local server is `npm run dev`


ntrogh (approved):

ntrogh:
Thanks!

PR #6954, opened 2024-01-11
docs: add client faq for older linux versions

deepak1556:
Follow-up to https://github.com/microsoft/vscode/pull/192513

### DIFF
--- a/docs/supporting/faq.md
+++ b/docs/supporting/faq.md
@@ -301,7 +301,7 @@ VS Code will no longer provide product updates or security fixes on macOS Mojave
 
 ## Can I run VS Code on older Linux distributions?
 
-VS Code desktop version starting with 1.86 (January 2024) is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, Ubuntu 20.04.
+Starting with VS Code release 1.86 (January 2024), VS Code desktop is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, or Ubuntu 20.04.
 
 If you are unable to upgrade your Linux distribution, the recommended alternative would be to use our [web client](/docs/editor/vscode-web.md). If you would like to use the desktop version then you can download the `1.85` release from [here](https://code.visualstudio.com/updates/v1_85). Depending on your platform, make sure to disable updates to stay on that version. A good recommendation is to set up the installation with [Portable Mode](/docs/editor/portable.md).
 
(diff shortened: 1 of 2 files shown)


docs/supporting/faq.md
```diff
@@ -299,6 +299,12 @@ VS Code desktop version starting with 1.83 (September 2023) is deprecating suppo
 
 VS Code will no longer provide product updates or security fixes on macOS Mojave (versions 10.14 and older) and VS Code version 1.85 will be the last available release for macOS Mojave (10.14 and older). You can learn more about upgrading your macOS version at [support.apple.com](https://support.apple.com/en-us/HT201260).
 
+## Can I run VS Code on old Linux distributions?
```
ntrogh, line 302:
```suggestion
## Can I run VS Code on older Linux distributions?
```

docs/supporting/faq.md
```diff
@@ -299,6 +299,12 @@ VS Code desktop version starting with 1.83 (September 2023) is deprecating suppo
...
+
+VS Code desktop version starting with 1.86 (January 2024) is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, Ubuntu 20.04.
```
ntrogh, line 304:
```suggestion
Starting with VS Code release 1.86 (January 2024), VS Code desktop is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, or Ubuntu 20.04.
```

docs/supporting/faq.md
```diff
@@ -299,6 +299,12 @@ VS Code desktop version starting with 1.83 (September 2023) is deprecating suppo
...
+
+If you are unable to upgrade your Linux distribution, the recommended alternative would be to use our [web client](/docs/editor/vscode-web.md). If you would like to use the desktop version then you can download the `1.85` release from [here](https://code.visualstudio.com/updates/v1_85). Depending on your platform, make sure to disable updates to stay on that version. A good recommendation is to set up the installation with [Portable Mode](/docs/editor/portable.md).
```
ntrogh, line 306:
```suggestion
If you are unable to upgrade your Linux distribution, the recommended alternative is to use our [web client](/docs/editor/vscode-web.md). If you would like to use the desktop version, then you can download the VS Code release 1.85 from [here](https://code.visualstudio.com/updates/v1_85). Depending on your platform, make sure to disable updates to stay on that version. A good recommendation is to set up the installation with [Portable Mode](/docs/editor/portable.md).
```

ntrogh (changes requested):
Minor text updates.

ntrogh (approved):

PR #6970, opened 2024-01-21
Fix example code in testing-extension.md

kt3k:
When I ran the example file in `Testing Extensions` docs, it failed with `ReferenceError: after is not defined`. The test file looks like written in [TDD interface](https://mochajs.org/#tdd) of mocha. I think we need to use `suiteTeardown` hook instead of `after` hook (which is only available in BDD interface).


ntrogh (approved):

ntrogh:
Thanks for contributing, @kt3k !

PR #6973, opened 2024-01-22
Use npx to invoke Yeoman

datenreisender:
Nowadays it is usually bad advice to install package globally and also not needed to just invoke a command once. So, it is better to use `npx` to invoke Yeoman here.

### DIFF
--- a/api/get-started/your-first-extension.md
+++ b/api/get-started/your-first-extension.md
@@ -9,10 +9,10 @@ MetaDescription: Create your first Visual Studio Code extension (plug-in) with a
 
 # Your First Extension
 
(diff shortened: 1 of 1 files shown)


aeschli:
Installing `yo` and `generator-code` globally just gives you a very nice command line `yo code`.

`npx --package yo --package generator-code -- yo code` is not something I would remember or type out.

Maybe we can suggest both?

ntrogh:
@datenreisender thanks for contributing! I suggest we include both options in the docs.

datenreisender:
Many prefer not to pollute their global space any longer but I can understand that for others a short `yo code` is preferable. So I pushed a second commit which offers both option. What do you think?

PR #6987, opened 2024-01-28
I updated  the download link for the MSYS2 installer, which is the most recent

dchidindu5:
Updated the download link for MSYS2 installer to the most recent one.


ntrogh:
Thanks for contributing @dchidindu5!

AlexandraKemperMS (approved):
Yes, that's the newest version. Thanks for adding!

PR #6989, opened 2024-01-29
Numbers stay consistent

green961:
https://code.visualstudio.com/api/extension-guides/command#using-a-custom-when-clause-context
搜狗截图20240129130814


ntrogh (approved):
@green961 Nice catch! Thanks for contributing.

PR #6991, opened 2024-01-30
docs: update supported distros for remote

deepak1556:
Fixes https://github.com/microsoft/vscode-remote-release/issues/9449

I haven't updated the section ` Updating glibc and libstdc++ on RHEL / CentOS 6` for RHEL 7 distros since I haven't tested the flow and found the required package names.

### DIFF
--- a/docs/remote/linux.md
+++ b/docs/remote/linux.md
@@ -13,11 +13,11 @@ Linux is a highly variable environment and the large number of server, container
 
 The extensions are known to work when connecting to recent stable/LTS version of:
 
-* **Ubuntu 64-bit x86, ARMv8l (AArch64)** (16.04+, IoT 18.04+)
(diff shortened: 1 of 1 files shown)


chrmarti (approved):
LGTM!

fulldecent:
I see that today RHEL 7 hosts is no longer working for remote SSH development. This PR is the change that documents this.

Therefore, it's going to be a long day for me. Maybe other people with RHEL will see this note first. And maybe we can provide some reference to what people using RHEL 7 can do. At current, I don't have the answer for this cohort.

ntrogh:
@fulldecent Thanks for reaching out. 
Our team has a conversation going on in the VS Code repo (https://github.com/microsoft/vscode/issues/203375). We will provide updates there in the next couple of days and will also be documenting this on our website.

PR #6997, opened 2024-01-31
Fix incorrect code sample in api/extension-guides/webview.md

gusalbukrk:
I couldn't run the code sample in question because it has syntax errors.


ntrogh (approved):
Thanks for your contribution, @gusalbukrk!

PR #7007, opened 2024-02-01
properly escape pipe command in General distribution notes

goi42:


ntrogh (approved):
Good catch @goi42 ! Thanks for contributing.

PR #7010, opened 2024-02-02
Add variable `${/}` new in v1.86

muzimuzhi:
`${/}` variable was implemented in microsoft/vscode#200750 and added to release note of v1.86 in 938983cf7379, but missing from the list of predefined variables.

See also https://code.visualstudio.com/updates/v1_86#_shorthand-for-path-separator-variable.


ntrogh (approved):
@muzimuzhi Thanks a lot for contributing! We hope you like the updates in our latest release.

PR #7013, opened 2024-02-02
docs/terminal/profiles.md: Suggest the `CHERE_INVOKING` environment v…

lhmouse:
…ariable when MSYS2 bash is to be launched within vscode


Tyriar:
Thanks @lhmouse! I'll also bring this change over to the built-in profile so if it's automatically detected it'll include this variable.

ntrogh (approved):

ntrogh:
@lhmouse Thanks for contributing to our docs!

lhmouse:
My pleasure.

PR #7017, opened 2024-02-03
Fix typo in release notes for v1.86

mihow:
"while holder [key]" changed to "while holding [key]"


ntrogh (approved):
@mihow Thanks for fixing this and making our content better!

PR #7022, opened 2024-02-06
update linux docs

isidorn:


docs/supporting/faq.md
```diff
@@ -301,9 +301,7 @@ VS Code will no longer provide product updates or security fixes on macOS Mojave
...
-
-If you are unable to upgrade your Linux distribution, the recommended alternative is to use our [web client](/docs/editor/vscode-web.md). If you would like to use the desktop version, then you can download the VS Code release 1.85 from [here](https://code.visualstudio.com/updates/v1_85). Depending on your platform, make sure to disable updates to stay on that version. A good recommendation is to set up the installation with [Portable Mode](/docs/editor/portable.md).
+Starting with VS Code release 1.86.1 (January 2024), VS Code desktop is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, or Ubuntu 20.04. VS Code will still allow users to connect to an OS that is not supported by VS Code (no support for glibc >= 2.28) until February 2025. We hope this will provide the needed time for you and your companies to migrate to newer Linux distributions. VS Code will show the appropriate dialog and banner when connecting to an OS version that is not supported by VS Code.
```
ntrogh, line 304: It says `(no support for glibc >= 2.28)`. Shouldn't this be `(no support for glibc < 2.28)`?

ntrogh: Minor text edits:
- "We hope this will provide the needed time for you.." > better to phrase this as "This provides time for you..."
- "VS Code will show the appropriate dialog and banner when connecting..." -> "VS Code will show a dialog and banner message when connecting..."

ntrogh (commented):
Minor edits.

docs/supporting/faq.md
ntrogh, line 304:
```suggestion
Starting with VS Code release 1.86.1 (January 2024), VS Code desktop is only compatible with Linux distributions based on glibc 2.28 or later, for example, Debian 10, RHEL 8, or Ubuntu 20.04. VS Code will still allow users to connect to an OS that is not supported by VS Code (OS that does not provide glibc >= 2.28 and libstdc++ >= 3.4.25) until February 2025. This allows time for you and your companies to migrate to newer Linux distributions. VS Code will show a dialog and banner message when you connect to an OS version that is not supported by VS Code.
```

ntrogh (changes requested):

docs/supporting/faq.md
isidorn: I think Deepak's suggestion is good to use here.
 
"VS Code will still allow users to connect to an OS that is not supported by VS Code (OS that does not provide glibc >= 2.28 and libstdc++ >= 3.4.25)"

ntrogh (approved):

lramos15 (approved):

PR #7041, opened 2024-02-15
add initial chat docs

isidorn:
Work in progress. I feel like we are at 90%

- [x] Currently I call it "agent". And I plan to do a find/replace to "Chat Participant".  Should we call it "Chat Participant" in docs as well? How about an "AI Extension"? Because in the guide I will talk about chat participants and LM access (which are separate). We already have guides for "Web Extensions", so a guide for an "AI Extensions" aligns with that.
- [x] Do we have anything smart to say about Testing these kind of extensions?
- [x] Should we cover prompt crafting recommendations? Do we have some internal guidelines I could be inspired by? Or skip this?


api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+```typescript
+const cat = vscode.chat.createChatParticipant('cat', handler);
+```
+
+> **Note:**: The Chat and Language Model API is in a [proposed state](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) and we are actively working on adding more functionality. Share your feedback in [this GitHub issue](https://github.com/microsoft/vscode/issues/199908) or create new issues.
+
+### Commands
+
+Chat participants can contribute `/commands`, which are shortcuts to specific functionality provided by the extension. One of the tasks when answering questions is to determine the intent, understanding what you want to do. VS Code can infer that "Create a new workspace with Node.js Express Pug TypeScript" means that you want a new project, but "`@workspace /new` Node.js Express Pug TypeScript" is explicit, concise, and saves typing time. By pressing `/` VS Code will offer a list of registered commands with their description.
```
roblourens, line 48: > VS Code can infer that "Create a new workspace with Node.js Express Pug TypeScript" means that you want a new project

We don't yet do this kind of intent detection by default, as far as I know

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+Using a response stream extensions can respond to user queries with different content types: markdown, images, references, buttons and file trees. For example to generate this response:
+
+![Response from the cat extension that includes code, markdown and a button](images/chat/stream.png)
+
+
+An extension can use the response stream in the following way:
+
+```typescript
+stream.markdown('```typescript \n')
```
roblourens, line 76: I think it would be better to show that you don't _have_ to call this with one line of markdown at a time

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+Orthogonal to chat participants, extensions can contribute `#variables`. Variables start with the `#` syntax and are resolved by an extension that contributes that variable or by VS Code if the variable of built-in (e.g `#file`, `#selection`...). Users explicitly include variables in their prompts, by pressing `#` VS Code will offer a list of registered variables with their description:
+
+![List of variables in chat](images/chat/variables.png)
+
+For example, a C++ extension might contribute a variable `#cpp_context` that would get resolved based on the state of the language service - what C++ version is being used and what C++ programming approach is preferred.
+
+Chat participants should not talk directly to each other, instead extensions can define an API that another chat participant extension depends on. Also, variables are resolved independent of the active chat participant, and thus they can be used as a mechanism to share context between participants. For example, `@workspace` already maintains an index of the current workspace. `@workspace` could contribute a variable `#codebase`, that would allow other extensions to leverage the codebase context in their prompts.
```
roblourens, line 122: > Chat participants should not talk directly to each other, instead extensions can define an API that another chat participant extension depends on

Is this just talking about extension API? If so it seems confusing to call it out in the middle of this section.

roblourens: > `@workspace` could contribute a variable `#codebase`, that would allow other extensions to leverage the codebase context in their prompts.

You can say that it _does_ do this :) and I think we should say that it allows _users_ to use the codebase context in their prompt

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+Variable resolvers can offer multiple length levels for the variable value and VS Code will use the one depending on how many tokens are left in an LLM prompt.
+
+```typescript
+vscode.chat.registerVariable('cat_context', 'Describes the state of mind and version of the cat', {
```
roblourens, line 127: This API will not be finalized initially, if that matters for this doc

isidorn: Thank you. For now that is fine. Let's see where we land on Friday, and then I can make it more explicit in docs if something is finalised and something is not.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+9. Chat input for the user to continue. `description` of `@cat` is used as a placeholder
+
+
+## Chat Extension
+
+The [chat sample](https://github.com/microsoft/vscode-extension-samples/tree/main/chat-agent-sample) can be used as a starting point for your chat extension. It is a simple cat tutor that explains computer science topics using cat metaphors.
+
+![Diagram showing how extension can contribute to chat](images/chat/diagram.png)
+
+To start and register a chat participant you can do the following:
```
ntrogh, line 38: This is an example of where I was missing where this piece of code fits into my extension. Let's say I'm new to building extensions and I want to extend Copilot Chat, what do I do after I've gone through the quickstart for creating an extension?

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+        return [
+            { name: 'teach', description: 'Pick at random a computer science concept then explain it in purfect way of a cat' },
+            { name: 'play', description: 'Do whatever you want, you are a cat after all' }
+        ];
+    }
+};
+```
+
+
+### Response Stream
```
ntrogh, line 66:
```suggestion
### Response stream
```

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+    }
+};
+```
+
+
+> **Tip:** Follow-ups should be written as questions or directions, not just concise commands.
+
+## Variables
+
+Orthogonal to chat participants, extensions can contribute `#variables`. Variables start with the `#` syntax and are resolved by an extension that contributes that variable or by VS Code if the variable of built-in (e.g `#file`, `#selection`...). Users explicitly include variables in their prompts, by pressing `#` VS Code will offer a list of registered variables with their description:
```
ntrogh, line 116: You might want to explain variables first. What purpose do they serve for users of the Chat extension? Are variables something that users can modify or only consume? Why would I define variables in my Chat extension?

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+![List of variables in chat](images/chat/variables.png)
+
+For example, a C++ extension might contribute a variable `#cpp_context` that would get resolved based on the state of the language service - what C++ version is being used and what C++ programming approach is preferred.
+
+Chat participants should not talk directly to each other, instead extensions can define an API that another chat participant extension depends on. Also, variables are resolved independent of the active chat participant, and thus they can be used as a mechanism to share context between participants. For example, `@workspace` already maintains an index of the current workspace. `@workspace` could contribute a variable `#codebase`, that would allow other extensions to leverage the codebase context in their prompts.
+
+Variable resolvers can offer multiple length levels for the variable value and VS Code will use the one depending on how many tokens are left in an LLM prompt.
```
ntrogh, line 124: This section is not clear to me, not being too familiar with tokens and LLMs in general.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+                    level: vscode.ChatVariableLevel.Full,
+                    value: 'I am a playful cat, version 1.3, this version prefer to explain everything using mouse and tail metaphors. I am ' + mood
+                }
+            ]
+        }
+    }
+});
+```
+
+## Language Model
```
ntrogh, line 150:
```suggestion
## Language model
```

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+
+ Language Model is a complimentary API that allows [access to the Language Model](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.languageModels.d.ts). Using this API, if an extension contributes a chat participant it can use the Language Model to process and answer the user query. VS Code passes the exact user prompt to contributed participant, and with this API, participants can conveniently transition those language prompts into specific backend API calls. Usage of this API is not tied to the chat experience, and extensions can get language model access to empower their traditional features. For example, the Rust extension might decide to improve it's rename experience and use the Language Model to offer default names.
```
ntrogh, line 152: Did you mean compl**e**mentary?

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+    new vscode.LanguageModelUserMessage('I want to understand recursion')
+];
+const chatRequest = access.makeChatRequest(craftedPrompt, {}, token);
+```
+
+ Language model response is streaming based, and as the response is coming back we suggest for extensions to report progress back to the user for a smooth experience.
+
+Extensions should responsibly use the model and be aware of rate limiting. VS Code will be transparent to the user regarding how extensions are using language models and how many requests each extension is sending and how that influences their respective quotas.
+
+### Prompt Engineering
```
ntrogh, line 177:
```suggestion
### Prompt engineering
```

api/extension-guides/chat.md
```diff
@@ -192,7 +192,7 @@ Chat extensions should explicitly ask for user consent if they are about to do a
 
 ## Testing your extension
 
-
+The part of the extension that is building prompts and interpreting Language Model responses is deterministic, and can thus be unit tested without language model access. However, interacting and getting responses from the Language Model itself is indeterministic and can not be easily tested. Extensions should not request Language Model access for integration tests due to rate-limitations. Internally, VS Code uses a dedicated non-production Language Model for simulation testing, and we are currently thinking how to provide a scalable Language Model testing solution for extensions that are facing a similar challange.
```
roblourens, line 195: I think "nondeterministic" rather than "indeterministic"

api/extension-guides/chat.md
isidorn: Great points. Fixed

api/extension-guides/chat.md
isidorn: Ok. Moved all into one markdown call

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+                }
+            ]
+        }
+    }
+});
+```
+
+## Use the language model
+
+Language Model is a complementary API that allows [access to the Language Model](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.languageModels.d.ts). Using this API, if an extension contributes a chat participant it can use the Language Model to process and answer the user query. VS Code passes the exact user prompt to contributed participant, and with this API, participants can conveniently transition those language prompts into specific backend API calls. Usage of this API is not tied to the chat experience, and extensions can get language model access to empower their traditional features. For example, the Rust extension might decide to improve it's rename experience and use the Language Model to offer default names.
```
bamurtaugh, line 236:
```suggestion
Language Model is a complementary API that allows [access to the Language Model](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.languageModels.d.ts). Using this API, if an extension contributes a chat participant it can use the Language Model to process and answer the user query. VS Code passes the exact user prompt to contributed participant, and with this API, participants can conveniently transition those language prompts into specific backend API calls. Usage of this API is not tied to the chat experience, and extensions can get language model access to empower their traditional features. For example, the Rust extension might decide to improve its rename experience and use the Language Model to offer default names.
```

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+## Testing your extension
+
+The part of the extension that is building prompts and interpreting Language Model responses is deterministic, and can thus be unit tested without language model access. However, interacting and getting responses from the Language Model itself is nondeterministic and can not be easily tested. Extensions should not request Language Model access for integration tests due to rate-limitations. Internally, VS Code uses a dedicated non-production Language Model for simulation testing, and we are currently thinking how to provide a scalable Language Model testing solution for extensions that are facing a similar challenge.
+
+## Publishing your extension
+
+Once you have created your AI extension and once we finalize the Chat and Language Model API (expected early April 2024), you can publish your extension to the Visual Studio Marketplace:
+
+- By publishing to the VS Marketplace your extension is adhering to the [GitHub Copilot extensibility acceptable development and use policy](TODO@isidorn add link).
+- Update the attributes in the `package.json` to make it easy for users to find your extension set the Category to "Chat" in your `package.json`.
```
bamurtaugh, line 289:
```suggestion
- Update the attributes in the `package.json` to make it easy for users to find your extension. Set the Category to "Chat" in your `package.json`.
```

bamurtaugh (commented):
This was a fantastic read, thanks for putting it together!

api/extension-guides/chat.md
isidorn: Good suggestoin. Applied. Thank you

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+You can further expand the functionality of the chat extension with the following optional features:
+
+- Register chat commands to provide users with a shorthand notation for common questions
+- Define suggested follow-up questions to help the user continue a conversation
+- Define chat variables to capture and share domain-specific context in chat conversations
+
+As a starting point for developing a chat extension, you can refer to our [chat extension sample](https://github.com/microsoft/vscode-extension-samples/tree/main/chat-agent-sample). This sample implements a simple cat tutor that can explain computer science topics using cat metaphors.
+
+<!-- TODO: update diagram to distinguish between optional and required -->
+![Diagram showing how extension can contribute to chat](images/chat/diagram.png)
```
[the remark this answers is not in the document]
isidorn: Thank you. Will fix to interpret.
We will always say "Language Model". It is a more general term. There are also "small language models" like Phi https://www.microsoft.com/en-us/research/blog/phi-2-the-surprising-power-of-small-language-models/
So we want to have a more general term to be flexible in the future.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+
+- TODO: add links to the APIs
+
+## Parts of the chat user experience
+
+<!-- TODO: reduce image size -->
+
+![Chat concepts explanation](images/chat/chat.png)
+
+This screenshot shows the following chat concepts with a sample extension:
```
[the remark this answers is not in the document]
isidorn: Awesome. Glad you found it helpful. I am not in love with the colour yellow, thinking of changing it :)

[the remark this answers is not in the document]
isidorn: Good point. Applied.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,204 @@
...
+1. `@cat` explicitly being called by the user via the `@` syntax
+2. Command explicitly called via the `/` syntax
+3. User query
+4. Copilot using the contributed `@cat` participant
```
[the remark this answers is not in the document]
isidorn: We are no longer using agents. Thus making this as resolved.

api/extension-guides/chat.md
isidorn: I think you have tackled this one. Thus marking as resolved.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+ContentId: ac3f00c8-78a8-408c-8af6-3e997a482972
+DateApproved: 02/1/2024
+
+# Summarize the whole topic in less than 300 characters for SEO purpose
+MetaDescription: A guide to creating an AI extension in Visual Studio Code
+---
+
+# Chat extensions
+
+Visual Studio Code's Copilot Chat architecture allows extension authors to integrate with the Copilot Chat experience. The unit of extensibility for Copilot Chat is an agent.
```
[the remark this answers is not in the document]
isidorn: Good point. Tackled.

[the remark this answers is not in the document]
isidorn: We got rid of the term "agent".
We now just use "chat participant"

@ntrogh I leave it up to you if you want us to bold "chat participant" every time we use it.
Both works for me.

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+
+Agents are domain experts that can answer the user query however they want - by fully using AI in the query processing or in a traditional way by forwarding it to a backend service. Agents can also provide the language model access to domain specific tools. With the help of the LLM, the agent might select a tool and define how to invoke it. VS Code ships with some built-in agents, for example `@workspace`. The `@workspace` agent knows about your workspace and can answer questions about it. Internally, the agent is powered by different tools: GitHub's knowledge graph combined with semantic search, local code indexes, and VS Code's language services.
```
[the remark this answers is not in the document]
isidorn: Good point. Removed that term.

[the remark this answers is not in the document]
isidorn: This is tackled. We now always use "Language Model". Thanks

api/extension-guides/chat.md
```diff
@@ -0,0 +1,290 @@
...
+
+> **Note:**: The Chat and Language Model API is in a [proposed state](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) and we are actively working on adding more functionality. Share your feedback in [this GitHub issue](https://github.com/microsoft/vscode/issues/199908) or create new issues.
```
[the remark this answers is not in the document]
isidorn: Good catch. Fixed.

api/extension-guides/chat.md
isidorn: You are right. I changed "can" -> "could". 
Let me know if you still think we might be over-selling here.

api/extension-guides/chat.md
isidorn: I added additional context for variables. Could you re-word this to tackle you concerns please?

api/extension-guides/chat.md
isidorn: This is removed. SHould be better now.

api/extension-guides/chat.md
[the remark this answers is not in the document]
ntrogh: Guidance is that we use italics for introducing new terms and is used only on the first occurrence. Bold is used for referring to UI elements. Will update accordingly.

PR #7043, opened 2024-02-18
Working without Tabs -> Simulating Buffers section

aguynamedben:
⚠️ This depends on new option `moveToActiveGroupIfOpen` which is being tracked in microsoft/vscode#204942 and has PR in microsoft/vscode#205442.

Adds section to Get Started -> Working without Tabs called "Simulating Buffers" which helps users coming from buffer-centric editors such as Emacs and vi. These users are accustomed to a model where any open file ("buffer") can be easily viewed in any split, without having to re-open the file a 2nd time. These users want to split window but don't wish to care about which Editor is in which Editor Group. `revealIfOpen` + `moveToActiveGroupIfOpen` enables that functionality—if a user opens a file that's already open, instead of just surfacing the Editor within it's previous Editor Group, the Editor will be moved into the active Editor Group and then revealed. This means users can use the Quick Open to jump to files from any split without inadvertently opening it multiple times in multiple editor groups.

### DIFF
--- a/docs/getstarted/userinterface.md
+++ b/docs/getstarted/userinterface.md
@@ -391,11 +391,11 @@ If you're coming from buffer-centric editors such as Emacs or vi, you can emulat
 ```json
 "workbench.editor.showTabs": "single",
 "workbench.editor.revealIfOpen": true,
-"workbench.editor.revealIfOpenInActiveGroup": true,
+"workbench.editor.moveToActiveGroupIfOpen": true,
 "workbench.editor.closeEmptyGroups": false,
 ```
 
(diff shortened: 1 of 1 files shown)


anthonyspencer88 (approved):

jokicdanijel:
[sprach agent translation_database.txt](https://github.com/user-attachments/files/20581810/sprach.agent.translation_database.txt)

jokicdanijel:
gobuster!

onyorset (approved):

Siyler1 (approved):

WWJ30 (approved):

jokicdanijel (approved):

sadhu89 (approved):

danhl8877-pixel (approved):

PR #7048, opened 2024-02-22
Fix manual .NET installation for Windows example

FinOCE:
The colon was missing from the path meaning when copy/pasted the snippet would not work


ntrogh (approved):
Nice one! Thanks for contributing and improving our docs, @FinOCE.

PR #7051, opened 2024-02-23
Update the bundle file location for esbuild scripts

yanxiaodi:
This is to fix the discrepancy of the bundle scripts. In other sections, e.g. `Using webpack`, the doc mainly uses `dist/extension.js` as the bundle file path. However, the script for esbuild uses `out/main.js`, which may cause confusions. 

I also tried to create a new project using the latest generator. It seems that the `package.json` uses `"main": "./out/extension.js",` by default, instead of `dist/extension.js`. This may cause more issues due to discrepancies.

It would be great to use an unified bundle path in all docs.


ntrogh (approved):
Thanks for your contribution, @yanxiaodi!

PR #7075, opened 2024-02-27
Update chat docs to use #cpp instead of #cpp_context

benmcmorran:
After discussion with the C++ team, we're planning to use a `#cpp` name instead of the more verbose `#cpp_context` name. The VS Code docs don't necessarily need to match, but it would be nice if they did.

cc @isidorn @lukka @esweet431 @lutzroeder

### DIFF
--- a/api/extension-guides/chat.md
+++ b/api/extension-guides/chat.md
@@ -238,7 +238,7 @@ cat.followupProvider = {
 
 ## Variables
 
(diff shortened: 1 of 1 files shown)


isidorn:
Looks good, thank you!

@ntrogh is it ok to merge this in `main`, or should we merge this against `vnext`?

dbaeumer (approved):

ntrogh:
@isidorn it can be merged into main. It will be released later today.

PR #7085, opened 2024-03-04
fix endregion comment

gnumoksha:


ntrogh (approved):
@gnumoksha Thanks contributing!

PR #7099, opened 2024-03-07
Data wrangler update ToC Next Steps heading

jmew:


docs/datascience/images/data-wrangler-quick-start/view-mode-ui.png
ntrogh, line 1: The updated images have a large white block. Is that intentional?

jmew: Huh seems like there was a glitch in the image.  Just pushed a new commit with it fixed.

ntrogh (approved):

PR #7100, opened 2024-03-07
identifiers: Add OCaml

kevinji:
This identifier is used by the [OCaml extension](https://github.com/ocamllabs/vscode-ocaml-platform).


ntrogh (approved):
Thanks @kevinji !

PR #7102, opened 2024-03-08
Remove extra commas

green961:


ntrogh (approved):
That's a good catch! Thanks @green961

PR #7111, opened 2024-03-12
fix: backslash to escape special char in markdown table

Balastrong:
I think a backslash is missing on this page.

Before:
image

After:
image


ntrogh (approved):
Yes! Thanks for catching and updating this in our docs, @Balastrong !

PR #7114, opened 2024-03-12
Modify a different path for admins on Windows with OpenSSH Server

### LINKED ISSUE #7107: [SSH tips] Scripts to authorize a machine to connect don't work for administrators on Windows OpenSSH host
Scripts [here](https://github.com/microsoft/vscode-docs/blob/main/docs/remote/troubleshooting.md?plain=1#L71) and [here](https://github.com/microsoft/vscode-docs/blob/main/docs/remote/troubleshooting.md?plain=1#L95) add the public key to `~/.ssh/authorized_keys` on the host. However, [the path used if the user belongs to administrator group is different for the Windows OpenSSH server](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_server_configuration#authorizedkeysfile) [that you recommend](https://github.com/microsoft/vscode-docs/blob/main/docs/remote/troubleshooting.md?plain=1#L428).

With the default file modified instead, `ssh` kept asking for password. This fallback is confusing, so it should also be recommended to have the client exclusively use the just-generated identity by adding the following to `~/.ssh/config`:
```
  IdentityFile ~/.ssh/id_ed25519
  IdentitiesOnly yes
```

pawelszramowski:
Fixes #7107.

I shortly investigated programmatic discovery of authorized keys file path but it doesn't seem easily possible (most importantly,
`authorizedkeysfile` is not shown by `ssh -G`) so let's leave the choice to the users.


pawelszramowski:

docs/remote/troubleshooting.md
```diff
@@ -70,6 +70,19 @@ Run one of the following commands, in a **local terminal window** replacing user
 
 * Connecting to a **Windows** SSH host:
 
+  * The host uses OpenSSH Server and the user belongs to the administrator group:
+
+    [A different host path is used then.](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_server_configuration#authorizedkeysfile)
```
bamurtaugh, line 75:
```suggestion
    [A different host path is used](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_server_configuration#authorizedkeysfile):
```

bamurtaugh:
Thanks so much for the contribution @pawelszramowski! Left one nit formatting suggestion.

cc @roblourens to ensure this looks good to you too

pawelszramowski:
@roblourens ping

bhavyaus (approved):

PR #7116, opened 2024-03-13
update `retainContextWhenHidden` docs

gusalbukrk:
Some parts of the docs state that webviews cannot receive messages when hidden even if `WebviewPanelOptions.retainContextWhenHidden` is enabled. However, that has been changed by [this commit](https://github.com/microsoft/vscode/commit/c2b02df9ab208b025b214599660ee86027ca7df9).

Futhermore, the same parts of the docs also state that webview's scripts are suspended when webview is hidden even when `retainContextWhenHidden` is enabled. Which, according to my tests, is not the case. For instance, in the Coding Cat sample, the counter keeps incrementing when tab is in the background if `retainContextWhenHidden` is enabled.


gusalbukrk:

greyaxie (changes requested):
how can i change

api/extension-guides/webview.md
```diff
@@ -1121,9 +1121,9 @@ function getWebviewContent() {
 }
 ```
 
-![persistence retrain](images/webview/persistence-retrain.gif)
+![retainContextWhenHidden demo](images/webview/retainContextWhenHidden.gif)
 
-Notice how the counter does not reset now when the webview is hidden and then restored. No extra code required! With `retainContextWhenHidden`, the webview acts similarly to a background tab in a web browser. Scripts and other dynamic content are suspended, but immediately resumed once the webview becomes visible again. You cannot send messages to a hidden webview, even when `retainContextWhenHidden` is enabled.
+Notice how the counter does not reset now when the webview is hidden and then restored. No extra code required! With `retainContextWhenHidden`, the webview acts similarly to a background tab in a web browser. Scripts and other dynamic content keep running even when tab is not active or visible. You can also send messages to a hidden webview when `retainContextWhenHidden` is enabled.
```
ntrogh, line 1126:
```suggestion
Notice how the counter does not reset now when the webview is hidden and then restored. No extra code required! With `retainContextWhenHidden`, the webview acts similarly to a background tab in a web browser. Scripts and other dynamic content keep running even when the tab is not active or visible. You can also send messages to a hidden webview when `retainContextWhenHidden` is enabled.
```

ntrogh (commented):
@gusalbukrk thanks for suggesting these docs updates! Just a nit in the text.

Adding @mjbvz to verify the updates.

mjbvz (approved):

PR #7119, opened 2024-03-14
update links

chrisdias:
got a report the links moved for the sample and the proposed d.ts files.


ntrogh (approved):
@chrisdias we also moved to use _chat participant_ terminology instead of _agent_. Also, _Agent API_ is now _Chat Participant API_. Want me to do another pass?

greyaxie (commented):
tell me the truet

greyaxie (commented):
ok

PR #7125, opened 2024-03-20
docs: update `glob` usage in testing extension

yoshinorin:
The `glob` library used in the testing example docs has been updated to use `promise` instead of `callback` starting from version `v9.0.0`. This PR will update the documentation for these changes.

https://github.com/isaacs/node-glob/blob/v9/changelog.md#90

Thanks :)


ntrogh (approved):
@yoshinorin thanks for updating this in our docs.

PR #7127, opened 2024-03-20
Explain onView activation event is not emit by built in views

grgar:
onView is only emitted by custom views, per microsoft/vscode#176805. It is additionally confusing that the documentation gives a couple of examples that cannot be applied because they are built-in views.


ntrogh (approved):
Thanks for contributing, @grgar.

PR #7134, opened 2024-03-21
Document OSC 1337 ; CurrentDir and 633 E nonce

Tyriar:
Fixes #7130
Fixes #7131

PR #7139, opened 2024-03-22
Adopt kb for a couple of shell integration keybindings

Tyriar:
Follow up from #7134


ntrogh (approved):
Thanks for the update!

PR #7141, opened 2024-03-22
hundreds->thousands of extensions

pfusik:


ntrogh (approved):
Good one! Thanks for contributing @pfusik.

PR #7155, opened 2024-03-27
Fix action name in task document

tsq:
**Quick open** is used to open a file instead of executing a command, so I think here is a typo and it should be **Command Palette**.


ntrogh:
@tsq thanks for contributing. Quick open is used for finding and opening files, but can also be used for running a task. If you type "task" and then press space, the list of tasks comes up.

tsq:
@ntrogh I'm so sorry, you are right. Thank you for having made me learn this usage.

PR #7158, opened 2024-03-27
Remove the advise on setting `explorer.openEditors.visible` to 0 as it is no longer possible

m1kola:
`explorer.openEditors.visible` no longer allows 0

It was changed in https://github.com/microsoft/vscode/pull/181633

### DIFF
--- a/docs/getstarted/userinterface.md
+++ b/docs/getstarted/userinterface.md
@@ -414,7 +414,3 @@ For example, to make the indent guides bright blue, add the following to your `s
     "editorIndentGuide.background": "#0000ff"
 }
 ```
-
-### Can I hide the OPEN EDITORS section in the Explorer?
-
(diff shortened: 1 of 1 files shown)


ntrogh (changes requested):
Thanks for contributing, @m1kola. Actually, you can hide the OPEN EDITORS view by using the Views menu in the Explorer.

image

m1kola:
@ntrogh thanks for pointing that out. I know that. However it doesn't change the fact that `explorer.openEditors.visible` no longer accepts `0` and the advice in docs is incorrect.

ntrogh:
@m1kola you're indeed correct that the zero value is no longer accepted and the current docs advice is incorrect. My suggestion was that there is a way to hide the open editors by using the menu option. Hence why I suggest to not completely remove the answer from the article, but to update the answer with those instructions.

m1kola:
@ntrogh thanks for the clarification. Unfortunately I currently do not have enough time to own a bigger change.

ntrogh:
@m1kola thanks for raising this. I'll merge this PR and will follow-up myself with another PR.

PR #7163, opened 2024-03-27
add terminal chat section

### LINKED ISSUE #7159: Add docs for terminal chat
https://code.visualstudio.com/docs/copilot/copilot-chat

meganrogge:
fixes [#7159](https://github.com/microsoft/vscode-docs/issues/7159)

### DIFF
--- a/docs/copilot/copilot-chat.md
+++ b/docs/copilot/copilot-chat.md
@@ -59,7 +59,7 @@ The following example shows the `preview` mode for inline chat.
 
 ## Experimental terminal inline chat
 
-Inline chat for the terminal can be enbled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
+Inline chat for the terminal can be enabled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
 
 The chat will use the ![@terminal](###terminal) chat participant, which has context about the integrated terminal shell and its contents.
 
--- a/docs/copilot/copilot-chat.md
+++ b/docs/copilot/copilot-chat.md
@@ -61,7 +61,7 @@ The following example shows the `preview` mode for inline chat.
 
 Inline chat for the terminal can be enbled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
 
-The chat will use the `@terminal` chat participant, which has context about the integrated terminal shell and its contents.
+The chat will use the ![@terminal](###terminal) chat participant, which has context about the integrated terminal shell and its contents.
 
 ## Chat view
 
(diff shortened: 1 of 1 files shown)


docs/copilot/copilot-chat.md
```diff
@@ -57,6 +57,12 @@ The following example shows the `preview` mode for inline chat.
 
 ![Copilot inline chat preview mode, showing the code changes side-by-side](images/copilot-chat/inline-chat-preview-mode.png)
 
+## Experimental terminal inline chat
+
+Inline chat for the terminal can be enbled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
+
+The chat will use the ![@terminal](###terminal) chat participant, which has context about the integrated terminal shell and its contents.
```
meganrogge, line 64: I wanted to do `![`@terminal`]` here but it didn't work in the preview...

Tyriar: The `!` prefix means it's an image, that's probably what was causing the issue.

docs/copilot/copilot-chat.md
```diff
@@ -57,6 +57,12 @@ The following example shows the `preview` mode for inline chat.
 
 ![Copilot inline chat preview mode, showing the code changes side-by-side](images/copilot-chat/inline-chat-preview-mode.png)
 
+## Experimental terminal inline chat
+
+Inline chat for the terminal can be enbled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
```
Tyriar, line 62:
```suggestion
Inline chat for the terminal can be enabled with `terminal.integrated.experimentalInlineChat` and invoked in a terminal (`kb(inlinechat.start)`).
```

docs/copilot/copilot-chat.md
Tyriar, line 64: I think the `#` fragment is wrong here

```suggestion
The chat will use the [`@terminal`](#_terminal) chat participant, which has context about the integrated terminal shell and its contents.
```

docs/copilot/copilot-chat.md
```diff
@@ -57,6 +57,12 @@ The following example shows the `preview` mode for inline chat.
 
 ![Copilot inline chat preview mode, showing the code changes side-by-side](images/copilot-chat/inline-chat-preview-mode.png)
 
+## Experimental terminal inline chat
```
Tyriar, line 60: While this is closely related to inline chat, it feels wrong to feature this before the chat view. @ntrogh any thoughts on where we should stick this? We could also hold off on adding a section until it's non-experimental.

Tyriar (changes requested):

docs/copilot/copilot-chat.md
meganrogge: Yeah I agree. I wasn't sure about this.

ntrogh: We typically hold off on putting experimental features in the main docs.

PR #7166, opened 2024-03-29
Add dev tunnels policy notes

bamurtaugh:


connor4312:
We should probably also change/remove the sections on those pages staring with "If you're part of an organization who wants to control access to..."

bamurtaugh:
Great catch @connor4312! Just removed both of those sections.

docs/editor/port-forwarding.md
```diff
@@ -45,10 +45,12 @@ By default, both hosting and connecting to a tunnel requires authentication with
 
 However, if you've opened a **Public** port, any user with your link can access the forwarded service. You should be careful to avoid hosting any confidential information or insecure services over such ports.
 
-If you're part of an organization that wants to control access to port forwarding, you can do so by allowing or denying access to the domain `global.rel.tunnels.api.visualstudio.com`.
-
```
connor4312, line 49: I think we should keep this info but group it down to the later section since group policies only apply to users running Windows devices: this is still useful if orgs prefer to firewall as a policy

bamurtaugh: Made the change based on your suggestion (added back the statement and just grouped it under the new policy section). Also added mention that group policies are for Windows. LMK what you think @connor4312!

connor4312 (approved):

curib (approved):

PR #7184, opened 2024-04-05
[WIP] Data Viewer deprecation and Data Wrangler GA doc updates

jmew:
Updated docs to remove the mention of the deprecated Data Viewer and added content for Data Wrangler in its place.

Only merge in after the deprecation goes into VS Code stable in May


Mauricionew (commented):
U

PR #7187, opened 2024-04-08
correct the description of subprocess default setting

xlbljz:


ntrogh:
@xlbljz Thanks for contributing. The default value of the `subProcess` setting hasn't changed, so closing this PR.

For reference: https://github.com/microsoft/vscode-python-debugger/blob/1a88f203a4a7ada221bbe4a988a32979045e2def/package.json#L263

PR #7192, opened 2024-04-09
docs(api): Update your-first-extension.md

RodSarhan:
Added a clarification of which command should be ran to start debugging the extension instead of only mentioning the key shortcut.

Added a hint to guide the reader to check if the vscode engine version is compatible with their installed version of VS Code.


ntrogh (approved):
Thanks for the additional clarifications, @RodSarhan.

PR #7199, opened 2024-04-14
Reword documentation to make usage of `publisher` as identifier more explicit

ViToni:
The docs sometimes state "publisher name" when instead it should be "publisher ID" / "publisher identifier".
Especially for someone who performs a setup for the first time, the distinction needs to be clearer.


ViToni:

ntrogh (approved):
Thanks for contributing, @ViToni.

PR #7202, opened 2024-04-15
Fix formatting of symbol server options

EronWright:
Fixes a formatting problem on this page:
https://code.visualstudio.com/docs/csharp/debugging


ntrogh (approved):
Nice one! Thanks for fixing this, @EronWright

PR #7214, opened 2024-04-20
DOC: Debug notebooks uses settings, not launch.json configurations.

Jacob-Stevens-Haas:
A few people have been asking about which  launch.json configuration notebook debugging picks up. [e.g.](https://github.com/microsoft/vscode-jupyter/discussions/13394).

Across debugging, enough people ask about justMyCode that it is mentioned now in a tooltip when you skip a frame.  But there's nothing in the notebook docs about how to set it, which is different from the standard launch.json.   I assume since [debugging tests](https://code.visualstudio.com/docs/python/testing#_debug-tests) included this tidbit, at least something minimal would be desirable here too.


ntrogh (approved):
@Jacob-Stevens-Haas Thanks for adding this clarification.

PR #7232, opened 2024-04-29
Improve paragraph introducing Copilot's code completions

johnd99:


ntrogh (approved):
Thanks for the updates. @johnd99

PR #7283, opened 2024-05-13
Update README.md

KPCOFGS:
Fixed a hyperlink. Correct me if I'm wrong, but I do believe now it should redirect users to the correct repo.


ntrogh (approved):
Ouch, that's a good one to catch! Thanks for your contribution, @KPCOFGS

Issue #7001, opened 2024-02-01
API doc generator mishandles linking for `QuickPick<T>` and friends


gjsjohnmurray:
1. Go to https://code.visualstudio.com/api/references/vscode-api
2. Search in the page for the text `createQuickPick` to find the doc for this function of the `vscode.window` object.
3. Observe that the return type `QuickPick<T>` is marked up as two hyperlinks:

![image](https://github.com/microsoft/vscode-docs/assets/6726799/bc063a72-57ff-4425-8b8e-1c85847ed8b5)

4. Click on the first, which seeks a non-existent in-page anchor `QuickPick` so nothing happens.
5. Similarly the second one seeks anchor `T` that doesn't exist.

I guess this is a result of the API generator not treating `FooBar<T>` as a special case that should be wrapped in a single link to `#FooBar&lt;T&gt;` instead of creating the two links that are appropriate for non-T cases such as the return type of `provideRenameEdits`, which is `ProviderResult<WorkspaceEdit>`.

Fixing this would significantly improve API navigation for extension developers.

gjsjohnmurray:
This seems to have been fixed.

PR #7305, opened 2024-05-21
fix: non-existent hyperlinks (#7001)

1chooo:
I have modified the markup to wrap the entire generic type `QuickPick<T>` in a single hyperlink

This ensures that the link correctly navigates to the appropriate section for `QuickPick<T>`, rather than trying to separately link to `QuickPick` and `T`, which don't have individual anchors in this context. This modification will indeed make the API documentation much more user-friendly.

fixed: https://github.com/microsoft/vscode-docs/issues/7001

### DIFF
--- a/api/references/vscode-api.md
+++ b/api/references/vscode-api.md
@@ -1963,7 +1963,7 @@ when <a href="#window.showInputBox">window.showInputBox</a> does not offer the r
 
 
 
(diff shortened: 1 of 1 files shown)


1chooo:
@ntrogh Could you please review this pull request? If there are no issues, could you merge it and allow contributions to the vscode-docs?

Thanks!

ntrogh:
@1chooo Appreciate your contribution. This article is generated from the VS Code API so these updates will be overwritten with the next release. We have a [related issue ](https://github.com/microsoft/vscode-docs/issues/7001) to fix this.

PR #7308, opened 2024-05-22
Update c-cpp-properties-schema-reference.md

imchuncai:
Change property dotconfig to dotConfig.


ntrogh (approved):
Thanks for the contribution, @imchuncai.

PR #7313, opened 2024-05-27
Document the fourth param to vscode.diff

rictic:
This being missing from the docs cost me an hour or two! I only found it when I searched for other code using the API and found that about half of callers passed this fourth option.

While it's called columnOrOptions, I've only been able to get it to work with a vscode.TextDocumentShowOptions.

### DIFF
--- a/api/references/commands.md
+++ b/api/references/commands.md
@@ -268,7 +268,7 @@ let success = await commands.executeCommand('vscode.openFolder', uri);
 * _left_ - Left-hand side resource of the diff editor
 * _right_ - Right-hand side resource of the diff editor
 * _title_ - Human readable title for the diff editor
-* _options_ - (optional) Editor options. See vscode.TextDocumentShowOptions
+* _options_ - (optional) Either the column in which to open, or editor options (see vscode.TextDocumentShowOptions)
 
 `vscode.prepareTypeHierarchy` - Prepare type hierarchy at a position inside a document
 
--- a/api/references/commands.md
+++ b/api/references/commands.md
@@ -268,6 +268,7 @@ let success = await commands.executeCommand('vscode.openFolder', uri);
 * _left_ - Left-hand side resource of the diff editor
 * _right_ - Right-hand side resource of the diff editor
 * _title_ - Human readable title for the diff editor
+* _options_ - (optional) Editor options. See vscode.TextDocumentShowOptions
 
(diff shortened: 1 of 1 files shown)


api/references/commands.md
```diff
@@ -268,6 +268,7 @@ let success = await commands.executeCommand('vscode.openFolder', uri);
 * _left_ - Left-hand side resource of the diff editor
 * _right_ - Right-hand side resource of the diff editor
 * _title_ - Human readable title for the diff editor
+* _options_ - (optional) Editor options. See vscode.TextDocumentShowOptions
```
ntrogh, line 271:
```suggestion
* _options_ - (optional) Either the column in which to open, or editor options (see vscode.TextDocumentShowOptions)
```

ntrogh (changes requested):
@rictic thanks for contributing and making our docs better for others. I've added some extra info to the parameter description.

api/references/commands.md
rictic: Happy to apply this, but FTR when I tried passing a ViewColumn like `vscode.ViewColumn.One` I got an error, reading something like "expected a column but got 1" which is a little surprising, since vscode.ViewColumn is a numeric enum.

rictic: In any case, I've checked the box to allow edits by maintainers, so feel free to edit into a form you like and merge.

ntrogh: > Happy to apply this, but FTR when I tried passing a ViewColumn like vscode.ViewColumn.One I got an error, reading something like "expected a column but got 1" which is a little surprising, since vscode.ViewColumn is a numeric enum.

Here's an example from our [API tests](https://github.com/microsoft/vscode/blob/e9a1af2326c28b3bc0d40ab7abaed2ce24941e57/extensions/vscode-api-tests/src/singlefolder-tests/window.test.ts#L432C3-L432C126):

'''ts
await commands.executeCommand('vscode.diff', leftDiff, rightDiff, 'Diff', { viewColumn: ViewColumn.Four, preview: false });
'''

rictic: Yeah, can confirm passing an options bag works. Passing just the view column as the fifth arg was where I got an error

ntrogh (approved):

ulugbekna (approved):

PR #7321, opened 2024-05-28
Update contribution-points.md - typo

apupier:


ntrogh (approved):
Thanks for the fix, @apupier

PR #7322, opened 2024-05-28
Specify extensions don't sync for remote

ntrogh:
Related to https://github.com/microsoft/vscode-docs/pull/7319

### DIFF
--- a/docs/editor/extension-marketplace.md
+++ b/docs/editor/extension-marketplace.md
@@ -352,3 +352,9 @@ The Marketplace also provides you with resources to make an informed decision ab
 * **Issues, Repository, and License** - Check if the publisher has provided these and if they have the support you expect.
 
 If you do see an extension that looks suspicious, you can report the extension to the Marketplace with the **Report Abuse** link at the bottom of the extension **More Info** section.
+
+### My extensions don't synchronize when connected to a remote window
+
+[Settings Sync](/docs/editor/settings-sync.md) lets you share your Visual Studio Code configurations such as settings, keybindings, and installed extensions across your machines so you are always working with your favorite setup.
+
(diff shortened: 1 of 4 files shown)


docs/editor/extension-marketplace.md
```diff
@@ -352,3 +352,9 @@ The Marketplace also provides you with resources to make an informed decision ab
 * **Issues, Repository, and License** - Check if the publisher has provided these and if they have the support you expect.
 
 If you do see an extension that looks suspicious, you can report the extension to the Marketplace with the **Report Abuse** link at the bottom of the extension **More Info** section.
+
+### My extensions don't synchronize when connected to a remote window
```
thernstig, line 356: Should this not say "remote" and not "remote window", to keep the terminology in the rest of the docs e.g. https://code.visualstudio.com/docs/remote/remote-overview?

I realize the text further below specifies this better, but I have not seen the term "remote window" used anywhere else. Just "remote".

(Same comment in the body of the text below, if you find this applicable, and in all other doc changes in this change)

ntrogh: Good point. There are a few instances where we mention _remote window_ and it's also mentioned in the hover text in VS Code. I acknowledge that we don't clearly define the terminology and use it consistently (for example, we also have occurrences of "remote context"). I'll create a follow-up issue to review our docs. Thanks for pointing this out, @thernstig!

image

thernstig: Interesting! I had not contemplated that `Remote Window` part. I wonder if that text should also in reality say just "Remote". The window is always local so to say.

You are the the expert in this area though, so I have no better input on if "Remote Window" makes sense in some cases or not.

ntrogh: This is a subtle one. We need to aim for consistency and clarity, so there's no confusion. I'll take it up with the team and definitely also appreciate your input. Keep it coming!

PR #7332, opened 2024-06-01
docs: correct instance of incorrect singularization

WellsOfLivingWater:
docs: correct instance of incorrect singularization ('description' -> 'descriptions')


ntrogh (approved):
Good catch! Thanks for fixing this, @WellsOfLivingWater

PR #7335, opened 2024-06-03
corrected a typo

yg-i:
there's no @oudated filter, only @updates


ntrogh (approved):
@yg-i Thanks for the update!

PR #7343, opened 2024-06-04
Update Linux pre-reqs

bamurtaugh:
@roblourens highlighted that we list Python as a pre-req for Remote Development, but I'm not aware of any Python dependencies. I've removed the Python requirement from the table in the Linux pre-reqs section.

I've also added a link to the Docker configuration documentation in the "Dev Containers" section.

Based on the latest info in https://github.com/microsoft/vscode/issues/206790, I think we've made the necessary changes (i.e. glibc >= 2.28 and libstdc++ >= 3.4.25 are the right version requirements, even though online it seems like glibc 2.28 may be out of support as I [only see down to 2.32 in lts docs](https://sourceware.org/glibc/wiki/Release)).

@deepak1556 as you've driven a lot of the server versioning work, I'd love to get your feedback on the couple of changes I've made here and any other changes we might want to consider.


docs/remote/linux.md
```diff
@@ -45,7 +45,7 @@ You may encounter issues with certain extensions with native dependencies with *
 
 | Distribution | Base Requirements | Remote - SSH Requirements | Notes |
 |--------------|-------------------|------------------|-------|
-| General |  kernel >= 4.18, glibc >=2.28, libstdc++ >= 3.4.25, Python 2.6 or 2.7, tar | OpenSSH server, `bash`, and `curl` or `wget` | Run `ldd --version` to check the glibc version. Run `strings /usr/lib64/libstdc++.so.6 \| grep GLIBCXX` to see if libstdc++ 3.4.25 is available. |
+| General |  kernel >= 4.18, glibc >=2.28, libstdc++ >= 3.4.25, tar | OpenSSH server, `bash`, and `curl` or `wget` | Run `ldd --version` to check the glibc version. Run `strings /usr/lib64/libstdc++.so.6 \| grep GLIBCXX` to see if libstdc++ 3.4.25 is available. |
 | General for Arm32 | `libatomic1` | No additional requirements. | |
 | Ubuntu 20.04+, Debian 10+, Raspberry Pi OS Buster/10+ and downstream distributions | `libc6 libstdc++6 python-minimal ca-certificates tar` | `openssh-server bash` and `curl` or `wget` | Requires kernel >= 4.18, glibc >= 2.28, libstdc++ >= 3.4.25. |
```
roblourens, line 50: The other rows mention python too, and I'd think that can be removed too.

bamurtaugh: Good point! Removed.

roblourens (dismissed):

deepak1556 (dismissed):
Sorry for the delay, I am also not aware of python requirement. LGTM

As for glibc 2.28, the minimum version is picked based on what the remote binary which is Node.js binary supports, when they update we will update as necessary. Right now it is 2.28 as of Node.js 22 https://github.com/nodejs/node/blob/main/BUILDING.md#platform-list

bamurtaugh:
Thanks for the feedback and reviews!

@ntrogh this is ready for your review and merge whenever you get a chance.

docs/remote/linux.md
```diff
@@ -35,7 +35,7 @@ If you are running Linux locally, the [VS Code prerequisites](/docs/supporting/r
...
+* **Dev Containers**: Docker CE/EE 18.06+ and Docker Compose 1.21+. Follow the [official install instructions for Docker CE/EE for your distribution](https://docs.docker.com/install/#supported-platforms). If you are using Docker Compose, follow the [Install Docker Compose directions](https://docs.docker.com/compose/install/) as well. (Note that the Ubuntu Snap package is not supported and packages in distributions may be out of date.) `docker` and `docker-compose` must also be in the path. However, Docker does not need to be running if you are [using a remote host](https://aka.ms/vscode-remote/containers/remote-host). You can learn more about ways to configure Docker in the [dev containers doc](/docs/devcontainers/containers.md#system-requirements).
```
ntrogh, line 38:
```suggestion
* **Dev Containers**: Docker CE/EE 18.06+ and Docker Compose 1.21+. Follow the [official install instructions for Docker CE/EE for your distribution](https://docs.docker.com/install/#supported-platforms). If you are using Docker Compose, follow the [Install Docker Compose directions](https://docs.docker.com/compose/install/) as well. (Note that the Ubuntu Snap package is not supported and packages in distributions may be out of date.) `docker` and `docker-compose` must also be in the path. However, Docker does not need to be running if you are [using a remote host](https://aka.ms/vscode-remote/containers/remote-host). You can learn more about ways to configure Docker in the [Dev Containers documentation](/docs/devcontainers/containers.md#system-requirements).
```

ntrogh (commented):
Minor comment

ntrogh (approved):
Thanks for the updates. LGTM!

PR #7345, opened 2024-06-05
explain how to get vscode.d.ts

isidorn:


isidorn:
per @egamma suggestion

PR #7351, opened 2024-06-06
`TextDocument.getLanguageId()` changed to `TextDocument.languageId`

yg-i:
It looks like TextDocument does not have (or no longer has) a `getLanguageId()` method

image


ntrogh (approved):
Indeed, that method got moved into a property. Thanks @yg-i

PR #7357, opened 2024-06-06
docs: release v1.90: change raw link to hyperlink with descriptive text

zaidhaan:
Was a bit surprised by the raw un-clickable link.

image

I reworded it a little and added a hyperlink.

### DIFF
--- a/release-notes/v1_90.md
+++ b/release-notes/v1_90.md
@@ -295,7 +295,7 @@ for await (const part of response.text) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


zaidhaan:
Well, just two paragraphs later there's another raw link!
image

I believe it would make sense if it's linked like this:

> If you're an extension author who is planning to use the chat and language model APIs, consider trying out the latest alpha release of this library: [@vscode/prompt-tsx](https://www.npmjs.com/package/@vscode/prompt-tsx).

So I'll go ahead and make that change too.

ntrogh (approved):
@zaidhaan Appreciate you taking the time to fix this! Those links slipped through the cracks of our review. Adding a note to our release checklist!

PR #7374, opened 2024-06-12
Update logpoints section

ntrogh:


docs/editor/debugging.md
```diff
@@ -174,10 +174,12 @@ Optionally, breakpoints can be shown in the editor's overview ruler by enabling
 
 ## Logpoints
 
-A Logpoint is a variant of a breakpoint that does not "break" into the debugger but instead logs a message to the console. Logpoints are especially useful for injecting logging while debugging production servers that cannot be paused or stopped.
+A Logpoint is a variant of a breakpoint that does not "break" into the debugger but instead logs a message to the debug console. Logpoints enable you to  inject logging while debugging without modifying the source code. Instead of adding a `console.log` statement, you can add a Logpoint during a debugging session and not risk having logging statements left in your code. For example, logpoints are especially useful for when you're debugging production servers that cannot be paused or stopped.
```
Tyriar, line 177:
```suggestion
A Logpoint is a variant of a breakpoint that does not "break" into the debugger but instead logs a message to the debug console. Logpoints enable you to  inject logging while debugging without modifying the source code. Instead of adding a `console.log` statement, you can add a Logpoint during a debugging session and not risk having logging statements left in your code. For example, logpoints are especially useful when you're debugging production servers that cannot be paused or stopped.
```

Tyriar, line 177: > and not risk having logging statements left in your code

A bigger benefit is the time saved adding/removing the statement imo

Tyriar (changes requested):

docs/editor/debugging.md
ntrogh: Good feedback. I'll incorporate it.

Tyriar (approved):

PR #7377, opened 2024-06-13
Update project name section

chrmarti:
Pointing to the more general section on docker.com since we now also support the `name` property in the `docker-compose.yml`.

Mentioning our override to avoid name collisions at the end.

### DIFF
--- a/remote/advancedcontainers/set-docker-compose-project-name.md
+++ b/remote/advancedcontainers/set-docker-compose-project-name.md
@@ -9,9 +9,11 @@ DateApproved: 06/05/2024
 ---
 # Set Docker Compose project name
 
-Visual Studio Code will respect the value of the [COMPOSE_PROJECT_NAME](https://docs.docker.com/compose/reference/envvars/#compose_project_name)  environment variable if set for the VS Code process or in a `.env` file in the root of the project.
+Visual Studio Code will respect the value you configure for the Docker Compose [project name](https://docs.docker.com/compose/project-name/).
 
-For example, after shutting down all VS Code windows, you can start VS Code from the command line as follows:
+For example, when the `COMPOSE_PROJECT_NAME` environment variable is set for the VS Code process or in a `.env` file in the same folder as the `docker-compose.yml`:
+
+After shutting down all VS Code windows, you can start VS Code from the command line as follows:
 
 ```bash
 # from bash
@@ -24,8 +26,10 @@ $env:COMPOSE_PROJECT_NAME=foo
 code .
 ```
 
-Or add the following to a `.env` file in the root of the project (**not** in the `.devcontainer` folder):
+Or add the following to a `.env` file in the same folder as the `docker-compose.yml`:
 
 ```
 COMPOSE_PROJECT_NAME=foo
 ```
+
+When no project name is configured and the `docker-compose.yml` is in the `.devcontainer` folder the Docker Compose default of using the `docker-compose.yml` folder's basename is overridden with `${project-folder-basename}_devcontainer` to avoid name collisions with other projects.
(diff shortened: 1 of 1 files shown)


jrieken (dismissed):

remote/advancedcontainers/set-docker-compose-project-name.md
```diff
@@ -24,8 +26,10 @@ $env:COMPOSE_PROJECT_NAME=foo
 code .
 ```
 
-Or add the following to a `.env` file in the root of the project (**not** in the `.devcontainer` folder):
+Or add the following to a `.env` file in the same folder as the `docker-compose.yml`:
```
ntrogh, line 29:
```suggestion
Alternatively, add the following entry to a `.env` file in the same folder as the `docker-compose.yml`:
```

remote/advancedcontainers/set-docker-compose-project-name.md
```diff
@@ -24,8 +26,10 @@ $env:COMPOSE_PROJECT_NAME=foo
...
 
 ```
 COMPOSE_PROJECT_NAME=foo
 ```
+
+When no project name is configured and the `docker-compose.yml` is in the `.devcontainer` folder the Docker Compose default of using the `docker-compose.yml` folder's basename is overridden with `${project-folder-basename}_devcontainer` to avoid name collisions with other projects.
```
ntrogh, line 35:
```suggestion
When no project name is configured and the `docker-compose.yml` is in the `.devcontainer` folder, the Docker Compose default of using the `docker-compose.yml` folder's basename is overridden with `${project-folder-basename}_devcontainer` to avoid name collisions with other projects.
```

ntrogh (commented):
Added some suggestions to polish the text.

remote/advancedcontainers/set-docker-compose-project-name.md
```diff
@@ -9,9 +9,13 @@ DateApproved: 06/05/2024
 ---
 # Set Docker Compose project name
 
-Visual Studio Code will respect the value of the [COMPOSE_PROJECT_NAME](https://docs.docker.com/compose/reference/envvars/#compose_project_name)  environment variable if set for the VS Code process or in a `.env` file in the root of the project.
+Visual Studio Code will respect the value you configure for the Docker Compose [project name](https://docs.docker.com/compose/project-name/).
 
-For example, after shutting down all VS Code windows, you can start VS Code from the command line as follows:
+You can set the `COMPOSE_PROJECT_NAME` environment variable for the VS Code process, or specify it in a `.env` file in the same folder as the `docker-compose.yml`.
```
chrmarti, line 14: @ntrogh I would also like to mention the name property, what do you think:

```suggestion
The top-level property `name` in the `docker-compose.yml` can be used to set the project name.

Alternatively, you can set the `COMPOSE_PROJECT_NAME` environment variable for the VS Code process, or specify it in a `.env` file in the same folder as the `docker-compose.yml`.
```

chrmarti: I'd like to mention the `name` property in the `docker-compose.yml` because we only recently added support for that on our side and it is probably the easiest way to adjust the project name.

The part about the env variable still makes sense as another way of changing the project name when changing the `name` property is impractical for some reason and because that comes with the additional hurdle of having to restart VS Code.

The env variable use could be considered "advanced" now that we support `name`. We could remove it from the documentation to keep it simple. 🤔

ntrogh: Ah, now it makes sense. Your update looks good.

chrmarti:
@ntrogh Good to merge then? (Requires an approval.) Thanks!

PR #7381, opened 2024-06-14
Fix model names

potatoqualitee:
Keeping `copilot-` in there breaks the extension response and sets the model to `undefined`.

### DIFF
--- a/api/extension-guides/language-model.md
+++ b/api/extension-guides/language-model.md
@@ -71,7 +71,7 @@ const craftedPrompt = [
 
 ## Send the language model request
 
-Once you've built the prompt for the language model, you need to select the language model by specifying the `vendor`, `id`, `family`, or `version` of the model you want to get. Currently, only `copilot-gpt-3.5-turbo` and `copilot-gpt-4` are supported. We expect that the list of supported models will grow over time. Once you have the model, you can send the request to it by using `sendRequest`.
+Once you've built the prompt for the language model, you need to select the language model by specifying the `vendor`, `id`, `family`, or `version` of the model you want to get. Currently, only `gpt-3.5-turbo` and `gpt-4` are supported. We expect that the list of supported models will grow over time. Once you have the model, you can send the request to it by using `sendRequest`.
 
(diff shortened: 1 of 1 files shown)


api/extension-guides/language-model.md
```diff
@@ -71,7 +71,7 @@ const craftedPrompt = [
 
 ## Send the language model request
 
-Once you've built the prompt for the language model, you need to select the language model by specifying the `vendor`, `id`, `family`, or `version` of the model you want to get. Currently, only `copilot-gpt-3.5-turbo` and `copilot-gpt-4` are supported. We expect that the list of supported models will grow over time. Once you have the model, you can send the request to it by using `sendRequest`.
+Once you've built the prompt for the language model, you need to select the language model by specifying the `vendor`, `id`, `family`, or `version` of the model you want to get. Currently, only `gpt-3.5-turbo` and `gpt-4` are supported. We expect that the list of supported models will grow over time. Once you have the model, you can send the request to it by using `sendRequest`.
```
ntrogh, line 74:
```suggestion
Once you've built the prompt for the language model, you need to select the language model by specifying the `vendor`, `id`, `family`, or `version` of the model you want to get. Learn more about these properties in the [API reference](/api/references/vscode-api#LanguageModelChat). Currently, only `gpt-3.5-turbo` and `gpt-4` are supported. We expect that the list of supported models will grow over time. Once you have the model, you can send the request to it by using `sendRequest`.
```

ntrogh: Suggest adding a link to the API reference to get more details about the different properties.

ntrogh (changes requested):
@potatoqualitee Thanks for the PR. I added a link to the API reference for more details. Otherwise, looks good to me.

api/extension-guides/language-model.md
potatoqualitee: Thanks, also suggestion LGTM 😊

ntrogh (approved):
Thanks for the updates.

PR #7387, opened 2024-06-18
GitHub app deservers a separate section

isidorn:
@ntrogh what do you think about this?

Just reading the docs and stumbling upon GitHub apps I found confusing. So I created a new short section, that might grow over time.


ntrogh (approved):
@isidorn LGTM. If this section grows bigger, we should consider moving it lower in the document, as the focus of the current article is the Chat API. For now, it's fine.

PR #7397, opened 2024-06-23
update some extension reporting guidlines

justschen:
- added documentation on contributing menu/issue reporter command

### DIFF
--- a/api/get-started/wrapping-up.md
+++ b/api/get-started/wrapping-up.md
@@ -29,7 +29,7 @@ Users can report an issue for an extension via the `workbench.action.openIssueRe
 
 Extension authors can attach any additional information when users report an issue. 
 
-To let extensions benefit from this in the separate case of navigating the normal issue reporter flow via `Help: Reporter Issues...`, extension authors can contribute a custom command (which will invoke `openIssueReporter`) and a menu contribution point to `issue/reporter`.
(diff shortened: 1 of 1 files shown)


ntrogh (commented):
@justschen I had trouble understanding how the different pieces fit together. I restructured the text a bit to how I think it works. Let me know if I didn't get it right.

It might be good to add a little more context about why you should use this approach versus the previous of contributing a `workbench.action.openIssueReporter` command.

justschen:
@ntrogh gotcha! I like these changes, appreciate it! The main reason we wanted extensions authors to be aware of this was so:
a) extensions can now attach additonal data from the extension -> issue reporting 
b) removes many of the "Exension...: Report Issue.." commands from the command palette (like `Python: Report Issue` or `Azure: Report Issue`). There was a former limitation where authors _had_ to contribute the command to have access to templates and such, but adding the command directly to the palette will no longer be needed with the updates to the contribution point.

Not sure how we can best convey that?

ntrogh:
@justschen I'll try and work something out.

ntrogh:
@justschen I've made some changes to the section to try and better communicate the value prop. Feel free to update.

justschen:
@ntrogh it looks good! thanks :D

DonJayamanne (approved):

Issue #7398, opened 2024-06-24
docker-compose doc references non-existent file


jason-matthew:
Doc consumers are confused by a unique file reference.

Current doc ( [9c405b7350a5adbcb2d633d09b79fcadb0ee8650/docs/devcontainers/create-dev-container.md#L415](https://github.com/microsoft/vscode-docs/blame/9c405b7350a5adbcb2d633d09b79fcadb0ee8650/docs/devcontainers/create-dev-container.md#L415) ) contains a single reference to `devcontainer.extend.yml`.  Similarly named `docker-compose.extend.yml` is referenced 4x in preceding lines.

ntrogh:
@jason-matthew Thanks for raising this.

ntrogh:
@bamurtaugh It looks like `devcontainer.extend.yml` should actually be `.devcontainer/docker-compose.extend.yml` instead. That way, the development settings override the defaults from the `docker-compose.yml` file. Can you confirm?

jason-matthew:
My attempts to stage a doc update has produced https://github.com/microsoft/vscode-docs/pull/7399.  

Apologies for the noise.

ntrogh:
Thanks for creating a PR. Looks good to me, but need @bamurtaugh to review and confirm.

PR #7399, opened 2024-06-24
Fix docker-compose file reference (#7398)

jason-matthew:
Fixes https://github.com/microsoft/vscode-docs/issues/7398


bamurtaugh (approved):
Thank you!

PR #7413, opened 2024-06-26
Typo corrected "cat" > "chat"

kdx-perbol:
Typo corrected "cat" > "chat". 🐱


ntrogh (approved):
@kdx-perbol thanks for contributing! The sample implements a 😺 chat participant but it might be confusing, unless you actually look at the sample code. Appreciate you taking the time to correct this.

PR #7421, opened 2024-06-27
Update linux ssh-agent snippet to include "ssh-add"

bedaberner:
as ssh-add does not persist in linux, the ssh-add command must be executed every time and should therefore be added to the snippet (see [this stackoverflow](https://stackoverflow.com/a/74704066) answer which led me to finding this).

Furthermore I found it annoying that `eval "cat $HOME/.ssh/ssh-agent"` echoes the PID of the ssh-agent and therefore polutes your terminal, therefore I suggest muting it's output.

### DIFF
--- a/remote/advancedcontainers/sharing-git-credentials.md
+++ b/remote/advancedcontainers/sharing-git-credentials.md
@@ -32,7 +32,7 @@ There are some cases when you may be cloning your repository using SSH keys inst
 You can add your local SSH keys to the agent if it is running by using the `ssh-add` command. For example, run this from a terminal or PowerShell:
 
 ```bash
-ssh-add $HOME/.ssh/github_rsa
+ssh-add $HOME/.ssh/<your ssh key>
 ```
 
 On Windows and Linux, you may get an error because the agent is not running (macOS typically has it running by default). Follow these steps to resolve the problem:
@@ -67,9 +67,12 @@ if [ -z "$SSH_AUTH_SOCK" ]; then
         ssh-agent -s &> $HOME/.ssh/ssh-agent
    fi
    eval `cat $HOME/.ssh/ssh-agent` > /dev/null
-   ssh-add 2> /dev/null
+   ssh-add $HOME/.ssh/<your ssh key> 2> /dev/null
 fi
 ```
+On the last line, replace `<your ssh key>` with your specific ssh key.
+
+For example `ssh-add $HOME/.ssh/id_ed25519 2> /dev/null`
 
 If you encounter any issues, you may want to check the Dev Containers extension's [known limitations](/docs/devcontainers/containers.md#known-limitations).
 
--- a/remote/advancedcontainers/sharing-git-credentials.md
+++ b/remote/advancedcontainers/sharing-git-credentials.md
@@ -66,7 +66,8 @@ if [ -z "$SSH_AUTH_SOCK" ]; then
         # Launch a new instance of the agent
         ssh-agent -s &> $HOME/.ssh/ssh-agent
    fi
-   eval `cat $HOME/.ssh/ssh-agent`
+   eval `cat $HOME/.ssh/ssh-agent` > /dev/null
+   ssh-add 2> /dev/null
 fi
 ```
 


bamurtaugh:
Thanks for opening! I'd love to get @roblourens's expertise here.

nooboobhuh (approved):

roblourens:
I'm not sure you want to run `ssh-add` on every login, just after creating a key, right? There's a line above about running it to setup your key.

The point about `eval "cat $HOME/.ssh/ssh-agent"`, I don't know whether it's an issue, I don't have an opinion there.

bedaberner:
I don't have deep knowledge about how `ssh-add` works but i can confirm that on Ubuntu 24.04 (running in wsl2) you have to execute `ssh-add` on every reboot. This is also mentioned in the stack-overflow issue I cited above. This can be confirmed by running `ssh-add -l` which will return an empty list after reboot.

> If we're going to recommend it, it might be better practice to add a particular key rather than running it to add all keys?

I have no preference whatsoever about adding a specific key or all of them. For my use case adding all of them works fine. 
> The point about eval "cat $HOME/.ssh/ssh-agent", I don't know whether it's an issue, I don't have an opinion there.

For me it's definitely an issue if my terminal echoes random stuff every time I open it. Maybe that's just me but I think it should not be recommended behavior, especially since the information provided is not useful.

roblourens:
Right, I think we should hide the output here.

And I think it's fine to run ssh-add here, other recommendations on the web are very scattered obviously, but I think we might as well update it to add one key by name, to match our other example above.

remote/advancedcontainers/sharing-git-credentials.md
```diff
@@ -66,7 +66,8 @@ if [ -z "$SSH_AUTH_SOCK" ]; then
         # Launch a new instance of the agent
         ssh-agent -s &> $HOME/.ssh/ssh-agent
    fi
-   eval `cat $HOME/.ssh/ssh-agent`
+   eval `cat $HOME/.ssh/ssh-agent` > /dev/null
+   ssh-add 2> /dev/null
```
ntrogh, line 70: @bedaberner per suggestion from @roblourens:

```suggestion
   ssh-add $HOME/.ssh/github_rsa 2> /dev/null
```

ntrogh (changes requested):
@bedaberner thanks for contributing. Can you incorporate the suggested updates?

remote/advancedcontainers/sharing-git-credentials.md
bedaberner: I think doing it like this is worse, as it introduces value that you need to change based on your individual configuration hidden inside a snippet 99% of people are just going to copy. Since we mute the output of ssh-add it will not be easily noticable to users that ssh-add could not add the key that does not exist on their system.

Especially because of the wording above, people will just copy the whole snippet.

> Then add these lines to your ~/.bash_profile or ~/.zprofile (for Zsh) so it starts on login:

In this context, just adding all keys is preferable, because it will work without user changes.
maybe we can add a commented version on how to just add a single key?

```suggestion
   ssh-add 2> /dev/null
   # ssh-add ~/.ssh/<your ssh key> 2> /dev/null # use this syntax to add a specific key
```

Regardless if you insist on adding a specific key, we should at least make it obvious that there is user input required by marking the key with <> (which should be done in the example above anyway, seeing as `github_rsa` is not created in the scope of this example) and maybe adding some text.

ntrogh: The default behavior should be "least privilege", so not in favor of only adding a comment to add a specific key.
Agree that the text should be updated to indicate that users have to add their specific key. 
Do you want to update the text, or otherwise you can remove the ssh-add updates from this PR and I'll create a follow-up PR.

nooboobhuh (approved):

nooboobhuh (approved):

nooboobhuh (approved):

remote/advancedcontainers/sharing-git-credentials.md
bedaberner: @ntrogh I've added a commit with some text.
I also changed `github_rsa` to `<your ssh key>` in both snippets to keep things consistent and make clear this requires user input.

remote/advancedcontainers/sharing-git-credentials.md
```diff
@@ -32,7 +32,7 @@ There are some cases when you may be cloning your repository using SSH keys inst
 You can add your local SSH keys to the agent if it is running by using the `ssh-add` command. For example, run this from a terminal or PowerShell:
 
 ```bash
-ssh-add $HOME/.ssh/github_rsa
+ssh-add $HOME/.ssh/<your ssh key>
```
nooboobhuh, line 35:
```suggestion
ssh-add $HOME/.ssh/<SHA256:c3yyMl5VGrtSrX89PKOrexF/Bzl4+u4I+n3JZvy3YJ8>
```

roblourens:
This seems fine to me, thanks

roblourens (approved):

PR #7422, opened 2024-06-27
Remove quote decoration for inline buttons

daviddossett:


daviddossett:
Closing since this actually breaks the layout

PR #7424, opened 2024-06-30
Fix missing close tag

gjsjohnmurray:


ntrogh (approved):
@gjsjohnmurray thanks for the support on the 1.91 release notes! Things are still in flux as the team's adding their content, so I was holding off on my final edit pass.

Issue #7447, opened 2024-07-21
Illustrations are not visible in Dark Mode


sameer-dudeja:
The illustrations on Language extensions docs are not visible in dark mode but are perfect on light mode

![image](https://github.com/user-attachments/assets/a1b3aff8-557d-4684-b009-f10a4c26ed66)

PR #7448, opened 2024-07-21
Fix Illustration Visibility Issue in Dark Mode

sameer-dudeja:
**Added New Constant Images which wont cause issues when changing themes** 

**Issue: ** #7447

The background of the new images is **solid white** instead of **transparent** which was causing the visibility issues.

multi-editor
multi-ls

Tested Visually
Dark Mode Preview 
image


sameer-dudeja:

ntrogh (approved):
Thanks for raising and fixing this, @sameer-dudeja ! Appreciate your contribution.

PR #7450, opened 2024-07-23
Update extension-marketplace.md

seaniyer:
Added info on how to contact VSM support


seaniyer:
@isidorn

isidorn (changes requested):
I like the wording. But I am not sure if this is the right section to add this.
The section you added it is about malicious extensions - and your addition is about help to manage your own extension?

Maybe a better fit would be under https://github.com/microsoft/vscode-docs/blob/main/api/working-with-extensions/publishing-extension.md

Since it is targeting extension authors? What do you think?

seaniyer:
@isidorn
That would indeed be a better location. Thanks for the suggestion. I'll send a separate PR for that.

PR #7493, opened 2024-08-07
tasks.md: add `source` property to `problemMatcher`

wjandrea:
Currently, the `source` property is only mentioned under [Modifying an existing problem matcher](https://code.visualstudio.com/docs/editor/tasks#_modifying-an-existing-problem-matcher), but with no details, and it applies to more than just existing matchers.

It would be great to add it as part of the example as demonstrated in this PR, though it's incomplete: This also requires updating the screenshot below.

---

Related feature request I just opened on the VSCode repo: [QoL improvement: Show task `label` as source of problem if `problemMatcher` doesn't specify a `source`](https://github.com/microsoft/vscode/issues/225072)

### DIFF
--- a/docs/editor/tasks.md
+++ b/docs/editor/tasks.md
@@ -674,6 +674,7 @@ Here is a finished `tasks.json` file with the code above (comments removed) wrap
             "problemMatcher": {
                 "owner": "cpp",
                 "fileLocation": ["relative", "${workspaceFolder}"],
+                "source": "gcc",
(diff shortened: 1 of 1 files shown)


alexkinyu:
hello

alexkinyu (commented):
......

ntrogh (approved):

ntrogh:
@wjandrea Thanks for contributing. I'll update the screenshot.

PR #7494, opened 2024-08-07
tasks.md: mention `problemMatcher.severity`

wjandrea:
Currently, the `severity` property is only mentioned under [Modifying an existing problem matcher](https://code.visualstudio.com/docs/editor/tasks#_modifying-an-existing-problem-matcher), but with no details, and it applies to more than just existing matchers. (cf. #7493)


ntrogh (approved):
@wjandrea Thanks for the update!

PR #7497, opened 2024-08-08
rather append to avoid overriding existing exes in $PATH

Konfekt:
Does adding the MinGW UCRT bin folder path to %PATH% not break existing exes with the same name, say of Git Bash ? Does Add mean appending?


ntrogh:
@Konfekt In this context, with the given steps, "Add" is more appropriate. Thanks for contributing to our docs.

PR #7528, opened 2024-08-29
Remove `new` from `vscode.LanguageModelChatMessage.User`

observerw:
Since it's just a static method:

```typescript
export class LanguageModelChatMessage {
    static User(content: string, name?: string): LanguageModelChatMessage;
}
```


ntrogh (approved):
Thanks for contributing, @observerw

PR #7586, opened 2024-09-13
Problem matcher updates

ntrogh:


docs/editor/images/tasks/problemmatcher.png
Tyriar, line 1: The warning underline in the editor no longer shows? I guess this comes from the extension not the task, but it made the error more obvious in the editor.

Tyriar (changes requested):

docs/editor/images/tasks/problemmatcher.png
ntrogh: Good catch. I updated the screenshot to include the error in the editor.

Tyriar (approved):

PR #7589, opened 2024-09-13
Typo, add missing inline code markup

muzimuzhi:
Add missing inline code markup for context names in api/references/when-clause-contexts.md, and normalize column names.

### DIFF
--- a/api/references/when-clause-contexts.md
+++ b/api/references/when-clause-contexts.md
@@ -257,10 +257,10 @@ Context name | True when
 
 You can have a when clause that checks if a specific [View](/api/ux-guidelines/views) is visible or focused.
 
-Context name | True when
------------- | ------------
+Context name  | True when
+------------- | ----------
 `view.${viewId}.visible` | True when specific view is visible.<br>Example: `"view.workbench.explorer.fileView.visible"`
-focusedView | True when specific view is focused.<br>Example: `"focusedView == 'workbench.explorer.fileView'"`
+`focusedView` | True when specific view is focused.<br>Example: `"focusedView == 'workbench.explorer.fileView'"`
 
 View identifiers:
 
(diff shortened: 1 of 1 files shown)


api/references/when-clause-contexts.md
```diff
@@ -288,11 +288,11 @@ View identifiers:
 
 You can have a when clause that checks if a specific [View Container](/api/ux-guidelines/views#view-containers) is visible
 
-Context key     | True when
---------------- | ------------
-activeViewlet   | True when view container is visible in the sidebar.<br>Example: `"activeViewlet == 'workbench.view.explorer'"`
-activePanel     | True when view container is visible in the panel.<br>Example: `"activePanel == 'workbench.panel.output'"`
-activeAuxiliary | True when view container is visible in the secondary sidebar.<br>Example: `"activeAuxiliary == 'workbench.view.debug'"`
+Context name      | True when
```
ankushkushwaha, line 291: @muzimuzhi is there any reason we changes `Context key` to `Context name`?

api/references/when-clause-contexts.md
```diff
@@ -257,10 +257,10 @@ Context name | True when
 
 You can have a when clause that checks if a specific [View](/api/ux-guidelines/views) is visible or focused.
 
-Context name | True when
------------- | ------------
+Context name  | True when
```
ankushkushwaha, line 260:
```suggestion
Context name | True when
```

ankushkushwaha (changes requested):

api/references/when-clause-contexts.md
muzimuzhi: Because on the same page, "Context name" is used in the two other similar places, so I think "Context key" was an exception:
https://github.com/microsoft/vscode-docs/blob/d239f68f168aad63d57202702898d623999d6649/api/references/when-clause-contexts.md?plain=1#L142-L143
https://github.com/microsoft/vscode-docs/blob/d239f68f168aad63d57202702898d623999d6649/api/references/when-clause-contexts.md?plain=1#L260-L261
https://github.com/microsoft/vscode-docs/blob/d239f68f168aad63d57202702898d623999d6649/api/references/when-clause-contexts.md?plain=1#L291-L292

api/references/when-clause-contexts.md
muzimuzhi: The added space makes the `|` on table title row aligned with that same characters on the separator row (`------------- | ----------`) and on the row for `` `focusedView` ``.

ntrogh (approved):
@muzimuzhi thanks for being so meticulous in improving our docs!

api/references/when-clause-contexts.md
ankushkushwaha: Okay, thanks for clarification

api/references/when-clause-contexts.md
ankushkushwaha: Okay. thanks

PR #7593, opened 2024-09-14
Remove incorrect statement about pre release tags

harrydowning:
It appears pre release tags are not ignored but rejected. See https://github.com/microsoft/vscode-vsce/issues/1052


alexkinyu:
hello

benibenj (approved):

PR #7639, opened 2024-09-29
docs: remove double entries in theme color references

ptrptrd:
Remove 2 properties, that are written twice in the theme-color.md


ntrogh (approved):
Thanks for contributing, @ptrptrd

PR #7640, opened 2024-09-29
fix URL redirect

partev:


ntrogh (approved):
Thanks for the update, @partev

PR #7646, opened 2024-10-02
fix: use youtube-nocookie for embeds

rzhao271:
Use youtube-nocookie.com/embed instead of youtube.com/embed to prevent YouTube from adding cookies to our website that violate internal cookie policies.

CC @ntrogh


ntrogh:
@rzhao271 thanks for sharing. We'll need to do a full pass, as there are multiple instances across our docs.

anthonykim1 (approved):

PR #7650, opened 2024-10-03
Fix enum docs, fixes #7511

### LINKED ISSUE #7511: contributes.configuration configuration property schema: Dropdown menu poorly specified (enum / enumDescriptions / enumItemLabels)
VS Code allows extensions to create settings with predetermined options based on a list of presets. [The _Contribution Points_ page's _Configuration property schema_ subsection](https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema) documents how to do that using the `enum` / `enumDescriptions` / `markdownEnumDescriptions` / `enumItemLabels` properties. But given how funky the schema is, it would be best to properly specify that part.

Here are a number of specific issues:

1. There is no code example of `enumItemLabels`.
2. `enum` is mentioned but not specified.
3. `markdownEnumDescriptions` is not mentioned in the title.
4. "will be rendered as Markdown" would be more accurate as "will be **parsed** as Markdown".
5. The relationship between these 4 properties is unspecified. What happens if both `markdownEnumDescriptions` and `enumDescriptions` are set, for instance?

Also, the following:

> The `workbench.iconTheme` setting uses both `enumDescriptions` and `enumItemLabels.` In the screenshot below, the hovered option has the item label "None", with enum description "No file icons" and enum value `null`.

…would be clearer as:
> For example, the `workbench.iconTheme` setting shown below uses both `enumDescriptions` and `enumItemLabels`. In this screenshot, the hovered option has the item label "None", with enum description "No file icons" and `enum` value `null`.

I wish this report is the occasion for a general review of that section but understand that the scope is a little wide. Feel free to consider this as solved once the first specific issue is solved. I will file different tickets for any remaining problems if needed.

rzhao271:
Fix #7511

### DIFF
--- a/api/references/contribution-points.md
+++ b/api/references/contribution-points.md
@@ -141,7 +141,7 @@ See the [Commands Extension Guide](https://code.visualstudio.com/api/extension-g
 
 ## contributes.configuration
 
-Contribute configuration keys that will be exposed to the user. The user will be able to set these configuration options as User Settings or as Workspace Settings, either by using the Settings editor or by editing the JSON settings file directly.
+Contribute configuration keys that will be exposed to the user. The user will be able to set these configuration options in the Settings editor or by editing a settings.json file directly.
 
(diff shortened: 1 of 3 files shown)


rzhao271:
Testing in vnext first

ntrogh (approved):

PR #7671, opened 2024-10-10
docs: add note about local editor context

joyceerhl:
cc @isidorn @sharonlo

### DIFF
--- a/api/extension-guides/chat.md
+++ b/api/extension-guides/chat.md
@@ -27,7 +27,11 @@ Participants can also contribute *commands*, which are a shorthand notation for 
 
 ## Extending GitHub Copilot via GitHub Apps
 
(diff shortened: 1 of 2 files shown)


isidorn:
Thanks Joyce. This looks good 👌 

Though we should not have this in the extension guide (the goal of those docs is to be straigh-forward for authors to get something up and running). Just having it documented in one place (copilot-chat.md) is enough. I also think it does not really have to be in a visible place (since we will link from the product direct to it).

@ntrogh could you fix this please?

ntrogh:
@isidorn on the extension guide, what is the behavior for an extension if the user doesn't allow access? Shouldn't extension devs have to be aware of how to handle this situation?
On the Copilot Chat doc, what do you mean with not in a visible place and linking from the product to it? It makes sense to separate out extension-contributed chat participants and their different behavior of asking for permissions.

isidorn:
@ntrogh this will only show up for service side extensions - so it will not show up for folk building on top of VS Code extensions and reading the docs. Thus I think we should not have it in our chat docs.

> On the Copilot Chat doc, what do you mean with not in a visible place and linking from the product to it? It makes sense to separate out extension-contributed chat participants and their different behavior of asking for permissions.

Joyce already has a place in product where she will link to these docs to. So I am fine if these docs are not super visible and folk just discover them direct from the product. Though the current place might work just fine.

PR #7679, opened 2024-10-12
doesn't provide built-in language support in the core editor

Cecil0o0:
Inside "API" outline, "LANGUAGE EXTENSIONS" section, "Overview" article, there is a paragraph on the top.

I was puzzled with sentence "VS Code doesn't provide built-in language support", because we have multiple language extension inside "extensions" at the root of VS Code codebase.

If I'm right, this pr would like to fix it to make this sentence clear~

### DIFF
--- a/api/language-extensions/overview.md
+++ b/api/language-extensions/overview.md
@@ -9,7 +9,7 @@ MetaDescription: Learn how to write a Language Extension (plug-in) to add suppor
 
 # Language Extensions Overview
 
-Visual Studio Code provides smart editing features for different programming languages through Language Extensions. VS Code not only provides built-in language support, such as bat, clojure etc., but also offers a set of APIs that enable rich language features. For example, it has a bundled [HTML](https://github.com/microsoft/vscode/tree/main/extensions/html) extension that allows VS Code to show syntax highlighting for HTML files. Similarly, when you type `console.` and `log` shows up in IntelliSense, it is the [Typescript Language Features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) extension at work.
+Visual Studio Code provides smart editing features for different programming languages through Language Extensions. VS Code without extensions doesn't provide built-in language support but offers a set of APIs that enable rich language features. For example, it has a bundled [HTML](https://github.com/microsoft/vscode/tree/main/extensions/html) extension that allows VS Code to show syntax highlighting for HTML files. Similarly, when you type `console.` and `log` shows up in IntelliSense, it is the [Typescript Language Features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) extension at work.
 
 Language features can be roughly put into two categories:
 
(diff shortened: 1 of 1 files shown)


ntrogh:
@Cecil0o0 Thanks for pointing out this confusion. I've suggested a modification to your PR to make sure it's technically correct.

Cecil0o0:
@ntrogh would you please describe more for me to work it out about the modification to "make sure it’s technically correct"?

I might not understand for now, that is to say, how could I make a modification to verify "technically correct"?

ntrogh:
@Cecil0o0 I've added a suggestion in the review comments to make the text correct.
Your edit mentioned that VS Code provides built-in language support, which is not the case. Everything is handled by extensions. And some of those extensions are bundled with the editor.

Cecil0o0:
@ntrogh I try to modify my edit for technically correct as well as fix of confusion.

In my opinion, I think this "VS Code" points to "workbench" which means `src/vs/code` and without `extensions`, for not introducing more complexity, I consider to reuse "extensions" word instead of "workbench" to represent same means. I don't know if it is a native-like style representation that I try to imitate.

ntrogh:
@Cecil0o0 I already added a suggestion for you to commit. You can just scroll up and commit my suggestion.

Cecil0o0:
@ntrogh 😂 I would like to commit your suggestion but I couldn't see it even if scroll up, I don't know what happen and how to deal with it

image

api/language-extensions/overview.md
```diff
@@ -9,7 +9,7 @@ MetaDescription: Learn how to write a Language Extension (plug-in) to add suppor
...
+Visual Studio Code provides smart editing features for different programming languages through Language Extensions. VS Code without extensions doesn't provide built-in language support but offers a set of APIs that enable rich language features. For example, it has a bundled [HTML](https://github.com/microsoft/vscode/tree/main/extensions/html) extension that allows VS Code to show syntax highlighting for HTML files. Similarly, when you type `console.` and `log` shows up in IntelliSense, it is the [Typescript Language Features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) extension at work.
```
ntrogh, line 12:
```suggestion
Visual Studio Code provides smart editing features for different programming languages through Language Extensions. VS Code doesn't provide built-in language support in the core editor but offers a set of APIs that enable rich language features.

For example, the [HTML](https://github.com/microsoft/vscode/tree/main/extensions/html) extension uses these APIs to show syntax highlighting for HTML files. Similarly, when you type `console.` and `log` shows up in IntelliSense, it is the [Typescript Language Features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) extension at work.

VS Code bundles some of these extensions with the editor to provide you with rich language support from the start.
```

ntrogh: @Cecil0o0 Here's the updated text.

Cecil0o0: Thank you!

ntrogh (approved):
Thanks for contributing, @Cecil0o0 !

PR #7681, opened 2024-10-15
add argument --update-extensions

oleschri:


ntrogh (approved):
@oleschri Thanks for contributing.

PR #7685, opened 2024-10-17
Add chat participant detection in extension guide

ntrogh:


api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
 
 > **Tip:** Follow-ups should be written as questions or directions, not just concise commands.
 
+### Implement participant detection
+
+To make it easier to use chat participants with natural language, you can implement participant detection. Participant detection is a way to automatically route the user's question to a suitable participant, without having to explicitly mention the participant in the prompt. For example, if the user asks "How do I optimizer this query?", the question would be automatically routed to the `@database` participant.
+
+To implement participant detection, you specify the `disambiguation` property in the extension `package.json` file. The `disambiguation` property contains a list of detection categories, each with a description and examples.
```
isidorn, line 276: Maybe add a sentence "Description and examples will be used by VS Code to determine if the user query should be sent to your participant"

isidorn:
This looks good to me -> approved.
Though would like to get @joyceerhl thoughts before we merge in. Esp around if there are any tips/tricks on make the disambiguation work well.

Also fyi @MRayermannMSFT in case you have feedback from your work on azure.

api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
...
+                            "Can you show me a cat playing with a laser pointer?"
+                        ]
+                    }
+                ]
+            }
+        ],
+    ]
+}
+```
+
```
isidorn, line 340: Should we have a sentence on suggestions on how to write good disambiguation content. And to urge the author to test it out and optimise the disambiguation based on how it behaves in VS Code?

api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
...
+                    "Explain to me what is a linked list in a simple way",
+                    "Can you explain to me what is a function in programming?"
+                ]
+            }
+        ]
+    ]
+}
+```
+
+This code snippet shows how to implement participant detection at the command level.
```
isidorn, line 311: Not sure if we need a full code snippet for commands as well (lot of repetition). Let's see what @joyceerhl thinks

api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
 
 > **Tip:** Follow-ups should be written as questions or directions, not just concise commands.
 
+### Implement participant detection
+
+To make it easier to use chat participants with natural language, you can implement participant detection. Participant detection is a way to automatically route the user's question to a suitable participant, without having to explicitly mention the participant in the prompt. For example, if the user asks "How do I optimizer this query?", the question would be automatically routed to the `@database` participant.
```
joyceerhl, line 274:
```suggestion
To make it easier to use chat participants with natural language, you can implement participant detection. Participant detection is a way to automatically route the user's question to a suitable participant, without having to explicitly mention the participant in the prompt. For example, if the user asks "How do I optimize this query?", the question would be automatically routed to the `@database` participant.
```

There is no builtin `@database` participant, so this paragraph is a bit misleading. Can we pick one of our builtin participants as the poster child instead?

api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
...
+    "chatParticipants": [
+        "id": "chat-sample.cat",
+        "fullName": "Cat",
+        "name": "cat",
+        "description": "Meow! What can I teach you?",
+
+        "commands": [
+            {
+                "name": "play",
+                "description": "Do whatever you want, you are a cat after all",
```
joyceerhl, line 324: This gets shown in the placeholder, so I'm not sure it's a good example of how to use the `description` field

api/extension-guides/chat.md
joyceerhl: +1

joyceerhl (changes requested):
This API will be finalized with the next stable release, so we should probably merge this PR to `vnext`

api/extension-guides/chat.md
```diff
@@ -269,6 +269,77 @@ cat.followupProvider = {
...
+
+To implement participant detection, you specify the `disambiguation` property in the extension `package.json` file. The `disambiguation` property contains a list of detection categories, each with a description and examples.
+
+| Property | Description | Examples |
+|----------|-------------|----------|
+| `category` | The detection category. If the participant serves different purposes, you can have a category for each.  | <ul><li>`cat`</li><li>`workspace_questions`</li><li>`web_questions`</li></ul> |
+| `description` | A detailed description of the kinds of questions that are suitable for this participant. | <ul><li>`The user wants to learn a specific computer science topic in an informal way.`</li><li>`The user just wants to relax and see the cat play.`</li></ul> |
+| `examples` | A list of representative example questions. | <ul><li>`Teach me C++ pointers using metaphors`</li><li>`Explain to me what is a linked list in a simple way`</li><li>`Can you show me a cat playing with a laser pointer?`</li></ul> |
```
MRayermannMSFT, line 282: Afaik, only the first example is used. This would be good to point out assuming if still true. @joyceerhl confirm?

api/extension-guides/chat.md
ntrogh: Added

api/extension-guides/chat.md
ntrogh: I added some general guidelines. @isidorn can you verify?

api/extension-guides/chat.md
ntrogh: I removed the commands example altogether.

api/extension-guides/chat.md
```diff
@@ -267,7 +268,59 @@ cat.followupProvider = {
 };
 ```
 
-> **Tip:** Follow-ups should be written as questions or directions, not just concise commands.
+> [!TIP]
+> Follow-ups should be written as questions or directions, not just concise commands.
+
+### Implement participant detection
+
+To make it easier to use chat participants with natural language, you can implement participant detection. Participant detection is a way to automatically route the user's question to a suitable participant, without having to explicitly mention the participant in the prompt. For example, if the user asks "How do I add a login page to my project?", the question would be automatically routed to the `@workspace` participant.
```
isidorn, line 276: "... routed to the `@workspace` participant because it can answer questions about the user's project"

api/extension-guides/chat.md
```diff
@@ -267,7 +268,59 @@ cat.followupProvider = {
...
+Similarly, you can also configure participant detection at the command level by adding a `disambiguation` property for one or more items in the `commands` property.
+
+Apply the following guidelines to improve the accuracy of participant detection for your extension:
+
+- **Be specific**: The description and examples should be as specific as possible to avoid conflicts with other participants. Avoid using generic terminology in the participant and command information.
+- **Use examples**: The examples should be representative of the kinds of questions that are suitable for the participant. Use synonyms and variations to cover a wide range of user queries.
+- **Use natural language**: The description and examples should be written in natural language, as if you were explaining the participant to a user.
+- **Test the detection**: Test the participant detection with a variation of example questions and verify there's no conflict with built-in chat participants.
+
+> [!IMPORTANT]
```
isidorn, line 322: Not sure if IMPORTANT or I think it is enough to make this a NOTE

isidorn (approved):
Overall looks good - left minor comments. Thank you!

api/extension-guides/chat.md
ntrogh: @joyceerhl Can you confirm if this is the case?

joyceerhl: This is currently true, we can call that out but we do plan to leverage more examples in future

joyceerhl (approved):

PR #7686, opened 2024-10-17
Update file paths in API extension guides

olguzzar:
Also added a quick fix in the intro of the Language Model tutorial!


ntrogh (approved):
Thank you!

PR #7691, opened 2024-10-18
doc: correct message in task-provider.md

srl295:
- it's actually the implementation of CustomBuildTaskTerminal not Pseudoterminal that is found. 
- without this change, it made it look like Pseudoterminal was a custom class when it's actually part of the API.

### DIFF
--- a/api/extension-guides/task-provider.md
+++ b/api/extension-guides/task-provider.md
@@ -118,7 +118,7 @@ The `ShellExecution` executes the `rake compile` command in the shell that is sp
 
 ## CustomExecution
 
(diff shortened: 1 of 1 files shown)


ntrogh:
@srl295 Thanks for raising this. I prefer to keep the docs as-is though:

The sample code actually provides the implementation of the `Pseudoterminal` interface by means of a helper class `CustomBuildTaskTerminal`. So, `CustomBuildTaskTerminal` is an implementation detail and might just as well be an anonymous function, as in the code snippet.

This is also mentioned above the code snippet in the docs, "The task provider is also responsible for implementing Pseudoterminal and returning it from the CustomExecution callback."

PR #7704, opened 2024-10-23
location is restricted in a limited area for debug toolbar as `floating`

Cecil0o0:
image

the location of debug toolbar has a limit `top` value of property of CSS, actually. As shown below: 

Perhaps it could be more fit for us to describe more for currently technically implement.

In addition, `editor` may not be appropriate because debug toolbar would appear outside the `editor`, therefore `window` could be right~

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -144,7 +144,7 @@ In addition, the **debug status** appears in the Status Bar showing the active d
 
 ## Debug actions
 
-Once a debug session starts, the **Debug toolbar** will appear on the top of the editor.
+Once a debug session starts, the **Debug toolbar** will appear on the top of the window.
 
 ![Debug Actions](images/debugging/toolbar.png)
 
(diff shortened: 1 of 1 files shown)


docs/editor/debugging.md
```diff
@@ -157,7 +157,7 @@ Once a debug session starts, the **Debug toolbar** will appear on the top of the
 | Restart <br> `kb(workbench.action.debug.restart)`           | Terminate the current program execution and start debugging again using the current run configuration.                                                             |
 | Stop <br> `kb(workbench.action.debug.stop)`                 | Terminate the current program execution.                                                                                                                            |
 
->**Tip**: Use the setting `setting(debug.toolBarLocation)` to control the location of the debug toolbar. It can be the default `floating`, `docked` to the **Run and Debug** view, or `hidden`. A `floating` debug toolbar can be dragged horizontally and also down to the editor area.
+>**Tip**: Use the setting `setting(debug.toolBarLocation)` to control the location of the debug toolbar. It can be the default `floating`, `docked` to the **Run and Debug** view, or `hidden`. A `floating` debug toolbar can be dragged horizontally and also down to the editor area with a limited maximum distance to top edge.
```
ntrogh, line 160:
```suggestion
>**Tip**: Use the setting `setting(debug.toolBarLocation)` to control the location of the debug toolbar. It can be the default `floating`, `docked` to the **Run and Debug** view, or `hidden`. A `floating` debug toolbar can be dragged horizontally and also down to the editor area (up to a certain distance from the top edge).
```

ntrogh (changes requested):
Thanks for the updates. Added a suggestion for one of the changes.

Cecil0o0:
@ntrogh Thanks for the suggestion.

ntrogh (approved):
@Cecil0o0 Thanks for clarifying this in our docs!

PR #7710, opened 2024-10-25
the interactivity of filtering in the VARIABLES section seems be vanished

Cecil0o0:
img_v3_02g1_dccbf35a-3bec-4010-a605-d0bb61b611eg

let's take a look at the gif above, it could be another interactivity which I try to describe at the changed line, how about that?

### DIFF
--- a/docs/editor/debugging.md
+++ b/docs/editor/debugging.md
@@ -209,9 +209,7 @@ Variables and expressions can also be evaluated and watched in the **Run and Deb
 
 ![Debug Watch](images/debugging/watch.png)
 
-Variable names and values can be filtered by typing while the focus is on the **VARIABLES** section.
-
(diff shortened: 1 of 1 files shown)


ntrogh:
@roblourens @connor4312 Did we change the behavior of filtering in the Debug > Variables view?

connor4312:
Doing so requires pressing ctrl/cmd+option+f to use the standard tree filtering

Cecil0o0:
@connor4312 Hi friend, I have tried to press `cmd` + `option` + `f` when `Debug > Variables` view was focused, but I got a navigation to `Search` view at Primary Sidebar.while, there was no response after `ctrl` + `shift` + `f` pressed.

Would I describe clearly and What should I do for additional?

PR #7717, opened 2024-10-28
The scope of the macro token map is incorrect

RainCmd:
https://github.com/microsoft/vscode/blob/main/src/vs/platform/theme/common/tokenClassificationRegistry.ts
Please look at line 562


ntrogh (approved):
@RainCmd Thanks for the fix!

PR #7724, opened 2024-10-30
fix: release notes v_1.95 link issue

Wadefeng1:


Wadefeng1:
In the current v_1.95 release note, click on Account Preference Quick Pick under Add additional accounts when changing account preferences,
the page will report an error:
Sorry!
It seems the page can't be found.
This PR has fixed the issue.
image
image

release-notes/v1_95.md
```diff
@@ -229,7 +229,7 @@ While most things should just continue to work with your existing extensions, so
 
 ### Add additional accounts when changing account preferences
 
-Last month, we introduced the [Account Preference Quick Pick](./v1_94.md#change-an-extensions-account-preference), which is useful for changing the preferred account to use for an extension should you need to change that for any reason.
+Last month, we introduced the [Account Preference Quick Pick](./v1_94#change-an-extensions-account-preference), which is useful for changing the preferred account to use for an extension should you need to change that for any reason.
```
ntrogh, line 232:
```suggestion
Last month, we introduced the [Account Preference Quick Pick](https://code.visualstudio.com/updates/v1_94#change-an-extensions-account-preference), which is useful for changing the preferred account to use for an extension should you need to change that for any reason.
```

ntrogh: Good catch. The URL needs a full URL to accommodate opening from within VS Code.

PR #7739, opened 2024-11-04
fix broken links

imbant:
Some hash links are broken because of the title in the target markdown file has been changed

I think `inline-suggestions` is renamed as `Code-completions-in-the-editor`
But I'm not sure what is the `chat-features`'s target since the content has been changed a lot


ntrogh (approved):
Thanks for fixing these links, @imbant

PR #7742, opened 2024-11-05
Clarify "compute"

jswhisperer:
docs typo compute to computer


ntrogh:
@gregpalaci Thanks for contributing. In this instance we're using "compute" as a term to refer to the broader category of compute  resources. So, closing this PR.

gregpalaci:
No worries, I feel like it's the wrong wording though. Perhaps "Instance" would make more sense.
It's a verb, it feels very awkward in this context, my own compute would be more like your own thoughts or reasoning, rather than a reference to a "compute" I owned like machine or cloud etc.

Food for thought.

ntrogh:
Maybe we can settle on "compute instance"?

ntrogh (approved):
@jswhisperer Thanks for updating and putting effort in making our docs better! Appreciate it.

PR #7782, opened 2024-11-15
append a commonly used issue/reason item to table.

Cecil0o0:
KeyBindings feature is an excellent experience inside VS Code, I have already unconsciously used in my work for a several years, just as usually as drinking water, the one of the most usage of mine is the `CMD+W` shortcut, I use it across my editing work on the code.

But this experience will be constrained in VS Code for the Web, it does not work as in desktop applicaiton since it has a conflict with shortcut of browser itself, while browser has the precedence.

I think it could be nicer to describe it for reader, shall we could append a item?

image


Cecil0o0:
@ntrogh could you please take a look for this, just leave a comment

(I want to improve my skill for writing while hoping to help other readers from my contributions

ntrogh (commented):
@Cecil0o0 there is already an item about not supporting the keybinding for closing the active editor. It mentions Ctrl/Cmd + F4 because there are two keybindings for closing the active editor.

PR #7786, opened 2024-11-17
fix: typo

Cecil0o0:
just fix to singular usage, make it right


ntrogh (approved):
Thanks for the fixes, @Cecil0o0

PR #7800, opened 2024-11-21
complete the jsonc code sample

Cecil0o0:
When I went through the `File Icon Theme` article tonight, I found that there is a code sample about defining an default icon for `latex` language, it seems to be un-completion. Let me complete it by enclosing the curly braces!


ntrogh (approved):

ntrogh:
Thanks @Cecil0o0

Cecil0o0:
@ntrogh, Thank you for you working on documenting Visual Studio Code, which already helps me a lot for understanding how it works. At the same time, I believe that it could be helpful for other readers too!

PR #7806, opened 2024-11-22
Update icons-in-labels.md

Code-Parth:
add copy icon which is not available in the live docs: https://code.visualstudio.com/api/references/icons-in-labels


ntrogh (approved):
Thanks for adding, @Code-Parth

PR #7809, opened 2024-11-23
Update bundling-extension.md

spcurtis13:
Error should be errors. Obviously not a big deal but I believe correct given the sentence structure.


ntrogh (approved):
@spcurtis13 a small but definitely useful fix. Thanks!

PR #7816, opened 2024-11-26
Add Dart to Known Language identifiers

zmuranaka:
`dart` is a known language identifier but it is not currently listed as such in `docs/languages/identifiers.md`


ntrogh (approved):
Thanks for adding this, @zmuranaka

PR #7823, opened 2024-11-27
substitude markdown link `vscode-nls` with `@vscode/l10n`

Cecil0o0:
This project [vscode-nls](https://github.com/Microsoft/vscode-nls) has deprecated for a new package `@vscode/l10n`, it's declaration shows below:

⚠️ This package is no longer receiving new features in favor of the new localization library, [vscode-l10n](https://github.com/microsoft/vscode-l10n). Please use that collection of libraries instead.

Shall we could catch it up to make this article a little bit more fresh?

I check out package.json of `@vscode/l10n`, and inspect the `browser` field and `main` field have been specified. I confirm that it would be appropriate with the description of list item.


ntrogh (approved):
Thanks @Cecil0o0

PR #7839, opened 2024-12-07
complete the code sample.

Cecil0o0:
In the tutorial for language model, a very good simple code snippet is not completed, it should be completed with a implementation of the `code-tutor.annotate` command.


ntrogh:
@Cecil0o0 We're only showing the change that needs to be made. The code for showing the message is part of the generated code.

Cecil0o0:
Oops, I reopen this document and then find it out, much thanks~

PR #7848, opened 2024-12-09
add what libraries are missing for bionic

costi:
Bionic doesn't have a mention that it's missing glibc 2.28.


ntrogh:
@deepak1556 can you take a look at this?

deepak1556 (approved):

chrmarti (approved):

PR #7852, opened 2024-12-11
Release 1.96

ntrogh:


Tyriar (approved):

Cecil0o0:
I have contributed with some requests pull in this `vscode-docs` repository last month in November, but I can’t see them on the v1.96 UPDATES pages, @ntrogh is that some reason for it? ^_^

ntrogh:
@Cecil0o0 Appreciate all the contributions so far! The release notes are focused on product updates (hence why we don't typically add significant docs updates either). Therefore, we usually only list contributions in product repos. There may have been a few occasions where the docs repo was listed, but not something we commonly do.
Again, very appreciative of the time and effort you put in improving our docs! 🙏

PR #7853, opened 2024-12-11
Addressing feedback about pickArgs

cwebster-99:
Addresses feedback from https://github.com/microsoft/vscode-python/issues/23003

### DIFF
--- a/docs/python/debugging.md
+++ b/docs/python/debugging.md
@@ -407,7 +407,7 @@ Specifies arguments to pass to the Python program. Each element of the argument 
 
 If you want to provide different arguments per debug run, you can set `args` to `"${command:pickArgs}"`. This will prompt you to enter arguments each time you start a debug session.
 
-> **Note**: There is a difference in how `"${command:pickArgs}"` and `["${command:pickArgs}"]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its on string.
+> **Note**: There is a difference in how `"${command:pickArgs}"` and `["${command:pickArgs}"]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its own string.
 
 ### `stopOnEntry`
 
--- a/docs/python/debugging.md
+++ b/docs/python/debugging.md
@@ -407,6 +407,8 @@ Specifies arguments to pass to the Python program. Each element of the argument 
 
 If you want to provide different arguments per debug run, you can set `args` to `${command:pickArgs}`. This will prompt you to enter arguments each time you start a debug session.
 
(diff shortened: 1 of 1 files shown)


docs/python/debugging.md
```diff
@@ -407,6 +407,8 @@ Specifies arguments to pass to the Python program. Each element of the argument
 
 If you want to provide different arguments per debug run, you can set `args` to `${command:pickArgs}`. This will prompt you to enter arguments each time you start a debug session.
 
+> **Note**: There is a difference in how `${command:pickArgs}` and `[${command:pickArgs}]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its on string.
```
karthiknadig, line 410:
```suggestion
> **Note**: There is a difference in how `"${command:pickArgs}"` and `["${command:pickArgs}"]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its on string.
```

I think you need the `"` for it to work.

cwebster-99:
Oh good catch - I think we need to update that elsewhere in the documentation as well

docs/python/debugging.md
```diff
@@ -407,6 +407,8 @@ Specifies arguments to pass to the Python program. Each element of the argument
 
 If you want to provide different arguments per debug run, you can set `args` to `${command:pickArgs}`. This will prompt you to enter arguments each time you start a debug session.
```
cwebster-99, line 408:
```suggestion
If you want to provide different arguments per debug run, you can set `args` to `"${command:pickArgs}"`. This will prompt you to enter arguments each time you start a debug session.
```

anthonykim1 (dismissed):

docs/python/debugging.md
```diff
@@ -405,7 +405,9 @@ Specifies arguments to pass to the Python program. Each element of the argument
 "args": ["--quiet", "--norepeat", "--port", "1593"],
 ```
 
-If you want to provide different arguments per debug run, you can set `args` to `${command:pickArgs}`. This will prompt you to enter arguments each time you start a debug session.
+If you want to provide different arguments per debug run, you can set `args` to `"${command:pickArgs}"`. This will prompt you to enter arguments each time you start a debug session.
+
+> **Note**: There is a difference in how `"${command:pickArgs}"` and `["${command:pickArgs}"]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its on string.
```
cwebster-99, line 410:
```suggestion
> **Note**: There is a difference in how `"${command:pickArgs}"` and `["${command:pickArgs}"]` are parsed, with specific notice to the usage of `[]`. As an array, all arguments are passed as a single string, without brackets each argument is passed as its own string.
```

ntrogh (approved):

PR #7855, opened 2024-12-12
Fix toc link for JS/TS paste

mjbvz:
I think should link to the specific JS/TS section here instead of the paste preferences settings


ntrogh (approved):
Thanks for updating, @mjbvz . I hadn't updated the bookmark after that section got added.

PR #7858, opened 2024-12-13
add note of minimum supported pytest version to docs

eleanorjboyd:
to help with [some confusion users](https://github.com/microsoft/vscode-python/issues/22457) had when getting incompatibility while using pytest 6.2.5 I would like to add a note of 7.0.0 as the min supported version for pytest in the extension. Open to suggestions if you think it should be presented / included in a different way.

Thanks!


ntrogh (commented):
For discoverability, maybe this could be moved to the top of the article? 
Either right after the intro sentence, or you could add an H2 Prerequisites section.

BTW, we now support the [GH markdown alert syntax](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) to make it pop more.

cwebster-99:
I like the idea of a pre-reqs section although I am not sure what else we could include there. I do think we should move it either to the beginning of the "Configure Tests" section or to the top of the article.

cwebster-99:
After looking at this, I am realizing we should probably ax some of the top section. It is very wordy and presents users with a wall of text before they learn about the features we offer.

ntrogh:
There's a Testing article now, which you can refer to instead of that introductory section.
Wdyt?

https://code.visualstudio.com/docs/editor/testing

eleanorjboyd:
I am good with anything- whatever yall decide works for me! Happy to make edits when you have a final answer or you can take over the PR, thanks!

cwebster-99:
Let's remove the entire "background on unit testing" section and start with the example walkthroughs. We can leave the links for background and add a link to VS Code's testing page 

>For a general background on unit testing, read [Unit testing](https://wikipedia.org/wiki/Unit_testing) on Wikipedia. For useful unit test examples, you can review https://github.com/gwtw/py-sorting, a repository with tests for different sorting algorithms.

eleanorjboyd:
@cwebster-99 if you want to remove that section I am happy to just merge it now as that is probably simpler than waiting?

eleanorjboyd:
@cwebster-99 I think im going to vote we do removing the other docs that explain testing in a different PR. I think it makes more sense to separate these with 2 commits and since then we can merge this in. Lmk what you think and approve if so. Thanks!

ntrogh (approved):

PR #7859, opened 2024-12-13
Replace "&" with "and" in 1.96 release notes

JohnVeness:
A pretty minor thing, but the "&" jumped out at me, especially as it was inconsistent with the "and" in "drag and drop"!


ntrogh (approved):
An eye for detail! Great and thanks for updating.

PR #7876, opened 2024-12-18
Fix a broken link

MitjaBezensek:
Looks like the linked header was updated but not the link.


ntrogh (approved):
Ow. Thanks for fixing this @MitjaBezensek!

PR #7903, opened 2025-01-01
Tweak logLevel in esbuild.js examples

felipecrs:
I see no good reason to not display esbuild's warnings. This led me to:

https://github.com/evanw/esbuild/issues/4023

Also, when enabling `ts-check` in the provided example, there's a missing check for location being `null`. I fixed that too.


ntrogh:
@felipecrs Thanks for creating this PR.

@aeschli Can you review this PR? Any reason why the esbuild warnings are muted?

aeschli (approved):

PR #7919, opened 2025-01-06
Update documentation to reflect remove of mingw wildcard support.

browntarik:


ntrogh (approved):
Thanks for updating the docs, @browntarik

Issue #7874, opened 2024-12-17
Document extension runtime security


isidorn:
We need to document that VS Code extensoins do not run sandboxed.

In addition to these, we should document what we are doing to invest in extension security, and what signals can users use when deciding what extension to install.

isidorn:
Probably we should take out https://code.visualstudio.com/docs/editor/extension-marketplace#_can-i-trust-extensions-from-the-marketplace and have a new enriched section

isidorn:
Related item that will link to these docs https://github.com/microsoft/vscode/issues/215527

PR #7920, opened 2025-01-07
Add extension runtime security article

ntrogh:
Fixes https://github.com/microsoft/vscode-docs/issues/7874

### DIFF
--- a/docs/editor/extension-runtime-security.md
+++ b/docs/editor/extension-runtime-security.md
@@ -26,7 +26,7 @@ As of VS Code release 1.97, when you first install an extension from a third-par
 
 The list of trusted publishers is maintained in the `setting(extensions.trustedPublishers)` user setting.
 
-Publishers for extensions that you installed previously are considered trusted and are automatically added to the list of trusted publishers. You can also manually edit the list of trusted publishers.
+Publishers for extensions that you installed previously are considered trusted and are automatically added to the list of trusted publishers. You can also edit the list of trusted publishers by modifying the setting.
 
 ## Determine extension reliability
 
--- a/docs/editor/extension-runtime-security.md
+++ b/docs/editor/extension-runtime-security.md
@@ -1,0 +1,69 @@
+---
+Order: 4
+Area: editor
+TOCTitle: Extension Runtime Security
+ContentId: b921a11a-ed69-4716-bc93-589ba8e01e22
+PageTitle: Visual Studio Code Extension Runtime Security
+DateApproved: 12/11/2024
+MetaDescription: Learn about the security measures in place for Visual Studio Code extensions, including permissions, user reliability checks, and Marketplace protections.
+---
+
+# Extension runtime security
+
+[Extensions](/docs/editor/extension-marketplace.md) greatly enhance the functionality of Visual Studio Code. They can also introduce risks, such as malicious code execution and data privacy concerns. The Visual Studio Marketplace has several measures in place to protect you from malicious extensions, and you can take steps to determine if a particular extension is reliable before installing it.
+
+This document outlines the runtime permissions of extensions in VS Code and the measures in place to protect you from malicious extensions. You will learn how you can make an informed decision about the reliability of an extension before installing it.
+
+## About extension runtime permissions
+
+The [extension host](/api/advanced-topics/extension-host.md) is responsible for running extensions in VS Code. The extension host has the same permissions as VS Code itself. This means that any action that VS Code can perform, an extension can also perform through the extension host.
+
+For example, an extension can read and write files on your machine, make network requests, run external processes, and modify workspace settings.
+
+## Determine the extension reliability
+
+Before you install an extension, you can take various steps to determine if it is reliable. The Visual Studio Marketplace provides you with information about the extension to help you make an informed decision.
+
+* **Ratings & Review** - Read what others think about the extension.
+
+* **Q & A** - Review existing questions and the level of the publisher's responsiveness. You can also engage with the extension's publisher if you have concerns.
+
+* **Issues, Repository, and License** - Check if the publisher provided these and if they have the support you expect.
+
+* **Verified Publisher** - Use the blue check mark next to the publisher name and domain as an extra signal of trust. The check mark indicates that the publisher has proven domain ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain and the good standing of the publisher on the Marketplace for at least six months.
+
+    ![Verified publisher](images/extension-marketplace/bluecheck.png)
+
+## Marketplace protections
+
+The Visual Studio Code Marketplace employs several measures to protect you from malicious extensions:
+
+* **Malware scanning** - The Marketplace runs a malware scan on each extension package that's published to ensure its safety. The scan, which uses several anti-virus engines, is run for each new extension and for each extension update. Until the scan is all clear, the extension won't be published in the Marketplace for public usage.
+
+* **Verified publishers** - Publishers can verify (blue check mark) their identity by proving domain ownership. It indicates that the publisher has proven domain ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain and the good standing of the publisher on the Marketplace for at least six months.
+
+* **Name squatting** - The Marketplace prevents extension authors from name-squatting on official publishers, such as Microsoft or RedHat, and popular extension names such as GitHub Copilot.
+
+* **Kill List** - If a malicious extension is reported and verified, or a vulnerability is found in an extension dependency, the extension is removed from the Marketplace added to a kill list. If the extension has been installed, it is automatically uninstalled by VS Code.
+
+* **Extension Signature Verification** - The Visual Studio Marketplace signs all extensions when they are published. VS Code verifies this signature when you install an extension to check the integrity and the source of the extension package.
+
+## Report suspicious extensions
+
+If you do see an extension that looks suspicious, you can report the extension to the Marketplace:
+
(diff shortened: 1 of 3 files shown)


ntrogh:
@seaniyer Here's the first draft of the dedicated article that discusses extension runtime security. I've reused most of the content that was previously in the FAQ section of the Extension Marketplace article. Can you review and provide feedback if there are other Marketplace measures we need to include? Thanks!

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,63 @@
...
+
+* **Malware scanning**: The Marketplace runs a malware scan on each extension package that's published to ensure its safety. The scan, which uses several antivirus engines, is run for each new extension and for each extension update. Until the scan is all clear, the extension won't be published in the Marketplace for public usage.
+
+* **Verified publishers**: Publishers can verify (blue check mark) their identity by proving domain ownership. It shows that the publisher has proven domain-name ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain and the good standing of the publisher on the Marketplace for at least six months.
+
+* **Name squatting**: The Marketplace stops extension authors from stealing the names of official publishers, such as Microsoft or RedHat, and popular extensions, like GitHub Copilot.
+
+* **Kill List**: If a malicious extension is reported and verified, or a vulnerability is found in an extension dependency, the extension is removed from the Marketplace and added to a *kill list*. If the extension has been installed, it's automatically uninstalled by VS Code.
```
isidorn, line 47: Is using the word "kill" fine in docs?
I think yes, but wanted to double check?

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,63 @@
...
+
+* **Extension Signature Verification**: The Visual Studio Marketplace signs all extensions when they're published. VS Code checks this signature when you install an extension to verify the integrity and the source of the extension package.
+
+## Report suspicious extensions
+
+If you do see an extension that looks suspicious, report the extension to the Marketplace team:
+
+1. Open the extension's page in the [Visual Studio Marketplace](https://marketplace.visualstudio.com/vscode).
+
+1. Select the **Report a concern** link at the bottom of the extension **More Info** section.
```
isidorn, line 57: We should add our SLA for response @seaniyer

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Related resources
```
isidorn, line 59: There need s to be a link somewhere to  https://code.visualstudio.com/docs/setup/enterprise#_configure-allowed-extensions

The approach would be "Do you want to configure what extensions are allowed in your orgranisation, check out ...."

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,63 @@
...
+* **Verified Publisher**: Use the blue check mark next to the publisher's name and domain name as an extra signal of trust. The check mark indicates that the publisher has proven domain-name ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain name and the good standing of the publisher on the Marketplace for at least six months.
+
+    ![Verified publisher](images/extension-marketplace/bluecheck.png)
+
+## Marketplace protections
+
+The Visual Studio Code Marketplace employs several mechanisms to protect you from malicious extensions:
+
+* **Malware scanning**: The Marketplace runs a malware scan on each extension package that's published to ensure its safety. The scan, which uses several antivirus engines, is run for each new extension and for each extension update. Until the scan is all clear, the extension won't be published in the Marketplace for public usage.
+
```
isidorn, line 42: We need a section about detonation. Not disclosing everything that is being done, but just that we have a dynamic detection mechanism in a clean room VM.

@seaniyer

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,63 @@
...
+* **Verified publishers**: Publishers can verify (blue check mark) their identity by proving domain ownership. It shows that the publisher has proven domain-name ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain and the good standing of the publisher on the Marketplace for at least six months.
+
+* **Name squatting**: The Marketplace stops extension authors from stealing the names of official publishers, such as Microsoft or RedHat, and popular extensions, like GitHub Copilot.
+
```
isidorn, line 46: Detection of inflated downloaded counts

@seaniyer

isidorn (changes requested):
This is a great first stab at this document.

I will share it with folk on the MP side and @sandy081 could also give it a read

fyi @joaomoreno

isidorn:
The goal of docs is to:
1) Give transparency about extensions running un-sandboxes
2) Show signals how users can decide if an extension is malicious
3) Provide some transparency about what we are doing to fight malicious extensions, and the investments we are making

docs/editor/extension-runtime-security.md
mariaghiondea: We should use another word. maybe "block list"?

docs/editor/extension-runtime-security.md
mariaghiondea: I'm hesitant to let bad actors know that we have this mechanism. Maybe if we keep it generic enough... otherwise we risk giving too many details and help them bypass.

docs/editor/extension-runtime-security.md
mariaghiondea: Initial response is 1 business day. That might only be an acknowledgement. We depend on a lot of factors for the actual action - 3 business days.

ntrogh: I added one day in the doc for initial response. Given that we depend on many factors, any other timeline might be incorrect?

docs/editor/extension-runtime-security.md
ntrogh: Added as a tip at the end of the Reliability section and also included a link in related resources at the end of the doc.

docs/editor/extension-runtime-security.md
ntrogh: I added this - can you review if this is correct and sufficiently detailed?

docs/editor/extension-runtime-security.md
ntrogh: @mariaghiondea  I added a generic item, can you review and let me know if correct and sufficient?

mariaghiondea: LG to me. Thank you

mariaghiondea (approved):

sandy081 (approved):

sandy081:
Thanks. LGTM

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,82 @@
...
+For example, an extension can read and write files on your machine, make network requests, run external processes, and modify workspace settings.
+
+## Extension publisher trust
+
+As of VS Code release 1.97, when you first install an extension from a third-party publisher, VS Code shows a dialog prompting you to confirm that you trust the publisher of that extension.
+
+The list of trusted publishers is maintained in the `setting(extensions.trustedPublishers)` user setting.
+
+Publishers for extensions that you installed previously are considered trusted and are automatically added to the list of trusted publishers. You can also manually edit the list of trusted publishers.
```
isidorn, line 29: ... edit the list of trusted publishers by modifying the setting.

docs/editor/extension-runtime-security.md
```diff
@@ -0,0 +1,82 @@
...
+
+Before you install an extension, you can take various steps to determine if it's reliable. The Visual Studio Marketplace provides you with information about the extension to help you make an informed decision:
+
+* **Ratings & Reviews**: Read what others think about the extension.
+
+* **Q & A**: Review existing questions and the level of the publisher's responsiveness. You can also engage with the extension's publisher if you have concerns.
+
+* **Issues, Repository, and License**: Check if the publisher provided these and if they have the support you expect.
+
+* **Verified Publisher**: Use the blue check mark next to the publisher's name and domain name as an extra signal of trust. The check mark indicates that the publisher has proven domain-name ownership to the Marketplace. It also shows that the Marketplace has verified both the existence of the domain name and the good standing of the publisher on the Marketplace for at least six months.
```
isidorn, line 41: Let's add a bullet point about "long-standing presence"
The goal is that publishers that have been publishing extensions for a long time should be more trust-worthy

isidorn (approved):
Looks good. Proposed two minor changes

isidorn:
Also @joaomoreno sketched this image

Not sure if we should include it, or some form of it - to make this document easier to parse.
What do you think?

image

ntrogh:
@isidorn The comparison image is definitely nice. However, because we're dealing with a sensitive topic, it might result in people oversimplifying the decision process for trusting an extension. Unless it's 100% exact and approved, I'd refrain from putting such guidance in the docs.

docs/editor/extension-runtime-security.md
ntrogh: We don't have clear information about publisher's long-standing presence, apart from going through the history of a specific extension's releases. Am I missing a specific signal?

isidorn: Good point
Marketplace should update this page https://marketplace.visualstudio.com/publishers/vadimcn to contain more publisher information. I will discuss with @seaniyer @mariaghiondea 
The dialog points to that page.

isidorn: Though I still think we can add that point, and users can investigate when was the extension first published.
@ntrogh I leave that decision up to you nick

isidorn (approved):

mariaghiondea (approved):

PR #7933, opened 2025-01-11
Update terminal inline chat to mention correct directional key

burkeholland:
This PR updates the ccomand to select the chat response from `Ctrl` + `down`, to `Ctrl` + `up`, which appears to be the correct command


karthiknadig (approved):

ntrogh (approved):
Thanks @burkeholland

PR #7937, opened 2025-01-13
Fix git credential manager for WSL

mwiemer-microsoft:
As of my git `2.47.1.windows.1` from November 2024, the new name does not end in `core`. Took some research, please confirm!


ntrogh (approved):
Thanks for fixing this, @mwiemer-microsoft

Issue #7951, opened 2025-01-19
The "Variables Reference" article is not accessible from the left menu


adriancuadrado:
https://code.visualstudio.com/docs/editor/variables-reference

![Image](https://github.com/user-attachments/assets/62b654d0-7af8-48c6-834c-253b6b500fb7)

It should be somewhere under "USER GUIDE", but it's not.

adriancuadrado:
This is the article in the source code: https://github.com/microsoft/vscode-docs/blob/main/docs/editor/variables-reference.md

ntrogh:
@adriancuadrado Holding off on your PR but keeping this issue on the backlog, as we're restructuring the table of contents.

PR #7952, opened 2025-01-19
Added an order to variables-reference.md so it shows in the left pane

adriancuadrado:
Added an order so it shows in the left pane

Fixes https://github.com/microsoft/vscode-docs/issues/7951

### DIFF
--- a/docs/editor/variables-reference.md
+++ b/docs/editor/variables-reference.md
@@ -1,5 +1,5 @@
 ---
-Order:
+Order: 27
 Area: editor
 TOCTitle: Variables reference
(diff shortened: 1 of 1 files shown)


docs/editor/variables-reference.md
```diff
@@ -1,5 +1,5 @@
 ---
-Order:
+Order: 27
```
adriancuadrado, line 2: I added 27 because that's the smallest number that I think would make this article show at the end of the list:

image

ntrogh:
@adriancuadrado Appreciate you taking the time to contribute to our docs! Closing this PR as we are revising the table of contents structure and this will render your PR invalid.

PR #8009, opened 2025-02-08
Document new predefined variable `${columnNumber}`

muzimuzhi:
New predefined variable `${columnNumber}` is added in v1.97, but not documented. This PR adds its doc and example.

https://code.visualstudio.com/updates/v1_97#_column-number-variable

Also, the missing example of `${fileDirnameBasename}` is added, and the order of examples of `${fileDirname}` and `${fileExtname}` is swapped to align with that in section ["Predefined variables examples"](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables).


ntrogh (approved):
@muzimuzhi thanks for the addition!

PR #8045, opened 2025-02-18
Updating Python settings references

cwebster-99:
Reordered settings to match the Settings UI. 

Adding:
- REPL Smart Send 
- debug just my code 

Removed:
- logging.level (this doesn't seem to exist under `python.`)

Once stable, we should also add shell integration and Python environments settings.


cwebster-99:
@luabud I can reorder the Pylance settings as well to reflect how they appear in the settings UI if desired. I am also curious if there are "key" settings you think should be added to the Pylance section?

cwebster-99:
@luabud I reordered the Pylance settings to reflect the order in the settings ui as well, let me know if the team would prefer something different! Are there other "key" settings you think should be added (or even settings that could be removed)?

karthiknadig (approved):

docs/python/settings-reference.md
```diff
@@ -20,19 +20,53 @@ For general information about working with settings in VS Code, refer to [User a
...
+## Testing settings
+
+### General testing
+
+| Setting<br/>(python.testing.) | Default | Description | See also |
+| --- | --- | --- | --- |
+| autoTestDiscoverOnSaveEnabled | `true` | Specifies whether to enable or disable auto run test discovery when saving a test file. | [Testing](/docs/python/testing.md) |
+| cwd | null | Specifies an optional working directory for tests. | [Testing](/docs/python/testing.md) |
+| debugPort | `3000` | Port number used for debugging of unittest tests. | [Testing](/docs/python/testing.md) |
+| promptToConfigure | `true` | Specifies whether VS Code prompts to configure a test framework if potential tests are discovered. | [Testing](/docs/python/testing.md)  |
```
eleanorjboyd, line 54: we now have a new setting I forgot to mention in the meeting: https://github.com/microsoft/vscode-python/commit/08e228df23b3efbfa404a472a249d482bcab9ee4

eleanorjboyd: this one impacts autoTestDiscoverOnSaveEnabled as well as it might change the file type which refresh happens on

eleanorjboyd (dismissed):
two comments on a new setting which can be included

docs/python/settings-reference.md
cwebster-99: Since this isn't a comprehensive setting list, I think I am going to leave this new setting out since we include `autoTestDiscoverOnSaveEnabled` which I assume to be more widely adopted. We should make sure we document this new setting in the testing docs when we update it later this iteration though.

luabud (approved):

PR #8090, opened 2025-02-26

codecat1111:
This update improves the documentation for creating a VS Code extension by ensuring developers properly compile their extension before debugging.

**Changes Made**
Added a build step `"build": "tsc"` in the scripts section of `package.json`.
Updated the documentation to instruct users to run npm run build before debugging to prevent missing module errors.

**Why This Change?**
Without compiling the extension first, users may encounter errors like:
`_Cannot find module 'out/extension.js'_`
This ensures a smoother debugging experience by making it clear that TypeScript files need to be compiled before running the extension.

**Testing**
Verified that `npm run build` successfully compiles TypeScript files into the `out/` directory.
Confirmed that the Hello World command appears in the debug window after building.

### DIFF
--- a/api/get-started/your-first-extension.md
+++ b/api/get-started/your-first-extension.md
@@ -54,7 +54,7 @@ Run the **Hello World** command from the Command Palette (`kb(workbench.action.s
 
 You should see the `Hello World from HelloWorld!` notification showing up. Success!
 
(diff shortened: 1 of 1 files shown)


ntrogh:
@codecat1111 Thanks for contributing to our docs. 

Closing the PR as this functionality is already covered. The Yeoman generated code has a build task defined in the `tasks.json` file. In the `launch.json`, which is invoked when you run/debug the project, there's a `preLaunchTask` that invokes this build task. So there's no need for an explicit build script in `package.json`.

PR #8101, opened 2025-02-27
Small grammar correction

ivar117:
"and select and **C++**" => "and select **C++**"


ntrogh (approved):
@ivar117 Thanks for correcting this!

PR #8122, opened 2025-03-07
Update Podman documentation

l0rd:
With the release of Dev Containers v0.399, there is no need to update a devcontainers.json to make it work with Podman (c.f. https://github.com/microsoft/vscode-remote-release/issues/10706).

And commands like “Dev Containers: Try a Dev Container Sample” and “Dev Containers: Clone Repository in Container Volume…” that used to fail with Podman are now fixed (c.f. https://github.com/microsoft/vscode-remote-release/issues/10706)

Also, some Podman details were outdated:
- Podman isn't Linux-only but currently supports Windows and macOS too
- Podman allows `Docker-from-Docker` now (in both rootless and rootful mode)
- Podman supports both Docker-Compose and Podman-Compose (with a priority for Docker-Compose)

### DIFF
--- a/remote/advancedcontainers/docker-options.md
+++ b/remote/advancedcontainers/docker-options.md
@@ -37,6 +37,14 @@ Colima automatically sets up a `colima` [Docker context](https://docs.docker.com
 
 > Note: Colima uses Alpine Linux, which isn't supported by Remote - SSH.
 
+## Podman
+
+[Podman](https://podman.io/) 5+ is mostly compatible with Docker's CLI commands and therefore does work if you update the **Docker Path** setting (via **Dev > Containers: Docker Path** in the Settings editor) to `podman` on Linux, Windows or macOS.
+
+![Docker Path setting](images/platform-options/docker-path-setting.png)
+
+Podman has a [`podman compose` command](https://docs.podman.io/en/latest/markdown/podman-compose.1.html) too but that requires a compose provider that can be either Docker Compose or [Podman Compose](https://github.com/containers/podman-compose).
+
 ## Linux
 
 If you're using Linux on your local machine, or already have a remote Linux machine with SSH access, you can reference the [Docker documentation](https://docs.docker.com/engine/install/) for installing Docker on Linux, with [specific information per distribution](https://docs.docker.com/engine/install/centos/).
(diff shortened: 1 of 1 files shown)


l0rd:
@microsoft-github-policy-service agree company="Red Hat"

chrmarti (approved):
Great, thanks!

chrmarti:
@l0rd Would the `--keep-id` flag still be needed when running Podman in rootless mode and using a regular user in the container? We previously used that to make sure the regular container user can access the files from the bind mount.

chrmarti:
I still need that flag plus `z` on the mount options. (Fedora 40 with Podman 5.4.0)
```
{
	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
	"workspaceMount": "type=bind,source=${localWorkspaceFolder},target=${containerWorkspaceFolder},z",
	"runArgs": ["--userns=keep-id"]
}
```

`z` seems to be needed for root and regular users in the container. `keep-id` seems to be needed because otherwise the workspace folder is owned by root in the container and the regular user cannot access it.

The configuration for this dev container is such that the container is started as root, but VS Code then uses a regular user to work with it. Most of our prebuilt images have this setup to allow for entrypoints running as root but then dev tools run as a regular user.

l0rd:
@chrmarti I have done some investigations and on Windows WSL and macOS things work out of the box: no need for `z` and `--userns` options.

On Linux things are a little bit more complicated but we may solve the problem if the Dev Containers extension:
- **sets the environment variable `PODMAN_USERNS=keep-id`** before running `podman/docker run` commands
- **adds `--security-opt label=disable`** for every `podman/docker run` command

With these two changes, setting the `z` and `--userns` options wouldn't be needed anymore. It works for both rootless and rootful Podman.

---

Here is a detailed explanation:

- For rootful podman, the option to relabel the file objects (`z` or `Z`) may (*) be required (`--security-opt label=disable` works too)
- For rootless podman, in addition to the relabelling, the option to keep the same user id in the namespace (`--userns=keep-id`) is required.
- Docker (rootful) has `--security-opt label=disable` and as a consequence doesn't need any specific flag/option.

(*) it may work without that option depending on the existing labels and permissions of the folder that we want to bind mount, but to always work, `z` is required.

To set the relabel option there are multiple options:
- use the `:z` or `:Z` volume option
- **add the `--security-opt label=disable` to `podman run` command**
- set `label = false` in `[containers]` section of `containers.conf`

To set userns to keep-id there are multiple options too:
- add the `--userns=keep-id` to `podman run` commands (but that doesn't seem like a valid value for Docker)
- **`export PODMAN_USERNS=keep-id`**
- set `userns = "keep-id"` in `[containers]` section of `containers.conf`

chrmarti:
@l0rd Thanks! Continuing in https://github.com/microsoft/vscode-remote-release/issues/10798.

PR #8123, opened 2025-03-09
Fix Refined Snippet API in 1.98 release notes

JohnVeness:
Added a full stop at the end, and changed to what I assume the sentence was supposed to say.


ntrogh (approved):
@JohnVeness Thanks for the update!

PR #8154, opened 2025-03-14
HTML: discuss character case

yecril25:
It feels unexpected that syntax highlighting in HTML code renders lowercase and uppercase tags differently.  I added a section to explain this feature.

### DIFF
--- a/docs/languages/html.md
+++ b/docs/languages/html.md
@@ -54,6 +54,14 @@ When modifying a tag, the linked editing feature automatically updates the match
 "editor.linkedEditing": true
 ```
 
+## Character case
+
+Intellisense suggestions and syntax highlighting
+prefer lowercase tags and attribute names.
+Uppercase tags will be treated as support classes
+and not as entity tags by the syntax highlighting engine.
+Formatting will not change markup character case.
+
 ## Color picker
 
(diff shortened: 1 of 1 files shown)


ntrogh:
Can you provide an example of a difference in rendering between lowercase and uppercase? By default, there shouldn't be any differences, as HTML is case insensitive.

yecril25:
Lowercase tags are blue, uppercase tags are green.

ntrogh:
> Lowercase tags are blue, uppercase tags are green.

This is what  get with default setup. 

image

yecril25:
I get blue and green on a Chromebook.  I examined the reason and lowercase is recognised as an entity while uppercase is recognised as support.

yecril25:
Syntax colouring in HTML flavour is a built-in extension, so if I manage to switch it off somehow, there will be no colouring and the issue will be moot.

ntrogh:
Can you share a screenshot of how the HTML is rendered in your environment? Unfortunately, I'm unable to reproduce this behavior in any of our supported environments.

I suggest you log an issue in the [vscode repo](https://github.com/microsoft/vscode/issues) if this is an expected behavior.

Issue #8170, opened 2025-03-24
Document Copilot BYOK (bring your own key)


isidorn:
We should document our Bring Your Own Key offering.

@ntrogh any thoughts where these docs should live?

ntrogh:
@isidorn Thought about adding an "AI Models" article in the Copilot docs. That would cover both changing the model for chat, completions, and BYOK.

isidorn:
This makes sense to me. Not sure if "AI Models" is the right name. I see you already use it, so it might work.
Alternative is "Language Models" or "AI Language Models"

ntrogh:
Will go with "AI Language Models"

PR #8172, opened 2025-03-24
Change model docs (includes BYOK)

ntrogh:
Fixes https://github.com/microsoft/vscode-docs/issues/8170

### DIFF
--- a/build/sitemap.xml
+++ b/build/sitemap.xml
@@ -376,6 +376,11 @@
         <priority>0.8</priority>
     </url>
     <url>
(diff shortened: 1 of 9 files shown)


lramos15:
Sorry for being slow here, a lot of stuff in BYOK is changing.


- Groq is not public
- Manage models / BYOK is only available to Copilot Free and Copilot Pro subscribers it's not available to Copilot Business or Enterprise.
- We should make it clear that this is a preview feature

ntrogh:
Updated

PR #8229, opened 2025-04-04
fix broken image link

lpas:
The image below Command Palette https://code.visualstudio.com/api/ux-guidelines/overview#command-palette doesn't link to the clicked image. It opens a dead page, this pr fixes this


ntrogh (approved):
@lpas Thanks for the #FridayFix!

PR #8230, opened 2025-04-04
Fix wording in  chat-agent-mode.md

fbehrens:


ntrogh (approved):
Thanks for this #FridayFix 🙏 @fbehrens

PR #8242, opened 2025-04-07
Fix doc mentions tests in the "Generate documentation" section

philippgille:
It was probably a copy paste error, as the sentence with "test" is just below in the section about tests


ntrogh (approved):
@philippgille Doh! Thanks for spotting and fixing this.

PR #8244, opened 2025-04-07
more closer to target for the links

Cecil0o0:
Invalid fragments in some links in `scm-provider` documentation that results far from target position of `contribution-points` documentation


ntrogh (approved):
@Cecil0o0 Thanks for the update.

PR #8277, opened 2025-04-14
Add known issue for csdk test explorer to faq

ocallesp:


ocallesp:
@webreidi

docs/csharp/cs-dev-kit-faq.md
```diff
@@ -215,6 +215,12 @@ If you are encountering issues with Test Explorer, you can enable diagnostic log
 
 This information will help in diagnosing and resolving the issue more efficiently.
 
+### Known Issues
```
webreidi, line 218:
```suggestion
### I'm using xUnit v3 with the Microsoft.Testing.Platform, why aren't my tests being discovered?
```

docs/csharp/cs-dev-kit-faq.md
```diff
@@ -215,6 +215,12 @@ If you are encountering issues with Test Explorer, you can enable diagnostic log
...
+
+- Using xUnit v3 with package Microsoft.Testing.Platform on macOS can cause test discovery and debugging issues in the Test Explorer for C# Devkit and Visual Studio.
```
AbhitejJohn, line 220: Is this specific to mac?

docs/csharp/cs-dev-kit-faq.md
```diff
@@ -215,6 +215,12 @@ If you are encountering issues with Test Explorer, you can enable diagnostic log
...
+Workaround: Downgraded to xUnit v2 or use VSTest instead.
```
webreidi, line 221:
```suggestion
Workaround: We are working on fixing this issue. To get this working now, you an either: 
- Downgraded to xUnit v2 and VSTest, or 
- stay on xUnit v3 and change to use VSTest instead of Microsoft.Testing.Platform.
```

AbhitejJohn, line 221: Can we link to an existing issue to help users follow along and/or share additional feedback?

docs/csharp/cs-dev-kit-faq.md
AbhitejJohn, line 220: We probably do not need to mention VS here since this doc is tailored to C# Dev Kit specifically.

AbhitejJohn (approved):

docs/csharp/cs-dev-kit-faq.md
ocallesp: Thanks for the suggestion Wendy.
I agree that framing it as a question makes sense for FAQ.

Maybe we could just adjust the heading to something like 'Are there known issues?'

webreidi: No, Everything on the FAQ page is a "known issue". The idea is to have them listed so they can easily find their issue and self-correct it.

peterwald:
I think adding this to known issues may be a bit premature. I am investigating the xUnit v3 issues. There are enough unknowns between platform and framework differences that I think we need to investigate further before posting this.

docs/csharp/cs-dev-kit-faq.md
ocallesp: I verified that this is also failing on VSCode C#DevKit on Windows.

As a side note. I verified that on VS it is also failing to attach the debugger.

AbhitejJohn: I mean if it is failing on all OS's I'd rather that we fix this than document it here.

docs/csharp/cs-dev-kit-faq.md
```diff
@@ -215,6 +215,14 @@ If you are encountering issues with Test Explorer, you can enable diagnostic log
 
 This information will help in diagnosing and resolving the issue more efficiently.
 
+### I'm using xUnit v3 with the Microsoft.Testing.Platform, why aren't my tests being discovered?
+
+- Using xUnit v3 with package Microsoft.Testing.Platform on macOS can cause test discovery and debugging issues in the Test Explorer for C# Devkit and Visual Studio.
```
Youssef1313, line 220: > and Visual Studio

How so? Discovery works on VS.

docs/csharp/cs-dev-kit-faq.md
Youssef1313: > I verified that this is also failing on VSCode C#DevKit on Windows.

It works for me on Windows. I think we need to have better understanding of what the issue is, when it's happening, so we are able to document it correctly, instead of guiding users in a wrong way.

Youssef1313 (changes requested):
I agree with @peterwald about not documenting this without at least investigating and having more information on what the problem is.

ocallesp:
I will abandon this PR since we are currently working on a fix

PR #8302, opened 2025-04-23
Document rescaleOverlappingGlyphs

Tyriar:
Forgot to document this feature from 1.90 https://code.visualstudio.com/updates/v1_90#_rescaling-overlapping-glyph-in-the-terminal

See confusion in recent issue https://github.com/microsoft/vscode/issues/247211

### DIFF
--- a/docs/terminal/appearance.md
+++ b/docs/terminal/appearance.md
@@ -180,7 +180,7 @@ This feature can be disabled by setting `"terminal.integrated.customGlyphs": fal
 
 ### Rescaling ambiguous width glyphs
 
-Some unicode characters have an ambiguous width where the backend and frontend of the terminal may not agree on the size. When [GPU acceleration](#gpu-acceleration) is enabled and this occurs, the glyph will be rescaled horizontally in order to fit in the a single cell in order to prevent overlapping.
+Some unicode characters have an ambiguous width where the backend and frontend of the terminal may not agree on the size. When [GPU acceleration](#gpu-acceleration) is enabled and this ambiguity occurs, the glyph will be rescaled horizontally to fit in a single cell and prevent overlapping.
 
 For example, the below image features roman numeral characters that are squashed into a single cell:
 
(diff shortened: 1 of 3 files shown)


docs/terminal/appearance.md
```diff
@@ -178,6 +178,18 @@ Below are some examples of these characters with custom line height and letter s
 
 This feature can be disabled by setting `"terminal.integrated.customGlyphs": false`.
 
+### Rescaling ambiguous width glyphs
+
+Some unicode characters have an ambiguous width where the backend and frontend of the terminal may not agree on the size. When [GPU acceleration](#gpu-acceleration) is enabled and this occurs, the glyph will be rescaled horizontally in order to fit in the a single cell in order to prevent overlapping.
```
ntrogh, line 183:
```suggestion
Some unicode characters have an ambiguous width where the backend and frontend of the terminal may not agree on the size. When [GPU acceleration](#gpu-acceleration) is enabled and this ambiguity occurs, the glyph will be rescaled horizontally to fit in a single cell and prevent overlapping.
```

docs/terminal/appearance.md
```diff
@@ -178,6 +178,18 @@ Below are some examples of these characters with custom line height and letter s
...
+
+For example, the below image features roman numeral characters that are squashed into a single cell:
+
+![VIII and XII characters would be rescaled horizontally so as to not overlap with following characters. They feature a thinner stroke width when this happens due to the scaling](images/appearance/rescale-on.png)
+
+This feature can be disabled by setting `"terminal.integrated.rescaleOverlappingGlyphs": false` which would result in the following overlapped rendering:
```
ntrogh, line 189:
```suggestion
This feature can be disabled by setting `setting(terminal.integrated.rescaleOverlappingGlyphs)` to `false`, which would result in the following overlapped rendering:
```

ntrogh (dismissed):
@Tyriar thanks for updating the docs. Added some optional suggestions.

docs/terminal/appearance.md
ntrogh: Updated

chrmarti (approved):

PR #8310, opened 2025-04-24
Update C++ Settings doc

AlexandraKemperMS:
The current settings doc for C++ points to the C_Cpp.default settings and contains inaccurate and incomplete information. Default settings are described in the general VS Code docs and only need to be linked to, and one exception needs to be covered. Additionally, this settings doc contains no information on the c_cpp_properties .json schema, which is in a separate doc that does not even have a place in the TOC right now. These two docs have been merged under "Settings Reference" to match Python's layout and updated with the most recent information.


docs/cpp/customize-cpp-settings-cpp.md
```diff
@@ -133,18 +148,18 @@ Note, this is an example of all fields. You do not need to specify all fields in
...
 - `compilerArgs` (optional)
-  Compiler arguments to modify the includes or defines used, for example `-nostdinc++`, `-m32`, etc. Arguments that take additional space-delimited arguments should be entered as separate arguments in the array, for example, for `--sysroot <arg>` use `\"--sysroot\", \"<arg>\"`.
+  Compiler arguments to modify the includes paths or defines used, for example `-nostdinc++`, `-m32`, etc. Arguments that take additional space-delimited arguments should be entered as separate arguments in the array, for example, for `--sysroot <arg>` use `\"--sysroot\", \"<arg>\"`.
 
 - `intelliSenseMode`
-  The IntelliSense mode to use that maps to an architecture-specific variant of MSVC, gcc, or Clang. If not set or if set to `${default}`, the extension will choose the default for that platform.
+  The IntelliSense mode to use that maps to an architecture-specific variant of MSVC, gcc, or Clang. If not set or if set to `${default}`, the extension chooses the default for that platform.
 
   Platform defaults:
   - Windows: `windows-msvc-x64`
```
fearthecowboy, line 164: Is it listed somewhere the actual values for the triplets?

Might be nice to have at least the viable parts listed `windows|macos|linux`-`msvc|clang|gcc`-`x86|x64|arm|arm64`

docs/cpp/customize-cpp-settings-cpp.md
```diff
@@ -154,19 +169,19 @@ Note, this is an example of all fields. You do not need to specify all fields in
...
 
 - `defines`
   A list of preprocessor definitions for the IntelliSense engine to use while parsing files. Optionally, use `=` to set a value, for example `VERSION=1`.
 
 - `cStandard`
-  The version of the C language standard to use for IntelliSense. For example, `c17`, `gnu23`, or `${default}`. Note that GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense will emulate the equivalent C standard version.
+  The version of the C language standard to use for IntelliSense. For example, `c17`, `gnu23`, or `${default}`. Note: GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense emulates the equivalent C standard version.
```
fearthecowboy, line 177: And again, are the supported values listed?

docs/cpp/customize-cpp-settings-cpp.md
```diff
@@ -1,21 +1,20 @@
...
+The C++ Extension settings are highly configurable. This article explains the schema for the `c_cpp_properties.json` file. For general information about settings in VS Code, refer to [User and workspace settings](/docs/configure/settings.md), as well as the [Variables reference](/docs/reference/variables-reference.md) and [Default VS Code Settings](/docs/reference/default-settings.md).
 
+Looking to get started with configuring your C++ project? Begin with [configure Intellisense](/docs/cpp/configure-intellisense.md).
 ## Example of Variables
 
-Note, this is an example of all fields. You do not need to specify all fields in your `c_cpp_properties.json` file. The extension will automatically fill in any missing fields with default values.
+Note, this json is an example of all fields supported in `c_cpp_properties.json`. You don't need to specify all fields in your `c_cpp_properties.json` file. The extension automatically fills in any missing fields with default values.
```
bobbrow, line 17: Maybe we don't need 3 configs in the example? Can it be made more generic? I'm fine with using linux-style paths (forward slashes) and /usr/bin/g++ as the compilerPath in the remaining config.

docs/cpp/customize-cpp-settings-cpp.md
```diff
@@ -133,18 +147,18 @@ Note, this is an example of all fields. You do not need to specify all fields in
 ## Configuration properties
 
 - `name`
-  A friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that will be autoselected on those platforms. The status bar in VS Code will show you which configuration is active. You can also select the label in the status bar to change the active configuration.
+  A friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that are autoselected on those platforms. The status bar in VS Code shows you which configuration is active. You can also select the label in the status bar to change the active configuration.
 
 - `compilerPath` (optional)
```
bobbrow, line 152: Can you remove the "(optional)" string?  While it is optional, so are most other properties. The difference being that this is one we *really* want people to set.

docs/cpp/customize-cpp-settings-cpp.md
bobbrow: The values will appear in the IntelliSense for the config file. I would say that we can forego listing all the combinations here.

docs/cpp/customize-cpp-settings-cpp.md
AlexandraKemperMS: Originally this was supposed to showcase every single variable we supported, but I am fine with shortening it since most of the variables are the same between different configurations. Are you asking for it to just be one platform's configuration (Ex. only Mac) as an example?

bobbrow: If we could shrink to one config, yes, I think macOS would be best since it has the extra property. I was just trying to avoid a JSON overload for the reader.  If you agree, then use clang as the compiler instead of gcc.

AlexandraKemperMS: Ok I updated the sample to make it more generic - let me know if that is what you had in mind

docs/cpp/customize-cpp-settings-cpp.md
AlexandraKemperMS: I think we need to be consistent - then we would need to list all the options for every variable. But considering that cstandard, C++ standard, and IntelliSenseMode all have dropdowns available in the Config UI I think people can find this information very easily without us specifying.

docs/cpp/customize-cpp-settings-cpp.md
AlexandraKemperMS: Ah I did not see your follow up comment - windows also has the extra property for sdk version, either one could be valuable. I can switch it to Mac with clang++, that is not a big deal

bobbrow: Ah, yes I forgot about the windows SDK property. I'll leave it up to you how you want to do it since if we only pick one platform we won't be able to show all the properties. (though I don't think it necessary to show all properties so long as they are documented).

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+## Top-level properties
+
+- `env`
+  An array of user-defined variables that are available for substitution in the configurations via the standard environment variable syntax: `${<var>}` or `${env:<var>}`. Strings and arrays of strings are accepted.
+
+- `configurations`
+  An array of configuration objects that provide the IntelliSense engine with information about your project and your preferences. By default, the extension creates a configuration for you based on your operating system. You can also add more configurations.
+
+- `version`
+  We recommend you don't edit this field. It tracks the current version of the `c_cpp_properties.json` file so that the extension knows what properties and settings should be present and how to upgrade this file to the latest version.
```
ntrogh, line 67:
```suggestion
  We recommend that you don't edit this field. It tracks the current version of the `c_cpp_properties.json` file, so that the extension knows what properties and settings should be present and how to upgrade this file to the latest version.
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `enableConfigurationSquiggles`
+  Set to `true` to report errors detected in `c_cpp_properties.json` file to the C/C++ Extension.
+
+## Configuration properties
+
+- `name`
+  A friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that are autoselected on those platforms. The status bar in VS Code shows you which configuration is active. You can also select the label in the status bar to change the active configuration.
```
ntrogh, line 75:
```suggestion
  A user-friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that are autoselected on those platforms. The Status Bar in VS Code shows you which configuration is active. You can also select the label in the Status Bar to change the active configuration.
```

bobbrow (approved):

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+  Platform defaults:
+  - Windows: `windows-msvc-x64`
+  - Linux: `linux-gcc-x64`
+  - macOS: `macos-clang-x64`
+
+  IntelliSense modes that only specify `<compiler>-<architecture>` variants (for example, `gcc-x64`) are legacy modes and are automatically converted to the `<platform>-<compiler>-<architecture>` variants based on the host platform.
+
+- `includePath`
+  An include path is a folder that contains header files (such as `#include "myHeaderFile.h"`) that are included by a source file. Specify a list of paths for the IntelliSense engine to use while searching for included header files. Searching on these paths is not recursive. Specify `/**` at the end of the path to indicate recursive search. For example, `${workspaceFolder}/**` searches through all subdirectories while `${workspaceFolder}` will not. If on Windows with Visual Studio installed, or if a compiler is specified in the `compilerPath` setting, the system include paths should not be listed here.
```
ntrogh, line 96:
```suggestion
  An include path is a folder that contains header files (such as `#include "myHeaderFile.h"`) that are included by a source file. Specify a list of paths for the IntelliSense engine to use while searching for included header files. Searching on these paths is not recursive. Specify `/**` at the end of the path to indicate recursive search. For example, `${workspaceFolder}/**` searches through all subdirectories while `${workspaceFolder}` will not. If you are on Windows with Visual Studio installed, or if a compiler is specified in the `compilerPath` setting, the system include paths should not be listed here.
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+- `forcedInclude`
+  A list of files that should be included before any text in the source file is processed. Files are included in the order listed.
+
+- `compileCommands`
+  The full path to the `compile_commands.json` file for the workspace. If there is a matching entry in `compile_commands.json` for a file open in the editor, that command line is to configure IntelliSense for that file, instead of the other fields of `c_cpp_properties.json`.
+  For more information about the file format, see the [Clang documentation](https://clang.llvm.org/docs/JSONCompilationDatabase.html). Some build systems, such as CMake, [simplify generating this file](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).
+
+- `dotConfig`
+  A path to a .config file created by the Kconfig system. The Kconfig system generates a file with all the defines needed to build a project. Examples of projects that use the Kconfig system are the Linux Kernel and NuttX RTOS.
```
ntrogh, line 129:
```suggestion
  A path to a `.config` file, created by the Kconfig system. The Kconfig system generates a file with all the defines needed to build a project. Examples of projects that use the Kconfig system are the Linux Kernel and NuttX RTOS.
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+- `browse`
+  The set of properties used when `"C_Cpp.intelliSenseEngine"` is set to `"Tag Parser"` (also referred to as "fuzzy" IntelliSense, or the "browse" engine). These properties are also used by the **Go to Definition/Declaration** features, or when the "default" IntelliSense engine is unable to resolve the `#includes` in your source files.
+
+- `recursiveIncludes`
+  A set of properties used to configure how the extension processes an `includePath` entry that specifies a recursive search.
+
+### Browse properties
+
+- `path`
+  A list of paths for the Tag Parser to search for headers included by your source files. If omitted, `includePath` is used as the `path`. Searching on these paths is recursive by default. Specify `*` to indicate nonrecursive search. For example: `${workspaceFolder}` searches through all subdirectories while `${workspaceFolder}/*` will not.
```
ntrogh, line 143:
```suggestion
  A list of paths for the Tag Parser to search for headers included by your source files. If omitted, `includePath` is used as the `path`. Searching on these paths is recursive by default. Specify `*` to indicate nonrecursive search. For example, `${workspaceFolder}` searches through all subdirectories while `${workspaceFolder}/*` does not.
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+  The priority of recursive include path searches when resolving `#include` statements. If set to `beforeSystemIncludes`, the recursive include paths are searched before system include paths. If set to `afterSystemIncludes`, the recursive include paths will be searched after system include paths. `beforeSystemIncludes` would more closely reflect the search order of a compiler leading to more predictability, while `afterSystemIncludes` might result in improved performance.
+
+- `order`
+  Whether subdirectories of recursive includes are searched `breadthFirst` or `depthFirst`.
+
+## Supported variables
+
+You can allow `tasks.json` or `launch.json` to query the current active configuration from `c_cpp_properties.json`. To do this, use the variable `${command:cpptools.activeConfigName}` as an argument in a `tasks.json` or `launch.json` script.
+
+### Default VS Code Settings
```
ntrogh, line 166:
```suggestion
### Default VS Code settings
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+All default VS Code settings, such as C_Cpp.default.includePath, are supported in c_cpp_properties.json. The only exception is
```
ntrogh, line 168:
```suggestion
All default VS Code settings, such as `C_Cpp.default.includePath`, are supported in `c_cpp_properties.json`. The only exception is:
```

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+```json
+C_Cpp.default.systemIncludePath : string[]
+```
+
+This setting allows you to specify the system include path separately from the include path. However, the selected system include path the C++ extension receives from the compiler is not passed to the IntelliSense process. This is only used in rare scenarios since it overwrites the standard compiler behavior, for example, if your compiler is not supported. Instead, use the setting `compilerArgs` and using the "-isystem" flag to specify system headers, which is a better solution in most scenarios.
```
ntrogh, line 174:
```suggestion
This setting allows you to specify the system include path separately from the include path. However, the selected system include path that the C++ extension receives from the compiler is not passed to the IntelliSense process. This is only used in rare scenarios since it overwrites the standard compiler behavior, for example, if your compiler is not supported. Instead, use the setting `compilerArgs` and using the "-isystem" flag to specify system headers, which is a better solution in most scenarios.
```

ntrogh (changes requested):
@AlexandraKemperMS mostly minor text updates.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
+---
+ContentId: 4E34F6AF-BFC6-4BBB-8464-2E50C85AE826
+DateApproved: 4/25/2025
+MetaDescription: How to customize the c_cpp_properties.json file for the C++ extension.
```
sean-mcmanus, line 4: It seems like the "C++ extension" references should be changed to "C/C++ extension".

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Configuration properties
+
+- `name`
+  A friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that are autoselected on those platforms. The status bar in VS Code shows you which configuration is active. You can also select the label in the status bar to change the active configuration.
+
+- `compilerPath`
+  The full path to the compiler you use to build your project, for example `/usr/bin/gcc`, to enable more accurate IntelliSense. The extension queries the compiler to determine the system include paths and default defines to use for IntelliSense.
+
+  Putting `"compilerPath": ""` (empty string) skips querying a compiler. This is useful if your preferred compiler doesn't support the arguments that are used for the query, as the extension defaults to any compiler it can find (like Visual C). Leaving out the `compilerPath` property does not skip the query.
```
sean-mcmanus, line 80: "Visual C" doesn't seem like a valid compiler?

sean-mcmanus (changes requested):

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
+---
+ContentId: 4E34F6AF-BFC6-4BBB-8464-2E50C85AE826
+DateApproved: 4/25/2025
+MetaDescription: How to customize the c_cpp_properties.json file for the C++ extension.
+---
+# C++ extension settings reference
+
+The C++ Extension settings are highly configurable. This article explains the schema for the `c_cpp_properties.json` file. For general information about settings in VS Code, refer to [User and workspace settings](/docs/configure/settings.md), as well as the [Variables reference](/docs/reference/variables-reference.md) and [Default VS Code Settings](/docs/reference/default-settings.md).
```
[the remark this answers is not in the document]
AlexandraKemperMS: FYI I got this link from the python doc's settings page, in this case those should also be updated

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+  A `configurationProvider` candidate extension must implement [vscode-cpptools-api](https://github.com/microsoft/vscode-cpptools-api).
+
+- `mergeConfigurations`
+  Set to `true` to merge include paths, defines, and forced includes with those from a configuration provider.
+
+- `windowsSdkVersion`
+  The versions of the Windows SDK include path to use on Windows, for example `10.0.17134.0`.
+
+- `macFrameworkPath`
+  A list of paths for the IntelliSense engine to use while searching for included headers from Mac frameworks. Only supported on configurations for macOS.
```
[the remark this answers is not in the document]
AlexandraKemperMS: We usually just use MacOS, but yes frameworks is technically correct. Updated.

docs/cpp/customize-cpp-settings.md
AlexandraKemperMS: Updated to MSVC, I think that was meant here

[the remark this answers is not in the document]
AlexandraKemperMS: I updated it to say supported, in the future we can also link to a list of these supported compilers.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
+---
+ContentId: 4E34F6AF-BFC6-4BBB-8464-2E50C85AE826
+DateApproved: 4/25/2025
+MetaDescription: How to customize the c_cpp_properties.json file for the C++ extension.
+---
+# C++ extension settings reference
+
+The C++ extension settings are highly configurable. This article explains the schema for the `c_cpp_properties.json` file. For general information about settings in VS Code, refer to [Configure settings](/docs/getstarted/personalize-vscode.md#configure-settings), as well as the [Variables reference](/docs/reference/variables-reference.md) and [Default VS Code Settings](/docs/reference/default-settings.md).
```
sean-mcmanus, line 8: Is the `Default VS Code Settings` supposed to match the title in the doc `Default settings reference`?

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,174 @@
...
+
+Looking to get started with configuring your C++ project? Begin with [configure Intellisense](/docs/cpp/configure-intellisense.md).
+## Example of variables
```
sean-mcmanus, line 11: It seems like there should be a blank newline before the `##` line.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+  An array of user-defined variables that are available for substitution in the configurations via the standard environment variable syntax: `${<var>}` or `${env:<var>}`. Strings and arrays of strings are accepted.
+
+- `configurations`
+  An array of configuration objects that provide the IntelliSense engine with information about your project and your preferences. By default, the extension creates a configuration for you based on your operating system. You can also add more configurations.
+
+- `version`
+  We recommend that you don't edit this field. It tracks the current version of the `c_cpp_properties.json` file, so that the extension knows what properties and settings should be present and how to upgrade this file to the latest version.
+
+- `enableConfigurationSquiggles`
+  Set to `true` to report errors detected in `c_cpp_properties.json` file to the C/C++ extension.
```
sean-mcmanus, line 71: " C/C++ extension"->C++ extension.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+- `browse`
+  The set of properties used when `"C_Cpp.intelliSenseEngine"` is set to `"Tag Parser"` (also referred to as "fuzzy" IntelliSense, or the "browse" engine). These properties are also used by the **Go to Definition/Declaration** features, or when the "default" IntelliSense engine is unable to resolve the `#includes` in your source files.
+
+- `recursiveIncludes`
+  A set of properties used to configure how the extension processes an `includePath` entry that specifies a recursive search.
+
+### Browse properties
+
+- `path`
+  A list of paths for the Tag Parser to search for headers included by your source files. If omitted, `includePath` is used as the `path`. Searching on these paths is recursive by default. Specify `*` to indicate nonrecursive search. For example, `${workspaceFolder}` searches through all subdirectories while `${workspaceFolder}/*` does not.
```
sean-mcmanus, line 144: The reference to `Tag Parser` might not be understood by readers. Also, it doesn't just use the "path" to search for headers -- the source files on the path are also tag parsed. Could you reword it to something like "A list of paths that have source files parsed for use in global symbol searches."?

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+
+- `limitSymbolsToIncludedHeaders`
+  When true, the Tag Parser only parses code files that are directly or indirectly included by a source file in `${workspaceFolder}`. When false, the Tag Parser parses all code files found in the paths specified in the `browse.path` list.
```
sean-mcmanus, line 147: The references to `true` and `false` should use backticks since they are literals. The existing text is also confusing/misleading because it doesn't mention that the setting only applies to header files and not source files, i.e. it should say something in regards to header files not referenced in the workspace folder are not used for global symbol search.

docs/cpp/customize-cpp-settings.md
sean-mcmanus: Also, in the future, the database may contain non-Tag Parser data, i.e. IntelliSense-based data, so by not referencing "Tag Parser" these docs will be future proof against that potential change.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+
+- `databaseFilename`
+  The path to the generated symbol database. This property instructs the extension to save the Tag Parser's symbol database somewhere other than the workspace's default storage location. If a relative path is specified, it is made relative to the workspace's default storage location, not the workspace folder itself. The `${workspaceFolder}` variable can be used to specify a path relative to the workspace folder (for example `${workspaceFolder}/.vscode/browse.vc.db`)
```
sean-mcmanus, line 150: Tag Parser's symbol could just be `workspace symbol`.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+  For more information about the file format, see the [Clang documentation](https://clang.llvm.org/docs/JSONCompilationDatabase.html). Some build systems, such as CMake, [simplify generating this file](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).
+
+- `dotConfig`
+  A path to a `.config` file, created by the Kconfig system. The Kconfig system generates a file with all the defines needed to build a project. Examples of projects that use the Kconfig system are the Linux Kernel and NuttX RTOS.
+
+- `customConfigurationVariables`
+  Custom variables that can be queried through the command `${cpptools:activeConfigCustomVariable}` to use for the input variables in `launch.json` or `tasks.json`.
+
+- `browse`
+  The set of properties used when `"C_Cpp.intelliSenseEngine"` is set to `"Tag Parser"` (also referred to as "fuzzy" IntelliSense, or the "browse" engine). These properties are also used by the **Go to Definition/Declaration** features, or when the "default" IntelliSense engine is unable to resolve the `#includes` in your source files.
```
sean-mcmanus, line 136: The docs on `browse` are not correct. The browse settings are used in conjunction with IntelliSense for global symbol searches. The reference to the "Tag Parser" should be removed (i.e. the entire sentence).

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+- `windowsSdkVersion`
+  The versions of the Windows SDK include path to use on Windows, for example `10.0.17134.0`.
+
+- `macFrameworkPath`
+  A list of paths for the IntelliSense engine to use while searching for included headers from Mac frameworks. Only supported on configurations for macOS frameworks.
+
+- `forcedInclude`
+  A list of files that should be included before any text in the source file is processed. Files are included in the order listed.
+
+- `compileCommands`
```
sean-mcmanus, line 125: compileCommands can also be an array of paths.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+  The full path to the `compile_commands.json` file for the workspace. If there is a matching entry in `compile_commands.json` for a file open in the editor, that command line is to configure IntelliSense for that file, instead of the other fields of `c_cpp_properties.json`.
```
sean-mcmanus, line 126: `is to` should be something like `is used to`

sean-mcmanus (changes requested):

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+- `version`
+  We recommend that you don't edit this field. It tracks the current version of the `c_cpp_properties.json` file, so that the extension knows what properties and settings should be present and how to upgrade this file to the latest version.
+
+- `enableConfigurationSquiggles`
+  Set to `true` to report errors detected in `c_cpp_properties.json` file to the C/C++ extension.
+
+## Configuration properties
+
+- `name`
+  A user-friendly name that identifies a configuration. `Linux`, `Mac`, and `Win32` are special identifiers for configurations that are autoselected on those platforms. The Status Bar in VS Code shows you which configuration is active. You can also select the label in the Status Bar to change the active configuration.
```
sean-mcmanus, line 76: I don't think `Status Bar` is supposed to be capitalized.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+
+  Platform defaults:
+  - Windows: `windows-msvc-x64`
+  - Linux: `linux-gcc-x64`
+  - macOS: `macos-clang-x64`
+
+  IntelliSense modes that only specify `<compiler>-<architecture>` variants (for example, `gcc-x64`) are legacy modes and are automatically converted to the `<platform>-<compiler>-<architecture>` variants based on the host platform.
+
+- `includePath`
+  An include path is a folder that contains header files (such as `#include "myHeaderFile.h"`) that are included by a source file. Specify a list of paths for the IntelliSense engine to use while searching for included header files. Searching on these paths is not recursive. Specify `/**` at the end of the path to indicate recursive search. For example, `${workspaceFolder}/**` searches through all subdirectories while `${workspaceFolder}` will not. If you are on Windows with Visual Studio installed, or if a compiler is specified in the `compilerPath` setting, the system include paths should not be listed here.
```
sean-mcmanus, line 97: This example is confusing, because it is not an example of an include path and not an example of a header file, it's an example of an include directive. If you want to use the example still, it seems like it should explain what it's an example of.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+  A list of preprocessor definitions for the IntelliSense engine to use while parsing files. Optionally, use `=` to set a value, for example `VERSION=1`.
+
+- `cStandard`
+  The version of the C language standard to use for IntelliSense. For example, `c17`, `gnu23`, or `${default}`. Note: GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense emulates the equivalent C standard version.
+
+- `cppStandard`
+  The version of the C++ language standard to use for IntelliSense. For example, `c++20`, `gnu++23`, or `${default}`. Note: GNU standards are only used to query the set compiler to get GNU defines, and IntelliSense emulates the equivalent C++ standard version.
+
+- `configurationProvider`
+  The ID of a VS Code extension that can provide IntelliSense configuration information for source files. For example, use the VS Code extension ID `ms-vscode.cmake-tools` to provide configuration information from the CMake Tools extension. If you specified a configurationProvider, the configuration it provides takes precedence over your other settings in `c_cpp_properties.json`.
```
sean-mcmanus, line 109: `configurationProvider` should be in backticks since it's a literal reference to the property.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+  A `configurationProvider` candidate extension must implement [vscode-cpptools-api](https://github.com/microsoft/vscode-cpptools-api).
+
+- `mergeConfigurations`
+  Set to `true` to merge include paths, defines, and forced includes with those from a configuration provider.
+
+- `windowsSdkVersion`
+  The versions of the Windows SDK include path to use on Windows, for example `10.0.17134.0`.
+
+- `macFrameworkPath`
+  A list of paths for the IntelliSense engine to use while searching for included headers from Mac frameworks. Only supported on configurations for macOS frameworks.
```
sean-mcmanus, line 120: I'm not understanding what `Only supported on configurations for macOS frameworks.` is supposed to mean.

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+- `configurationProvider`
+  The ID of a VS Code extension that can provide IntelliSense configuration information for source files. For example, use the VS Code extension ID `ms-vscode.cmake-tools` to provide configuration information from the CMake Tools extension. If you specified a configurationProvider, the configuration it provides takes precedence over your other settings in `c_cpp_properties.json`.
+
+  A `configurationProvider` candidate extension must implement [vscode-cpptools-api](https://github.com/microsoft/vscode-cpptools-api).
+
+- `mergeConfigurations`
+  Set to `true` to merge include paths, defines, and forced includes with those from a configuration provider.
+
+- `windowsSdkVersion`
+  The versions of the Windows SDK include path to use on Windows, for example `10.0.17134.0`.
```
sean-mcmanus, line 117: versions->version

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+        }
+    ],
+    "version": 4,
+    "enableConfigurationSquiggles": true
+}
+```
+
+## Top-level properties
+
+- `env`
```
sean-mcmanus, line 61: Is the formatting for these properties correct? It seems sort of hard to read to me.

image

The property name looks visually indistinguishable from a backtick literal reference:

`enableConfigurationSquiggles` Set to `true` to report errors detected in `c_cpp_properties.json` file to the C/C++ extension.

Should there be a `:` added afterwards or something to render it differently from a backticked literal?

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+## Supported variables
+
+You can allow `tasks.json` or `launch.json` to query the current active configuration from `c_cpp_properties.json`. To do this, use the variable `${command:cpptools.activeConfigName}` as an argument in a `tasks.json` or `launch.json` script.
+
+### Default VS Code settings
+
+All default VS Code settings, such as `C_Cpp.default.includePath`, are supported in `c_cpp_properties.json`. The only exception is:
+
+```json
+C_Cpp.default.systemIncludePath : string[]
```
sean-mcmanus, line 172: Why does this have a red background?

docs/cpp/customize-cpp-settings.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+This setting allows you to specify the system include path separately from the include path. However, the selected system include path that the C++ extension receives from the compiler is not passed to the IntelliSense process. This is only used in rare scenarios since it overwrites the standard compiler behavior, for example, if your compiler is not supported. Instead, use the setting `compilerArgs` and using the "-isystem" flag to specify system headers, which is a better solution in most scenarios.
```
sean-mcmanus, line 175: It seems like `"-isystem"` should use backticks instead of quotation marks -- other example compiler args above use backticks.

docs/cpp/customize-cpp-settings.md
AlexandraKemperMS: No reason - it shows up normally (without red) in VS Code for me as well as in the actual docs preview, must be something by github that is specific for PRs. Maybe because it is JSON? It will not show up in the actual published doc

docs/cpp/customize-cpp-settings.md
AlexandraKemperMS: I updated them to add a ":" after each, it looks better and matched the copilot settings now

docs/cpp/customize-cpp-settings.md
AlexandraKemperMS: Removed - we support this on all OS now

docs/cpp/customize-cpp-settings.md
AlexandraKemperMS: Status Bar is the official name so it is capitalized, leaving as capitals.

ntrogh (approved):

PR #8338, opened 2025-05-02
Replace broken links with working links

Rashi455:
- The invalid extensions.md link was replaced with with the azure extensions information in overview.md for every instance
- replaced download link with VSCode download page for every instance
- replaced /updates/v1_34.md#machinespecific-settings with /release-notes/v1_34.md#Machine-specific-settings for every instance


Rashi455:

ntrogh (changes requested):
@Rashi455 Thanks for the updates! 
The `/download` relative links should remain in place and the changes to `/docs/cpp/customize-default-settings-cpp.md` are part of another active PR. 

Can you revert your changes for these?

PR #8339, opened 2025-05-02
Fix incorrect cp command in remote toolchain example

Monologue2:
This PR fixes a syntax error in the remote FAQ documentation related to `cp` usage.

The original line was:
```bash
cp <path-to-config-file> > .config
```

This is invalid syntax and results in the following error:

```bash
cp: missing destination file operand
```

It has been corrected to:
```bash
cp <path-to-config-file> .config
```

This change aligns the example with standard cp command usage and prevents confusion during setup.

Thanks!


ntrogh (approved):
@Monologue2 Thanks for fixing this.

PR #8344, opened 2025-05-06
Fix spelling in VSCode Web docs

therealparmesh:
See commit.


ntrogh (approved):
@therealparmesh Thanks for correcting this!

PR #8349, opened 2025-05-07
Update Baseline feature description

rviscomi:
- clarify that the updated browser support info also applies to HTML attributes
- add a link to the MDN doc about Baseline for more info


rviscomi:
FYI @aeschli

ntrogh (approved):
Thanks for clarifying @rviscomi

PR #8367, opened 2025-05-12
Add AI Toolkit model conversion docs

ssss141414:
- Add a documentation for model conversion feature in AI Toolkit, which is going to be released soon.
- Remove lagacy model conversion doc to reference.


ntrogh:
@ssss141414 **Important:** we only support a single sub-level within the table of contents. so the reference sub-node is not compatible.

ssss141414:
> @ssss141414 **Important:** we only support a single sub-level within the table of contents. so the reference sub-node is not compatible.

Sure. Will fix.

PR #8390, opened 2025-05-17
Update quickstart-python.md: use stable Django docs link

collinanderson:
The 3.0 docs give a security warning. Use "stable" which redirects to latest version.


ntrogh (approved):
@collinanderson Thanks for correcting this!

Issue #8391, opened 2025-05-17
Contradictions in the docs


racz16:
I found some contradictions in the docs. Could you clarify which information is correct and update the docs accordingly?

**`keywords` field in `package.json`**

The [Extension Manifest](https://code.visualstudio.com/api/references/extension-manifest#fields) page of the docs says:

> An array of **keywords** to make it easier to find the extension. These are included with other extension **Tags** on the Marketplace. This list is currently limited to 5 keywords.

The [Publishing Extensions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#i-get-a-you-exceeded-the-number-of-allowed-tags-of-10-error-when-i-try-to-publish-my-extension) page of the docs says:

> The Visual Studio Marketplace does not allow an extension package to have more than 10 `keywords` in the `package.json`. Keep the number of keywords/tags to less than 10 to avoid this error.

So, the 1st page says that the limit is 5, the 1st sentence of the 2nd page says that the limit is 10, while the 2nd sentence of the 2nd page implies that the limit is 9. At the same time, some extensions, for example, [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) clearly have more `keywords` than any of these limits.

**`icon` field in `package.json`**

The [Extension Manifest](https://code.visualstudio.com/api/references/extension-manifest#fields) page of the docs says:

> The path to the icon of at least 128x128 pixels (256x256 for Retina screens).

The [Publishing Extensions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#marketplace-integration) page of the docs says:

> Set an icon by specifying a relative path to a 128x128px PNG file included in your extension via the `icon` property in `package.json`.

The 1st page says **at least** 128x128px, and the 2nd page says specifically 128x128px.

Thank you

DhruvTheDev1:
Hi,
Good find! I just had a look at Keywords limit information.
I believe that Visual Studio Code currently allows up to 10 keywords in the package.json and the wording "less than 10" is probably a caution to avoid going over the limit. The "5 keywords" limit seems to be outdated information.

Also, prettier may show more tags as the marketplace can add extra tags automatically.

DhruvTheDev1:
@ntrogh Ahh okay! Thank you!!

PR #8417, opened 2025-05-27
Updated keywords limit from 5 to 10

DhruvTheDev1:
Updated the keyword limit from 5 to 10, as the "5 keywords" limit appears to be outdated.
Focuses on Issue [#8391 ](https://github.com/microsoft/vscode-docs/issues/8391)


ntrogh:
@DhruvTheDev1 Thanks for your support, however this information is incorrect and should be 30. I've already filed a PR to correct the documentation.

PR #8434, opened 2025-05-31
fix: swap "experimental" settings reference to "preview"

shouples:
Replaces an "experimental settings" reference with "preview settings" in the description of the **Preview** state to match the focus of the section (since experimental settings are focused in the previous section).


ntrogh (approved):
@shouples Thanks for fixing this oversight! Appreciate it.

PR #8437, opened 2025-06-02
[mcp-servers] Switch the header example to something more likely 

julien-c:
i expect most keys will be passed as `Authorization: Bearer` headers, so this would make the doc simpler to understand


isidorn:
Thank you. Makes sense. Though let's see if @digitarald or @connor4312 have any feedback.

connor4312:
I'm fine keeping the example header as API_KEY. Specifically for authorization, newly implemented servers should tend to use MCP's Authorization capability rather than explicit Authorization headers.

digitarald:
Thanks for contributing more realistic docs.

I agree with Connor's point, to not showcase auth via keys; and we should even call it out for this line as best practice.

ntrogh:
@julien-closing this PR, given the above feedback. I'll update the docs to include additional guidance about authorization.

julien-c:
that works, thanks for reviewing!

Issue #8150, opened 2025-03-14
Update terminal shell integration docs


Tyriar:
Changes:

- Include `HasRichCommandDetection`
- Explain difference between rich and basic shell integration

Sonichigo:
Can i take this issue up? and I'm sorry to raising the PR directly. Feel free to close the current pr if necessary.

Tyriar:
@Sonichigo responded on the PR

PR #8450, opened 2025-06-05
Document HasRichCommandDetection and shell integration quality

Tyriar:
Fixes https://github.com/microsoft/vscode-docs/issues/8150

### DIFF
--- a/docs/terminal/shell-integration.md
+++ b/docs/terminal/shell-integration.md
@@ -87,7 +87,7 @@ When using shell integration, it has a "quality" associated with it that declare
 
 - **None**: No shell integration is active.
 - **Rich**: Shell integration is active and command detection is working in an ideal way.
-- **Basic**: Shell integration is active but command detection may not be ideal. This typically means the shell has very basic command detection that knows where a command was run but not what it is or what the exit status was.
+- **Basic**: Shell integration is active, but command detection might not support all functionality. For example, the command run location is detected, but not its exit status.
 
 To view the shell integration quality, hover the terminal tab. Optionally, select **Show Details** on the hover to view more detailed information.
 


docs/terminal/shell-integration.md
```diff
@@ -81,6 +81,16 @@ code --locate-shell-integration-path bash
 [[ "$TERM_PROGRAM" == "vscode" ]] && . "/path/to/shell/integration/script.sh"
 ```
 
+## Shell integration quality
+
+When using shell integration, it has a "quality" associated with it that declares the capabilities of it. These qualities are determined by how the shell integration script behaves.
+
+- **None**: No shell integration is active
```
ntrogh:
```suggestion
- **None**: No shell integration is active.
```

ntrogh: Consistent punctuation with the next items in the list

docs/terminal/shell-integration.md
```diff
@@ -81,6 +81,16 @@ code --locate-shell-integration-path bash
...
+- **Rich**: Shell integration is active and command detection is working in an ideal way.
```
ntrogh, line 89: The use of "ideal" is vague - is there a way to make it more explicit what is supported?

docs/terminal/shell-integration.md
```diff
@@ -81,6 +81,16 @@ code --locate-shell-integration-path bash
...
+- **Basic**: Shell integration is active but command detection may not be ideal. This typically means the shell has very basic command detection that knows where a command was run but not what it is or what the exit status was.
```
ntrogh:
```suggestion
- **Basic**: Shell integration is active, but command detection might be limited: the command run location is detected, but not its exit status.
```

ntrogh: Suggested rewording - if you prefer to keep yours, do replace "may" with "might".

docs/terminal/shell-integration.md
```diff
@@ -81,6 +81,16 @@ code --locate-shell-integration-path bash
...
+
+To view the shell integration quality, hover the terminal tab. Optionally click "Show Details" on the hover to view more detailed information.
```
ntrogh:
```suggestion
To view the shell integration quality, hover the terminal tab. Optionally, select **Show Details** on the hover to view more detailed information.
```

docs/terminal/shell-integration.md
```diff
@@ -249,6 +259,8 @@ These sequences should be ignored by other terminals, but unless other terminals
 
   - `Cwd` - Reports the current working directory to the terminal.
   - `IsWindows` - Indicates whether the terminal is using a Windows backend like winpty or conpty. This may be used to enable additional heuristics as the positioning of the shell integration sequences are not guaranteed to be correct. Valid values are `True` and `False`.
+  - `HasRichCommandDetection` - Indicates whether the terminal has rich command detection capabilities. This property is set to `True` when the shell integration script acts ideally as VS Code expects it, specifically sequences should come in the expected positions in the order `A, B, E, C, D`.
```
ntrogh: Might be good to align the style of this list with the other list (use of colon instead of dash to separate value and description) - you can keep inline code-style because we're referring to property values.

ntrogh (commented):
@Tyriar Added a couple of minor text suggestions.

docs/terminal/shell-integration.md
Tyriar: It's kind of elaborated below. Explaining what ideal means here is too technical imo as this section of the page is targeting a user of shell integration, not a shell developer

docs/terminal/shell-integration.md
Tyriar: I read that as this is exactly the case when it's basic but that's not true, basic is a spectrum from very basic up until rich.

docs/terminal/shell-integration.md
Tyriar: Up to you, but this is the format for the whole of the Supported escape sequences section. The reason it's like this is because the source of truth is a ts file and `Id`: isn't as clear when read as code:

https://github.com/microsoft/vscode/blob/eb0eb3c57903986589bbf9396eb4c89259e729e0/src/vs/platform/terminal/common/xterm/shellIntegrationAddon.ts#L208-L217

ntrogh: It's somewhat jarring to see different list styles in the same doc. I don't suggest changing the code if it hampers legibility there. However, for docs, I'd prefer consistent styles (ack that there's many occurrences across our docs that need to be fixed). Maybe a quick Copilot prompt to align style?

docs/terminal/shell-integration.md
ntrogh:
```suggestion
- **Basic**: Shell integration is active, but command detection might not support all functionality. For example, the command run location is detected, but not its exit status.
```

bpasero (approved):

PR #8451, opened 2025-06-06
Fix broken link in nodejs-debugging.md

alefteris:


DhruvTheDev1 (commented):
Nice fix! Confirmed that the updated link correctly points to the right section.

ntrogh (approved):
@alefteris Thanks for fixing the broken link!

PR #8454, opened 2025-06-08
Remove unrelated para from "AI ... Convert Format String ..." section

kazarmy:
This PR removes an unrelated paragraph from the "Contributions to extensions -> Python -> AI Code Actions: Convert Format String (Experimental)" section under the 1.100 release notes. The removal is because it doesn't appear that one should get a new symbol when converting Python string concatenations to f-string or format(), much less a class or function, and "... let AI handle the implementation" -- what would AI implement?

### DIFF
--- a/release-notes/v1_100.md
+++ b/release-notes/v1_100.md
@@ -516,8 +516,6 @@ This experience is enabled via the following setting:
 "python.analysis.aiCodeActions": {"convertFormatString": true}
 ```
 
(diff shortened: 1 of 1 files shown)


ntrogh:
@kazarmy Yep, seems unrelated and was already mentioned in 1.97 release. Thanks.

ntrogh (approved):

PR #8458, opened 2025-06-10
Update copilot-customization.md

jirihofman:
The doc suggests `Chat: New Instruction File` but the actual command is `Chat: New Instructions File`.

### Docs
image

### VSC
image


DhruvTheDev1 (commented):
Nice find! Also good use of the screenshot to demonstrate the correct command which is `Chat: New Instructions File` (added s)

PR #8501, opened 2025-06-19
Fix incorrect mention of security issues in 1.101.1 release notes

JohnVeness:
I've only glanced at them, but the listed issues seem to be "normal" issues to me, not security issues.

### DIFF
--- a/release-notes/v1_101.md
+++ b/release-notes/v1_101.md
@@ -11,7 +11,7 @@ DownloadVersion: 1.101.1
 
 _Release date: June 12, 2025_
 
-**Update 1.101.1**: The update addresses these security [issues](https://github.com/microsoft/vscode/issues?q=is%3Aissue+is%3Aclosed+milestone%3A%22May+2025+Recovery+1%22+).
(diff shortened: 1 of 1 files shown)


JohnVeness:
Thanks @ntrogh :).

Assuming this happened due to a copy and paste from previous release notes, I wonder if it might be better if, when security issues *are* mentioned, to make both of the words "security issues" the link. E.g. like:

The update addresses these [security issues](https://github.com/microsoft/vscode/issues?q=is%3Aissue+is%3Aclosed+milestone%3A%22April+2025+Recovery+1%22+).

rather than the current convention of:

The update addresses these security [issues](https://github.com/microsoft/vscode/issues?q=is%3Aissue+is%3Aclosed+milestone%3A%22April+2025+Recovery+1%22+).

That way, the word "security" stands out more, and might be more noticeable (thus more likely to be removed when not needed) when copying and pasting into new release notes.

ntrogh:
@JohnVeness Indeed, the curse of copy-paste and routine 🤦. That's actually a good suggestion!

PR #8512, opened 2025-06-20
Fix anchor in `contribution-points.md`

outloudvi:
Somehow the actual anchors now have dots removed.


ntrogh:
@outloudvi Thanks for raising this. This is due to a bug in the publishing pipeline. Fixing this now, so this PR is no longer needed.

outloudvi:
Thanks, @ntrogh! Hopefully the fix can be published soon, so that people browsing over configuration points can have a easier time.

ntrogh:
@outloudvi We had some other changes in the pipeline which needed verification. The fix should be live later today.

PR #8686, opened 2025-07-30
Clarify you can disable Copilot via settings

remcohaszing:
There are some specific instructions to disable Copilot. However, these instructions don’t work if the user has set `window.titleBarStyle` to `native`. This change documents an alternative that works for all users.

### DIFF
--- a/docs/copilot/faq.md
+++ b/docs/copilot/faq.md
@@ -60,7 +60,7 @@ See [Use a different GitHub account with Copilot](/docs/copilot/setup.md#use-a-d
 
 ### How can I remove Copilot from VS Code?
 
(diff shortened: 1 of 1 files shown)


AugustinMauroy (approved):
LGTM ! BTW tested on :
```
Version: 1.102.2 (Universal)
Commit: c306e94f98122556ca081f527b466015e1bc37b0
Date: 2025-07-22T12:15:48.520Z (1 wk ago)
Electron: 35.6.0
ElectronBuildId: 11847422
Chromium: 134.0.6998.205
Node.js: 22.15.1
V8: 13.4.114.21-electron.0
OS: Darwin arm64 24.5.0
```

and work as expected !

ntrogh:
@remcohaszing Using the command from the Command Palette should work regardless of the `window.titleBarStyle`. Also, setting the `chat.commandCenter.enabled` only hides the item but doesn't actually remove all Copilot-related experiences.

PR #8936, opened 2025-09-26
Redirecting to julialang.org/install for recommended way of installin…

rkube:
The recommended way to install julia is through juliaup.
This patch modifies the plugin readme to point to the updated instructions


ntrogh (approved):
@rkube Thanks for contributing

PR #8981, opened 2025-10-17
Fix todo list setting reference in agent mode documentation

demershov:
The setting for controlling the visibility and position of the todo list control in chat has been renamed.

**Documentation updates for todo list setting:**

* Updated the configuration key for the todo list widget from `chat.todoListWidget.position` to `chat.agent.todoList.position` in both the usage documentation (`chat-agent-mode.md`) and the settings reference (`copilot-settings.md`). 

* Updated the default value for the `chat.agent.todoList.position` in the settings reference to `"default"`.


demershov:

ntrogh:
@demershov Thanks for updating these settings IDs!

PR #8983, opened 2025-10-20
Fix code sample of `ignoreSync` contribution point

cjohnson19:
Removes a duplicate example for `ignoreSync` which had invalid JSON and incorrect markdown formatting for a code block.


ntrogh (approved):
@cjohnson19 Thanks for updating this!

PR #8984, opened 2025-10-20
Clarify the location of md files for Copilot Chat customization within user profiles

CraftyCraftress:
This change clarifies:
- That md files for customizing Copilot Chat need to go in the prompts sub-directory of a profile.
- Clarifies where files associated with the default profile are stored.

### DIFF
--- a/docs/configure/profiles.md
+++ b/docs/configure/profiles.md
@@ -405,13 +405,13 @@ If the profile specified does not exist, a new empty profile with the given name
 
 ### Where are profiles kept?
 
-The Default Profile is stored in the root User configuration folder similar to your user settings and keyboard shortcuts.
+Files associated with the Default Profile are stored in the User configuration folder.
 
 * **Windows** `%APPDATA%\Code\User`
 * **macOS** `$HOME/Library/Application\ Support/Code/User`
 * **Linux** `$HOME/.config/Code/User`
 
-Additional profiles are stored under the `profiles` sub-directory.
+Additional profiles are stored in the `profiles` subdirectory within your User configuration folder:
 
 * **Windows** `%APPDATA%\Code\User\profiles`
 * **macOS** `$HOME/Library/Application\ Support/Code/User/profiles`


ntrogh:
@CraftyCraftress We don't include this information as we don't recommend users create those files directly in their user profile. The location of the files in the user profile might be subject to change.

CraftyCraftress:
> @CraftyCraftress We don't include this information as we don't recommend users create those files directly in their user profile. The location of the files in the user profile might be subject to change.

@ntrogh  What is the suggested way to create & sync prompts/instructions across machines that are not repo/workspace specific?

ntrogh:
@CraftyCraftress When you create a prompt/instructions file (**Chat: New Prompt File** or **Chat: New Instructions File**), you can choose between workspace or user profile.

For user-profile prompts or instructions, they are subject to [settings sync](https://code.visualstudio.com/docs/configure/settings-sync), so will roam with you across machines.

image

CraftyCraftress:
> @CraftyCraftress When you create a prompt/instructions file (**Chat: New Prompt File** or **Chat: New Instructions File**), you can choose between workspace or user profile.
> 
> For user-profile prompts or instructions, they are subject to [settings sync](https://code.visualstudio.com/docs/configure/settings-sync), so will roam with you across machines.

@ntrogh Thanks! As a bit of feedback, it has taken longer than it should have for multiple people on my team to figure out how to manage instructions we want to share across machines.

I think there's a couple reasons why:

1. The text on pages like [custom instructions](https://code.visualstudio.com/docs/copilot/customization/custom-instructions) talks about the file system location for instructions files in workspaces and then says, "User instructions files are stored in the [current profile folder](https://code.visualstudio.com/docs/configure/profiles)" with a link to the profiles page. The context & wording makes me think about managing instructions in my user profile as files in a folder.

2. As a developer, I want to be able to browse the instructions as files in a folder - not having to use different commands to add an instructions file, edit a current instructions file, rename an instructions file, etc... I also want to easily be able to move instructions files between my user profile and the workspace (or other locations).

In my opinion, it would be better for the documentation to just tell devs where the files are with the caveat that the location may change instead of trying to abstract it away and providing a sub-par experience that artificially introduces overhead.

ntrogh:
@CraftyCraftress Thanks for providing additional input. Reopening PR.

msabry613-ops (approved):

msabry613-ops (approved):

msabry613-ops (changes requested):
nnnn

msabry613-ops (approved):

msabry613-ops (approved):

ntrogh:
@CraftyCraftress We've has some docs updates in the meantime that cause a merge conflict. Could you resolve? Otherwise LGTM!

msabry613-ops:
No

في الثلاثاء، ٢٨ أكتوبر ٢٠٢٥ ١٠:٥٣ م Nick Trogh ***@***.***>
كتب:

PR #9005, opened 2025-10-29
Fix typo in tools configuration instruction

mraafayv:
Change 'then' to 'them' for clarity


ntrogh:
@mraafayv Thanks for fixing!

PR #9032, opened 2025-11-07
MCP server gallery

ntrogh:


docs/setup/enterprise.md
```diff
@@ -94,6 +94,24 @@ Duplicate key values are not supported. For example, including both `"microsoft"
 
 If you want to learn more about extensions in VS Code, refer to the [extensions documentation](/docs/configure/extensions/extension-marketplace.md).
 
+## Configure MCP server access
+
+By default, VS Code allows developers to [add any MCP server](/docs/copilot/customization/mcp-servers.md) to their environment and retrieves the list of MCP servers in the Extensions view from the GitHub MCP registry.
+
+Organizations can restrict which MCP servers are allowed to be used by developers in VS Code. The following configuration options are available:
+
+* Configure from which source MCP servers can be added (controls the `setting(chat.mcp.access)` VS Code setting)
```
sandy081: This is not about adding MCP servers. Irrespective of what value this setting is set, users can add MCP servers from any source. This setting controls which MCP servers are enabled in VS Code.

docs/setup/enterprise.md
```diff
@@ -94,6 +94,24 @@ Duplicate key values are not supported. For example, including both `"microsoft"
 
 If you want to learn more about extensions in VS Code, refer to the [extensions documentation](/docs/configure/extensions/extension-marketplace.md).
 
+## Configure MCP server access
+
+By default, VS Code allows developers to [add any MCP server](/docs/copilot/customization/mcp-servers.md) to their environment and retrieves the list of MCP servers in the Extensions view from the GitHub MCP registry.
```
sandy081: GitHub MCP registry is still behind preview flag therefore most of the users do not see it unless they opt in. So might confuse users if the registry is mentioned as they might have not seen it.

ntrogh: @sandy081 Isn't the MCP gallery (from GH MCP registry) available to all users in 1.106? In that case, it does make sense to mention that in the Extensions view the MCP servers are coming from there.

sandy081: Not yet. It is kept behind the flag still in 1.106

ntrogh: Updated the docs

ntrogh:
@sandy081 I updated the docs based on your feedback

sandy081:
@ntrogh Thanks for the updates. LGTM

PR #9064, opened 2025-11-14
FAQ for cmd not in terminal

Tyriar:
People have been confused about this so it needs a FAQ somewhere.

### DIFF
--- a/docs/copilot/reference/copilot-settings.md
+++ b/docs/copilot/reference/copilot-settings.md
@@ -173,20 +173,6 @@ The [Chat Sessions view](/docs/copilot/copilot-coding-agent.md) provides a centr
 | `setting(accessibility.voice.autoSynthesize)`<br/>Controls whether a textual response should automatically be read out aloud when speech was used as input. | `"off"` |
 | `setting(accessibility.voice.speechTimeout)`<br/>The duration in milliseconds that voice speech recognition remains active after you stop speaking. | `1200` |
 
-## Frequently asked questions
-
-### Why isn't Copilot using Command Prompt as the terminal shell?
-
-The agent uses the shell you have configured as the default for the terminal, except when it's cmd. This is because [shell integration](https://code.visualstudio.com/docs/terminal/shell-integration) is not supported with Command Prompt, which means the agent has very limited visibility into what's going on inside the terminal. Instead of getting direct signals for when commands are being run or have finished running, the agent needs to rely on timeouts and watching for the terminal to idle to continue. This leads to a slow and flaky experience.
-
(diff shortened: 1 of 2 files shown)


docs/copilot/reference/copilot-settings.md
```diff
@@ -173,6 +173,20 @@ The [Chat Sessions view](/docs/copilot/copilot-coding-agent.md) provides a centr
 | `setting(accessibility.voice.autoSynthesize)`<br/>Controls whether a textual response should automatically be read out aloud when speech was used as input. | `"off"` |
 | `setting(accessibility.voice.speechTimeout)`<br/>The duration in milliseconds that voice speech recognition remains active after you stop speaking. | `1200` |
 
+## Frequently asked questions
+
+### Why isn't Copilot using Command Prompt as the terminal shell?
+
+Copilot will use whatever shell you have configured as the default for the terminal unless it's cmd specifically. The reason for this is because [shell integration](https://code.visualstudio.com/docs/terminal/shell-integration) cannot be supported with Command Prompt which means Copilot has very limited visibility into what's going on inside the terminal. Instead of getting direct signals for when commands are being run or have finished running, Copilot needs to rely on timeouts and watching for the terminal to idle to continue. This leads to a slow and flaky experience.
```
ntrogh:
```suggestion
The agent uses the shell you have configured as the default for the terminal, except when it's cmd. This is because [shell integration](https://code.visualstudio.com/docs/terminal/shell-integration) is not supported with Command Prompt, which means the agent has very limited visibility into what's going on inside the terminal. Instead of getting direct signals for when commands are being run or have finished running, the agent needs to rely on timeouts and watching for the terminal to idle to continue. This leads to a slow and flaky experience.
```

docs/copilot/reference/copilot-settings.md
```diff
@@ -173,6 +173,20 @@ The [Chat Sessions view](/docs/copilot/copilot-coding-agent.md) provides a centr
...
+
+Note that it's still possible to configure command prompt if you really want it to be used by using the `setting(chat.tools.terminal.terminalProfile.windows)` setting, just keep in mind you will get an inferior experience compared to PowerShell:
```
ntrogh:
```suggestion
It's still possible to configure Command Prompt to be used with the `setting(chat.tools.terminal.terminalProfile.windows)` setting, however keep in mind that you will get an inferior experience compared to using PowerShell.
```

Tyriar:
> Wouldn't it make more sense to include this in the Tools article's FAQ, as we're discussing terminal tools in this article: https://code.visualstudio.com/docs/copilot/chat/chat-tools#_frequently-asked-questions

I was just trying to find the place that the terminal profile setting was mentioned, I also started writing on an older version of the docs before pulling. Where ever you think is ideal is fine with me, could you finish up the PR by moving where you think is appropriate?

ntrogh:
Will do. Thanks for adding this FAQ

PR #9066, opened 2025-11-14
Fix subcommand

karlhorky:
It appears that the first subcommand on [the VS Code October 2025 (version 1.106) release notes blog post](https://code.visualstudio.com/updates/v1_106) is incomplete, if I'm understanding correctly

But maybe it breaks on the first space too - if so, maybe my PR should change to some words  explaining the breaking on the space.

### DIFF
--- a/release-notes/v1_106.md
+++ b/release-notes/v1_106.md
@@ -306,7 +306,7 @@ Auto approval status has moved from being inline inside the chat view to the too
 
 #### Auto approve parser improvements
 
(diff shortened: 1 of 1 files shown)


ntrogh:
The subcommand would be `echo` because it would consider `"a` as the parameter for the subcommand. I don't think it merits updating the release notes to clarify this further, as the new behavior is different because of the parser approach.

karlhorky:
Ok, thanks for the consideration.

PR #9072, opened 2025-11-17
Add instructions for using multiple GH accounts

ntrogh:
Fixes https://github.com/microsoft/vscode-docs/issues/9068

### DIFF
--- a/docs/copilot/setup.md
+++ b/docs/copilot/setup.md
@@ -95,8 +95,6 @@ Follow these steps to configure which GitHub account to use for Copilot. This co
         ```
 
     1. Re-sign in to your GitHub Enterprise account if you're not already signed in
-    1. Select **GitHub Copilot Chat** from the list of extensions
-    1. Choose the GitHub Enterprise account you want to use for Copilot in the current workspace and profile
 
 ## Remove AI features from VS Code
 


docs/copilot/setup.md
```diff
@@ -47,15 +47,42 @@ If your Copilot subscription is associated with another GitHub account, follow t
...
+    1. Run **Preferences: Open User Settings (JSON)** or **Preferences: Open Workspace Settings (JSON)** from the Command Palette (`kb(workbench.action.showCommands)`)
+    1. Add the following setting to specify GitHub Enterprise as the authentication provider for Copilot:
+
+        ```json
+        "github.copilot.advanced": {
+            "authProvider": "github-enterprise"
+        }
+        ```
+
+    1. In the Accounts menu in the Activity Bar, select **Manage Extension Account Preferences**
```
TylerLeonhardt: It's unlikely the user will have multiple GHE accounts. I think maybe this can be simplified to "re-sign in to Copilot"

docs/copilot/setup.md
```diff
@@ -47,15 +47,42 @@ If your Copilot subscription is associated with another GitHub account, follow t
...
+
+* For GitHub.com accounts:
+
+    1. In the Accounts menu in the Activity Bar, select **Manage Extension Account Preferences**
+    1. Select **GitHub Copilot Chat** from the list of extensions
+    1. Choose the GitHub account you want to use for Copilot in the current workspace and profile
+
+* For GHE.com accounts:
+
+    1. Run **Preferences: Open User Settings (JSON)** or **Preferences: Open Workspace Settings (JSON)** from the Command Palette (`kb(workbench.action.showCommands)`)
```
TylerLeonhardt, line 87: This is exactly what we wanna call out here.

One thing I notice is that this is the first and only reference to signing in to GHE.com in this doc which makes me worried that for folks who want to sign in with their GHE.com account will set this instead of going through the normal flow (Continue with GHE.com)

Image

> NOTE: this sets that setting but at the user level

ntrogh: Thanks for pointing that out. I'll make sure to clarify this more.

ntrogh:
@TylerLeonhardt thanks for your feedback. I've added a dedicated section for signing in with a GHE.com account to avoid confusion with steps for multiple accounts.

docs/copilot/setup.md
```diff
@@ -47,15 +57,46 @@ If your Copilot subscription is associated with another GitHub account, follow t
...
+
+        ```json
+        "github.copilot.advanced": {
+            "authProvider": "github-enterprise"
+        }
+        ```
+
+    1. Re-sign in to your GitHub Enterprise account if you're not already signed in
+    1. Select **GitHub Copilot Chat** from the list of extensions
+    1. Choose the GitHub Enterprise account you want to use for Copilot in the current workspace and profile
```
TylerLeonhardt: These look like leftovers from the previous wording

ntrogh: Ow, I should stop pretending I'm in Pacific timezone... :)

ntrogh:
@TylerLeonhardt Should be the last iteration...

PR #9098, opened 2025-11-24
Add TIP explaining IntelliSense behavior inside comments and string literals

NA-V10:
This PR adds a helpful TIP to `intellisense.md` explaining that IntelliSense does
not appear inside comments or regular string literals unless the language extension
explicitly supports it.

### ✔ Why this is useful
Many users expect IntelliSense everywhere and may think it is broken when typing
inside comments or strings. This clarification improves accuracy and helps avoid
confusion.

## Checklist
- [x] Documentation only (no functional changes)
- [x] Follows VS Code Docs callout format
- [x] Change added in correct section
- [x] Branch created from fork, targeting `main`

### DIFF
--- a/release-notes/v1_106.md
+++ b/release-notes/v1_106.md
@@ -942,7 +942,6 @@ Contributions to our issue tracking:
 
 Contributions to `vscode`:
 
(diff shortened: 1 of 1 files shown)


NA-V10:

ntrogh:
@NA-V10 Closing this PR as the updates don't provide a significant update. Thanks for your effort.

PR #9116, opened 2025-11-28
Fix broken .md links

FluoriteCafe-work:
Some links are broken potentially due to `.md` extension in href links.


ntrogh:
@FluoriteCafe-work Thanks for catching these. There was an issue with our docs publishing pipeline where it didn't strip out the ".md" suffix in relative links. This problem will be solved with the next website publishing.

So closing this PR as it will be resolved at a platform level. Appreciate your support to make the docs better!!

PR #9179, opened 2025-12-17
fix: correct typo from "actuall" to "actually"

adex-hub:
Just a docs change to ensure things are accurate.


adex-hub:
@ntrogh please review.

PR #9199, opened 2025-12-25
Update manually install apt repository

wcbing:
- Write key by gpg to simplify the steps
- The apt-transport-https is a transitional package, moved into the apt in 1.5.
  https://packages.debian.org/sid/apt-transport-https


ntrogh (approved):
Thanks

PR #9201, opened 2025-12-26
docs(reference): add missing when clause context keys

DakotaB75:
This PR updates the when clause context reference to include context keys that are already documented and used across other VS Code extension guides (Testing, SCM, Web Extensions, Workspace Trust), but were missing from the central reference table.

The change is documentation-only and does not introduce new behavior or assumptions.


ntrogh (changes requested):
@DakotaB75 Make sure to use the correct styling for the context items.

ntrogh:
@DakotaB75 When creating a PR, can you reference the original issue in the PR?

Issue #9210, opened 2025-12-31
Documentation for `lineComment` in `language-configuration.json` is out-of-date


eisenwave:
https://github.com/microsoft/vscode-docs/blob/0a4ca79ec77e2daafaa1832f7f2533b1b46c65c1/api/language-extensions/language-configuration-guide.md?plain=1#L28-L31

Following this format gives me a warning:

<img width="517" height="133" alt="Image" src="https://github.com/user-attachments/assets/db932707-90bb-4b7d-bf9f-8117ca432599" />

That is because at some point, the `lineComment` property got expanded from being just a `string` to being an `object` with `comment` and `noIndent` property.

https://github.com/microsoft/vscode/blob/d2894816734dc9141885b802805801b67d70e6e9/src/vs/workbench/contrib/codeEditor/common/languageConfigurationExtensionPoint.ts#L549C5-L565

DakotaB75:
I’ll take this.

The guide still shows lineComment as a string, but the configuration now supports an object form (comment, noIndent).
I’ll update the example and text to reflect the current format and avoid the warning.

ntrogh:
@eisenwave Thanks for reporting this. The language configuration JSON defines `lineComment` as on `object` however the extension API supports both `string` and `object`. So, the language configuration  definition should be updated to `type: ['string', 'object'],` instead. Issue created in vscode repo: https://github.com/microsoft/vscode/issues/285594

In addition, the API docs should be updated to describe both configuration formats (string and object).

DakotaB75:
Documented both supported lineComment formats in the language configuration guide.

PR #9213, opened 2025-12-31
docs: update lineComment examples to reflect current configuration shape

DakotaB75:
The Language Configuration Guide still documents `comments.lineComment` as a string value. The configuration was expanded to support an object-based format (e.g. `comment`, `noIndent`), and following the current examples produces a warning in VS Code. This change updates the documented examples to reflect the current shape, so users can copy them without running into validation warnings.

Ref: https://github.com/microsoft/vscode-docs/issues/9210


ntrogh:
@DakotaB75 Closing this PR

DakotaB75:
Thanks for the feedback understood. I’ll make sure to apply review changes in the original PR going forward.

PR #9297, opened 2026-01-26
chore: Update Known language identifiers

RedCMD:
based on comment
* https://github.com/microsoft/vscode/issues/289812#issuecomment-3787979414

https://code.visualstudio.com/docs/languages/identifiers#_known-language-identifiers

Languages were ripped directly from VSCode's Language Picker

image

https://github.com/microsoft/vscode/blob/d04cec02c65f5da9aaaab00206253f782c1c66f1/src/vs/workbench/browser/parts/editor/editorStatus.ts#L1172

```regex
^(?!.+"id"|.+"label").+\n?|\t+

"id": "(.+)",\n"label": "(.+)",
$2 | `$1`
```

### DIFF
--- a/docs/languages/identifiers.md
+++ b/docs/languages/identifiers.md
@@ -58,71 +58,76 @@ The following table lists known language identifiers:
 
 Language | Identifier
 -------- | ----------
-ABAP | `abap`
-Windows Bat | `bat`
+Agent | `chatagent`
+Batch | `bat`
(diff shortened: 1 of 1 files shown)


RedCMD:
or do you only want to add to the list?
(not remove any old languages/non-built-in languages)

ntrogh:
@RedCMD thanks for updating and yes, let's remove non-built-in languages.

PR #9300, opened 2026-01-30
fix Chromium link in Settings Sync docs

reillylm:
In the Settings Sync docs, we link to some Linux Desktop Environment detection logic from the Chromium codebase. However, the link isn't associated with a particular commit, and the Chromium file has since changed and the highlighted line numbers point elsewhere now. This changes the link to a commit-specific link and points the highlighting where it should go.

We added the link in #6431, which is dated Jun 26, 2023. Here's the [version of the Chromium file from that time](https://source.chromium.org/chromium/chromium/src/+/main:base/nix/xdg_util.cc;l=146-169;bpv=1;bpt=0;drc=153865a15b33852de58d6356625be98e6266202c).


ntrogh (approved):
@reillylm Thanks for the update!

PR #9318, opened 2026-02-04
Update blog writing instructions file

hawkticehurst:
Update `blog-writing.instructions.md` to include a more thorough description and guidance of current YAML frontmatter schema and how/when to use them in blog posts.

_*Note: Will change this again once blog pages redesign is done._


ntrogh (approved):
Thanks @hawkticehurst !!

PR #9328, opened 2026-02-05
Update section "Add elements to AI chat"

jruales:


docs/debugtest/integrated-browser.md
```diff
@@ -41,9 +41,8 @@ Move the browser to its own floating window by right-clicking the editor tab and
 
 Select elements from a web page to add them as context to your chat prompt. This is useful for getting help with specific HTML elements, CSS styles, or debugging UI issues.
 
-1. Enable element selection with the `setting(chat.sendElementsToChat.enabled)` setting.
```
jruales, line 44: Removing this instruction because that setting is enabled by default

docs/debugtest/integrated-browser.md
```diff
@@ -41,9 +41,8 @@ Move the browser to its own floating window by right-clicking the editor tab and
...
 1. Open the integrated browser and navigate to your web app.
-1. Select the **Start** button in the browser toolbar to enter selection mode.
```
jruales, line 46: The "Start" button is only there in the old Simple Browser and Live Preview experiences and not in the new Integrated Browser.

docs/debugtest/integrated-browser.md
jruales: I wonder if we should we mention "`localhost` URLs from Terminal, Chat, etc." here instead of just "`localhost` URLs"?

PR #9329, opened 2026-02-06
Change setting name for advanced settings display

mdeyanov83:
Updated the setting name for always showing advanced settings in the Settings editor.


ntrogh (approved):
@mdeyanov83 Appreciate you fixing this!

PR #9444, opened 2026-02-20
Configuration structure "servers" and "inputs" types were switched?

anentropic:
`servers` "contains the list of MCP servers"

while `inputs` is "optional placeholders for sensitive information like API keys"

So the former is clearly a list and I'm guessing the latter is an object of key/value pairs and the types were just switched by mistake


ntrogh:
@anentropic Apologies for the confusion. The `servers` contains an object of key/value pairs, where the key is the server name. The `inputs` is an array of input variable definitions.
You can see an example in the article: https://code.visualstudio.com/docs/copilot/customization/mcp-servers#_input-variables-for-sensitive-data

I'll clarify this better in the article.

PR #9445, opened 2026-02-21
Fix the language client cleanup in the language server guide

remcohaszing:
The method `client.stop()` stops the language server. The method `client.dispose()` does a bit more: It sets an internal state to disposing, then stops the client, then sets this internal state to disposed. If one of these states is set, the client refuses to start again.

This matters if the client is exposed via other means. For example, it’s common to implement a restart command. If a user would restart a language server while the extension is being deactivated, that would be a race condition.

This change also sets the client to `undefined` after deactivation.

Also `client.start()` is awaited, since it’s asynchronous. I don’t know if this actually matters.

### DIFF
--- a/api/language-extensions/language-server-extension-guide.md
+++ b/api/language-extensions/language-server-extension-guide.md
@@ -146,9 +146,9 @@ import {
   TransportKind
 } from 'vscode-languageclient/node';
 
-let client: LanguageClient;
+let client: LanguageClient | undefined;
 
-export function activate(context: ExtensionContext) {
+export async function activate(context: ExtensionContext) {
   // The server is implemented in node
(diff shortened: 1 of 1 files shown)


remcohaszing:
I suppose you could also push the client to the `context.subscriptions` instead. I don’t know what difference that makes.

remcohaszing:
@ntrogh Do you happen to know if it makes a difference if you push the client to `context.subscriptions` or if you dispose it in `deactivate`? I’m not sure for example whether either of them is awaited if a promise is returned. I would love to know this.

ntrogh:
VS Code disposes subscriptions synchronously, while `LanguageClient.dispose()` returns `Promise<void>` and needs async shutdown. Pushing the `LanguageClient` to subscriptions would trigger a non-awaited disposal, potentially leaving the language server process dangling.

PR #9450, opened 2026-02-23
fix: correct typo `user-invokable` to `user-invocable`

ilfroloff:


ntrogh (approved):
@ilfroloff Thanks! Changes are now in Stable (1.110)

PR #9633, opened 2026-04-08
Improve Aspire Dashboard instructions and add screenshot

JamesNK:
Updated the OTLP endpoint from gRPC to HTTP and modified the related configuration settings. This simplifies configuration.

Add the Aspire Dashboard trace screenshot image reference to the monitoring-agents guide. The image file was already present but wasn't referenced in the markdown.


ntrogh (approved):
Thanks @JamesNK

PR #9645, opened 2026-04-13
path used to store trust

theking2:


ntrogh:
@theking2 Closing this PR as the location of the workspace trust information is an implementation detail. It's not intended to be used directly as it might change over time. Appreciate your contribution.

PR #9668, opened 2026-04-19
Fix table of contents link in release notes

mon-jai:
The correct hash should be `_github-copilot-is-now-builtin`: https://code.visualstudio.com/updates/v1_116#_github-copilot-is-now-builtin


ntrogh (approved):

ntrogh:
Thanks @mon-jai

PR #9675, opened 2026-04-21
Fix: agent hooks usage example for blocking dangerous commands not working

painfulexistence:
The block-dangerous-commands usage example doesn't work because the tool_name check in the script doesn't match the actual tool name, which should be "run_in_terminal".

VSCode version: 1.116.0
OS: Darwin arm64 24.6.0

Steps to reproduce:
1. Copy the block-dangerous-commands example code from the Usage Scenarios section
2. Ask the agent to run `rm -rf ${some_test_file}`
3. Agent hook doesn't block the command

### DIFF
--- a/docs/copilot/customization/hooks.md
+++ b/docs/copilot/customization/hooks.md
@@ -599,12 +599,12 @@ INPUT=$(cat)
 TOOL_NAME=$(echo "$INPUT" | jq -r '.tool_name')
 TOOL_INPUT=$(echo "$INPUT" | jq -r '.tool_input')
 
-if [ "$TOOL_NAME" = "runTerminalCommand" ]; then
+if [ "$TOOL_NAME" = "run_in_terminal" ]; then
   COMMAND=$(echo "$TOOL_INPUT" | jq -r '.command // empty')
 
   if echo "$COMMAND" | grep -qE '(rm\s+-rf|DROP\s+TABLE|DELETE\s+FROM)'; then
     echo '{"hookSpecificOutput":{"permissionDecision":"deny","permissionDecisionReason":"Destructive command blocked by security policy"}}'
-    exit 0
+    exit 2
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


docs/copilot/customization/hooks.md
```diff
@@ -599,12 +599,12 @@ INPUT=$(cat)
...
 TOOL_INPUT=$(echo "$INPUT" | jq -r '.tool_input')
 
-if [ "$TOOL_NAME" = "runTerminalCommand" ]; then
+if [ "$TOOL_NAME" = "run_in_terminal" ]; then
   COMMAND=$(echo "$TOOL_INPUT" | jq -r '.command // empty')
 
   if echo "$COMMAND" | grep -qE '(rm\s+-rf|DROP\s+TABLE|DELETE\s+FROM)'; then
     echo '{"hookSpecificOutput":{"permissionDecision":"deny","permissionDecisionReason":"Destructive command blocked by security policy"}}'
-    exit 0
+    exit 2
```
ntrogh, line 607: The user would get a generic "unexpected error" with exit code 2. With the exit code 0, the command is denied and the provided reason is shown. Can you revert back to exit code 0?

ntrogh (changes requested):
@painfulexistence Thanks for the PR. Can you revert the exit code?

docs/copilot/customization/hooks.md
painfulexistence: Hi @ntrogh, thanks for the review!

I did try exit code 0 and it didn't work. The command did not be blocked when exit code is either 0 or 1.
(I've double checked one minutes ago)

ntrogh: @painfulexistence Ok, let me check if this is a bug. Thanks for looking into this!

ntrogh: @painfulexistence I can't repro with exit code 0. The hook blocks the command from executing. I do notice that the regex in the sample only blocks `rm -rf` and not plan `rm`. Can you check in the agent debug logs (**...** > **Show Agent Debug Logs**) which command is being run by the LLM?

painfulexistence: Actually I tested it with `git add` command.
FYI, I ran in autopilot mode. I can still reproduce this issue with exit code 0. Could you try it with autopilot mode?

ntrogh: Hm, this might actually be a side-effect of Autopilot, which auto-approves all tool calls. 
I'll create an issue in the vscode repo.

Can you confirm the hook works with exit code 0 when you use default permissions?

painfulexistence: Okay, I tried Default Approvals mode, but exit code 0 still did not block specified commands. Only exit code 2 works for me.

ntrogh: @painfulexistence This seems to be a product issue. Can you log an issue in the microsoft/vscode repo and provide the repro steps for the scenario that doesn't work with exit code 2?

PR #9691, opened 2026-04-23
docs: updated old screenshot with PHP extensions

jakubmisek:
Refresh the old 2021 screenshot with the updated list of extensions, their statuses, features, and UI changes in VS Code.

The new screenshot keeps the same dimensions and the same VS Code Theme. It is unmodified and as it is from VS Code to the current date.


ntrogh (approved):
@jakubmisek Thanks!

PR #9709, opened 2026-04-25
Update accessibility documentation for zoom settings

vermiceli:
Clarify the explanation of persisted zoom levels and added details about using decimal values for finer adjustments using the `window.zoomLevel` setting markdown description found in [desktop.contribution.ts in vscode repo](https://github.com/microsoft/vscode/blob/5886d8601a2fce479d0330bd6e18616662821dac/src/vs/workbench/electron-browser/desktop.contribution.ts#L207).

The original description wasn't clear to me.  I wasn't sure what an increment/decrement meant nor it wasn't explained that decimals could be used.  This clarifies that increment/decrement means increase or decrease by 1 and shows the setting explicitly allows decimals for finer grain zoom granularity.

### DIFF
--- a/docs/configure/accessibility/accessibility.md
+++ b/docs/configure/accessibility/accessibility.md
@@ -21,7 +21,7 @@ You can adjust the zoom level in VS Code with the **View** > **Appearance** > **
 
 ### Persisted zoom level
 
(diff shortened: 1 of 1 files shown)


ntrogh (approved):
Thanks @vermiceli . I made a few minor edits.

alexr00 (approved):

PR #9719, opened 2026-04-28
Document ExecutionPolicy requirements for Terminal Shell integration

jonahbeckford:
Without the ExecutionPolicy enabled (which I ran into after I reinstalled Windows), you get the following in the Chat Windows:

```
Enable shell integration to improve command detection
```

It was difficult to debug what is going wrong. The manual integration steps gave the clue because the hidden PowerShell windows opened by the Chat Window would show in red:

```
. : File C:\Users\USERNAME\OneDrive\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1 cannot be loaded because running scripts is disabled on this system. For 
more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.
```


ntrogh (approved):
Thanks!

chrmarti (approved):

PR #9721, opened 2026-04-29
fix: correct setting name for DevContainer lockfile in release notes

demershov:
Changed the setting name from `dev.containers.lockfile` to `dev.containers.experimentalLockfile` in the release notes to accurately reflect the current configuration option.


ntrogh:
@demershov The setting is correct in the latest release.

demershov:
@ntrogh Thanks.

PR #9750, opened 2026-05-07
Change logic for | separators in template-release-note-*.md

JohnVeness:
Changed so that the | separators are only shown if there are more things appearing on the line. I.e. make it so there won't be a lone hanging | at the end of the line.


ntrogh (approved):
Thanks @JohnVeness

PR #9751, opened 2026-05-07
Update template-release-note-insiders.md

JohnVeness:
Fix mistake I introduced in b049bb2 (PR #9750) regarding the order of | and %. Oops!

(template-release-note-endgame.md is OK).

Paging @ntrogh


ntrogh (approved):
@JohnVeness Thanks. I should've noticed this myself :)

ntrogh:
FYI, these templates are only used as part of endgame prep, not with every build. I'll fix the published release notes retroactively.

PR #9347, opened 2026-02-09
Add dev server functionality based on docsify

ntrogh:
Add a local documentation preview workflow using [Docsify](https://docsify.js.org/). The changes make it easier for contributors to preview and navigate the documentation site locally.

The website does not exactly match the production website in terms of style, but enables previewing content changes and verifying cross-linking.

To preview docs locally, run the following command in the terminal:

```bash
npm install
npm run serve
```

PR #9785, opened 2026-05-17
Fix broken in-page anchor links across API reference docs

### LINKED ISSUE #9760: Anchor links on several pages are broken due to included underscore prefix
_Edit: Updated the title to reflect that the issue affects various pages._

Three in-page links related to views are currently broken on https://code.visualstudio.com/api/references/contribution-points because they are prefixed with "_" (identified by running `$$("a[href^='#_']")` in the console):
- section [contributes.views](https://code.visualstudio.com/api/references/contribution-points#contributes.views):
    > - [Custom view containers](https://code.visualstudio.com/api/references/contribution-points#_contributes.viewsContainers) contributed by Extensions.
- section [contributes.viewsContainers](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsContainers):
  > Contribute a view container into which [Custom views](https://code.visualstudio.com/api/references/contribution-points#_contributes.views) can be contributed.
- section [contributes.viewsWelcome](https://code.visualstudio.com/api/references/contribution-points#contributes.viewsWelcome):
  > Contribute welcome content to [Custom views](https://code.visualstudio.com/api/references/contribution-points#_contributes.views). 

All links are correct in the Markdown source and haven't been changed in a while, which suggests the the build step as source of the issue.

Dotify71:
Hey! Found a few broken in-page anchor links while browsing the reference pages. 

It seems the build pipeline prefixes purely relative hash links (like `#extension-manifest`) with an underscore (resulting in `#_extension-manifest`), whereas the actual generated heading IDs do not have that prefix (e.g. `id="extension-manifest"`). This causes those links to break on the live site.

Adding the absolute page path prefix (like `/api/get-started/extension-anatomy#extension-manifest`) bypasses this issue entirely. I've updated the relative anchors in:
- `api/extension-capabilities/extending-workbench.md`
- `api/get-started/extension-anatomy.md`
- `api/references/contribution-points.md`
- `api/references/when-clause-contexts.md`

Closes #9760

### DIFF
--- a/api/extension-capabilities/extending-workbench.md
+++ b/api/extension-capabilities/extending-workbench.md
@@ -22,10 +22,10 @@ VS Code provides various APIs that allow you to add your own components to the W
 
 ![workbench-contribution](images/extending-workbench/workbench-contribution.png)
 
-- Activity Bar: The [Azure App Service extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice) adds a [View Container](#views-container)
-- Side Bar: The built-in [NPM extension](https://github.com/microsoft/vscode/tree/main/extensions/npm) adds a [Tree View](#tree-view) to the Explorer View
-- Editor Group: The built-in [Markdown extension](https://github.com/microsoft/vscode/tree/main/extensions/markdown-language-features) adds a [Webview](#webview) next to other editors in the Editor Group
-- Status Bar: The [VSCodeVim extension](https://marketplace.visualstudio.com/items?itemName=vscodevim.vim) adds a [Status Bar Item](#status-bar-item) in the Status Bar
+- Activity Bar: The [Azure App Service extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice) adds a [View Container](/api/extension-capabilities/extending-workbench#views-container)
+- Side Bar: The built-in [NPM extension](https://github.com/microsoft/vscode/tree/main/extensions/npm) adds a [Tree View](/api/extension-capabilities/extending-workbench#tree-view) to the Explorer View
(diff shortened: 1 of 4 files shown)


mootari (changes requested):
I'm afraid this is not a suitable fix:
- You're only updating the files that I already listed in the issue, but those were meant to be examples. This bug affects potentially hundreds of files ([see search](https://github.com/search?q=repo%3Amicrosoft%2Fvscode-docs+%22%5D%28%23%22+language%3AMarkdown+path%3Adocs&type=code)).
- Even if you were to update all those files, it wouldn't prevent future links from breaking.

This needs to be addressed at the build stage, preferably without breaking existing in-page links that point to the correct anchors.

Dotify71:
Hey @mootari! Thanks for the review.

I totally agree that this should be fixed at the root level in the build pipeline rather than patching individual markdown files. However, looking at the CI configuration, it seems the build pipeline and markdown processing are handled in the microsoft/vscode-website repository, which is private.

Since I don't have access to the vscode-website repository to fix the parser logic, I'm unable to address this at the build stage. Should I go ahead and close this PR so a core team member with access to the build pipeline can take over the fix or do something else? Let me know how you'd like to proceed!

mootari:
> Should I go ahead and close this PR so a core team member with access to the build pipeline can take over the fix

Yeah, that might be best I think, otherwise it will just crop up again and again. Ideally the Docs team would include a CI step to check for broken links. (cc @ntrogh for visibility)

(Also, TIL that the Docsify dev server only provides [an approximation](https://github.com/microsoft/vscode-docs/pull/9347).)

Dotify71:
Closing as discussed since the root cause needs to be fixed in the private build pipeline. Thanks for the review @mootari!

PR #9786, opened 2026-05-17
docs: clarify terminal line wrapping for multiline problem matchers

### LINKED ISSUE #6851: Wrapped Lines Break Custom Problem Matcher
Does this issue occur when all extensions are disabled?: Yes

- VS Code Version: 1.84.2
- OS Version: Windows 10

I saw this issue crop up on your GitHub for other problem matchers, but tbh it was super unclear if this was supposed to be fixed or not. microsoft/vscode#32042 makes it sound fixed, but then we have microsoft/vscode#152634 which sounds very similar.

Steps to Reproduce:

1. Create a dummy script the outputs long lines (like replicating the output from running tests with `dotnet test`)
2. Associate that script to a task.
3. Create a problem matcher that attempts to get the `error message`, `fully qualified test name`, and `line number`.
4. Problem matcher consistently fails when the terminal is not scaled large enough for the fully qualified class test (method) name to fit on one line, but seems to work "most" of the time when the terminal large enough and does not wrap the lines.

### Sample Task
```json
{
    "windows": {
        "options": {
            "shell": {
                "executable": "pwsh.exe",
                "args": [
                    "-NoProfile",
                    "-ExecutionPolicy",
                    "Bypass",
                    "-Command"
                ]
            }
        }
    },
    "label": "Test MultiLine",
    "type": "shell",
    "command": "${workspaceFolder}/TestMultiLine.ps1",
    "group": {
        "kind": "test",
        "isDefault": true
    },
    "problemMatcher": {
        "fileLocation": ["absolute"],
        "severity": "error",
        "pattern": [
            {
                "regexp": "^\\s+(Assert.*)$",
                "message": 1
            },
            {
                "regexp": "^\\s+Stack Trace:$"
            },
            {
                "regexp": "^\\s+at\\s+(.*)\\s+in\\s+(.*):line\\s+(\\d+)\\s*$",
                "file": 2,
                "line": 3
            }
        ]
    }
}
```

### TestMultiLine.ps1
```powershell
Write-Output '  Failed MethodNameThatSeemsToGoOnForever [6 ms]'
Write-Output '  Error Message:'
Write-Output '   Assert.ThrowsException failed. No exception thrown. ThisExceptionHappensSometimes() exception was expected.'
Write-Output '  Stack Trace:'
Write-Output '     at Hello.World.This.IsAVeryLongClassNameWithAnEquallyLong.MethodNameThatSeemsToGoOnForever() in C:\Also\The Path To The File Is Also\Very Large\Which\Means.This\Will\LikelyWrapForMostPeople.cs:line 110'
```

### Footnote
I spent like 6 hours trying to understand what the heck was going on, so some clarity on this would be wonderful.

As a **side note** the documentation on multi-line problem matcher patterns is super vague, from I've been able to discern mostly by trial, error, and searching VS Code GitHub. Is that when you have a multi-line pattern, you seemingly need to match every line in between, even if you do not care about those lines. 

As far as I can tell nothing mentions that explicitly and the example in the docs [Defining a multiline problem matcher](https://code.visualstudio.com/docs/editor/tasks#_defining-a-multiline-problem-matcher) "works" because the multi-lines they are matching are immediately after the first line they are matching. The docs do say **This way you define a pattern per each line you want to match**, but that's also not super obvious because I **DON'T** want to match them, I don't care about those lines. If someone could also confirm this that would be nice or I can log a documentation issue to give some better clarity on this so other people don't spend their entire night playing with regexes that **work** and consistently **don't work** in VS Code problem matchers....

Dotify71:
Fixes #6851. Adds a note about terminal line wrapping breaking multi-line problem matchers, which was a point of confusion for users trying to write custom problem matchers that match across lines wrapped by the terminal window.

### DIFF
--- a/docs/debugtest/tasks.md
+++ b/docs/debugtest/tasks.md
@@ -727,7 +727,10 @@ test.js
 Our problem matcher is line-based so we need to capture the file name (test.js) with a different regular expression than the actual problem location and message (1:0   error  Missing "use strict" statement).
 
 To do this, use an array of problem patterns for the `pattern` property. This way you define a pattern per each line you want to match.
+
 > **Note:** In a multi-line problem matcher, each line in the output must be matched sequentially by the pattern array. Intermediate lines cannot be skipped, even if they are not needed for capturing values.
+
+> **Note:** Terminal line wrapping can break multi-line problem matchers. If a line of output is too long and the terminal wraps it into two lines, VS Code processes it as two separate lines. Your problem matcher will fail unless your pattern array accounts for the extra wrapped line. To avoid this, ensure your terminal is wide enough or configure your build tool to disable line wrapping if possible.
 
 The following problem pattern matches the output from ESLint in stylish mode - but still has one small issue that we need to resolve next. The code below has a first regular expression to capture the file name and the second to capture the line, column, severity, message, and error code:
 
--- a/docs/debugtest/tasks.md
+++ b/docs/debugtest/tasks.md
@@ -379,7 +379,7 @@ You can specify a task's run behaviors using the `runOptions` property:
   * `terminateNewest` - Terminate the newest running instance.
   * `terminateOldest` - Terminate the oldest running instance.
   * `warn` - Don't start a new instance (show warning).
-  
+
 ## Customizing auto-detected tasks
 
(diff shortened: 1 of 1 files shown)


marabyryk-sys (commented):
G

ntrogh (approved):
Updated the PR to include a clarifying example, as requested by the OP.

TylerLeonhardt (approved):

PR #9826, opened 2026-05-25
Remove questoion mark-it's not a question

MarcelloCuoghi:


MarcelloCuoghi (commented):
Remove the question mark because the phrase is not a question.

ntrogh (approved):
Thanks!

PR #9835, opened 2026-05-27
fix(tasks): update tasks appendix schema

### LINKED ISSUE #243556: Inconsistencies task.json vs. task.json schema
Type: <b>Bug</b>

When trying to validate my task.json against the task.json schema published in the [Appendix](https://code.visualstudio.com/docs/reference/tasks-appendix) I noticed some inconsistencies.

1. `beginsPattern`/ `endsPattern`
    The schema defines them this way in the `BackgroundMatcher` interface

    ```json
    beginsPattern?: string;
    endsPattern?: string;
    ```

    But in VS Code I can define them this way:
    ```json
    "problemMatcher": {
      "base": "$tsc-watch",
      "background": {
        "activeOnStart": true,
        "beginsPattern": {
          "regexp": "(.*?)"
        },
        "endsPattern": {
          "regexp": "bundle generation complete"
        }
      }
    },
    ```

    So while the schema restricts the value to be string, it also accepts `regexp` in reality. Also IntelliSense provides this as a valid entry.

2. Missing property "type"

    The `TaskDescription` interface defines the `type` like this

    ```json
    type: 'shell' | 'process';
    ```

    It is therefore defined as a mandatory field. But in reality it is not. Otherwise it would not be possible to create a compound task as described in [Compound Tasks](https://code.visualstudio.com/docs/editor/tasks#_compound-tasks)

3. Missing property "command"

    The `TaskDescription` interface defines the `command` like this

    ```json
    command: string;
    ```

    It is therefore defined as a mandatory field. But in reality it is not. Otherwise it would not be possible to create a compound task as described in [Compound Tasks](https://code.visualstudio.com/docs/editor/tasks#_compound-tasks)

VS Code version: Code 1.98.1 (2fc07b811f760549dab9be9d2bedd06c51dfcb9a, 2025-03-10T15:38:08.854Z)
OS version: Windows_NT x64 10.0.22631
Modes:
Remote OS version: Linux x64 5.15.167.4-microsoft-standard-WSL2

lizuju:
## Summary
- mark task `type` and `command` as optional in the tasks appendix schema snippet
- document object-form background `beginsPattern`/`endsPattern` via a `WatchingPattern` interface

Fixes microsoft/vscode#243556

## Verification
- Checked the updated appendix for optional task fields and object-form background patterns with a local Node assertion.
- Compared against `microsoft/vscode` source: `problemMatcher.ts` defines `beginsPattern?: string | IWatchingPattern` and `endsPattern?: string | IWatchingPattern`; `jsonSchema_v2.ts` requires only `label` for the generic task description and adds `command`/`type` only for process tasks.
- Ran `git diff --check`.

### DIFF
--- a/docs/reference/tasks-appendix.md
+++ b/docs/reference/tasks-appendix.md
@@ -44,13 +44,13 @@ interface BaseTaskConfiguration {
      * The type of a custom task. Tasks of type "shell" are executed
      * inside a shell (e.g. bash, cmd, powershell, ...)
      */
-    type: "shell" | "process";
+    type?: "shell" | "process";
 
     /**
      * The command to be executed. Can be an external program or a shell
      * command.
      */
-    command: string;
+    command?: string;
 
     /**
      * Specifies whether a global command is a background task.
@@ -136,13 +136,13 @@ interface TaskDescription {
      * The type of a custom task. Tasks of type "shell" are executed
      * inside a shell (e.g. bash, cmd, powershell, ...)
      */
-    type: "shell" | "process";
+    type?: "shell" | "process";
 
     /**
      * The command to execute. If the type is "shell" it should be the full
      * command line including any additional arguments passed to the command.
      */
-    command: string;
+    command?: string;
 
     /**
      * Whether the executed command is kept alive and runs in the background.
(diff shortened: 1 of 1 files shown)


lizuju:

ntrogh (approved):
Thanks

PR #9905, opened 2026-06-09
Document search exclusion settings for agent text search and grep

### LINKED ISSUE #9898: Recommend strict `files.exclude` and `search.exclude` to minimize noise in `#textSearch` results and reduce token consumption
Since Copilot has transitioned to usage-based billing, it is recommended to document best practices for controlling token consumption somewhere.

To prevent the built-in tool `grep_search`, or `#textSearch`, from returning irrelevant results and filling the context window with noise, it is advisable to strictly configure `files.exclude` and `search.exclude`. This is recommended because `grep_search` respects these settings by default, which can help keep junk out of the search results. 

Even if the agent does not read a file in the search results, if the file is in a format like `jsonl`, the search results alone will consume many tokens.

---

See:

- https://github.com/microsoft/vscode/blob/fdabaa87e945796e7abb30e92ad3d31c94288436/extensions/copilot/package.json#L395

>  "Whether to include files that would normally be ignored according to .gitignore, other ignore files and `files.exclude` and `search.exclude` settings. Warning: using this may cause the search to be slower. Only set it when you want to search in ignored folders like node_modules or build outputs."

1chooo:
## Summary
- Add guidance on how `files.exclude`, `search.exclude`, and `.gitignore` affect agent text search and grep
- Explain that search match snippets consume agent context, not just opened files
- Include example exclusion configs for Node.js, Python, and AI/data projects
- Cross-link from optimize-usage and best-practices guides

Fixes #9898

## Test plan
- [ ] Verify links resolve correctly on the docs site
- [ ] Review wording with docs team for accuracy of text search/grep exclusion behavior

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/agents/best-practices.md
+++ b/docs/agents/best-practices.md
@@ -33,6 +33,7 @@ Tips for effective project configuration:
 * **Keep instruction files concise.** They load on every chat interaction. Focus on information the AI can't infer from code, such as non-default conventions, architectural decisions, or environment setup.
 * **Scope instructions with `applyTo` patterns.** Enter `/instructions` to create language-specific or folder-specific instruction files instead of putting everything in one file.
 * **Limit enabled tools.** Fewer active tools means faster, more relevant responses. Enable tools only when the task needs them.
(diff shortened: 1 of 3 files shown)


ntrogh:
@mjbvz Can you verify if these updates to workspace context are correct? Thanks!

ntrogh (approved):
Thanks!

Yoyokrazy (approved):

1chooo:
Thanks @ntrogh

PR #9945, opened 2026-06-17
Fix Linux policy JSON examples

### LINKED ISSUE #9864: Enterprise policies: Linux section shows XML examples instead of JSON
The "Linux JSON policies" section in `docs/enterprise/policies.md` correctly states the format is JSON and the file path is `/etc/vscode/policy.json`, but the **Step 2 examples** use macOS plist XML syntax (`<key>`, `<string>`, `<true/>`).

This causes parse failures because VS Code's `FilePolicyService` uses `JSON.parse()` to read the file.

**Current (incorrect):**
```xml
<key>AllowedExtensions</key>
<string>{"microsoft": true, "github": true}</string>

<key>EnableTelemetry</key>
<false/>
```

**Expected (correct JSON):**
```json
{
  "AllowedExtensions": {"microsoft": true, "github": true},
  "UpdateMode": "start",
  "EnableFeedback": true,
  "EnableTelemetry": false
}
```

Source confirmation: [`FilePolicyService`](https://github.com/microsoft/vscode/blob/main/src/vs/platform/policy/common/filePolicyService.ts) uses `JSON.parse`, and the [build fixture](https://github.com/microsoft/vscode/blob/main/build/lib/test/fixtures/policies/linux/policy.json) is plain JSON.

NONAN23x:
Fixes #9864

## What changed

Updated the Linux enterprise policy examples to use JSON syntax for /etc/vscode/policy.json instead of plist XML syntax.

## Why

The Linux policy file is parsed as JSON, so the previous examples were misleading.

## Validation

- Verified the JSON example parses successfully.
- Ran git diff --check.

### DIFF
--- a/docs/enterprise/policies.md
+++ b/docs/enterprise/policies.md
@@ -276,40 +276,24 @@ Starting from VS Code version 1.106, each release ships with a sample `.policy.j
 
 1. Edit the policy values according to your requirements:
 
-    **String policies** - policies that accept text values or JSON strings:
+    Example policy file:
 
-    ```xml
-    <!-- Example: Allow extensions from specific publishers -->
-    <key>AllowedExtensions</key>
-    <string>{"microsoft": true, "github": true}</string>
-
-    <!-- Example: Set update mode to a specific value -->
-    <key>UpdateMode</key>
-    <string>start</string>
+    ```json
+    {
+      "AllowedExtensions": {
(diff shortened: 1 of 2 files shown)


NONAN23x:

ntrogh:
@NONAN23x Closing this PR as it was already implemented by another PR.

PR #9984, opened 2026-06-26
rm "insider" because since v1.122 custom endpoints are in stable

turbotimon:
remove "insider" because since v1.122 custom endpoints are in stable

see: https://code.visualstudio.com/updates/v1_122#_custom-endpoint-provider-in-stable


ntrogh (approved):
@turbotimon Thanks

PR #9985, opened 2026-06-26
Fix YouTube embed link in agents guide

softchris:
Updated the YouTube embed link format for the video on using tools with agents.


softchris:
@ntrogh @theReynald vs code breaks in production, this will fix the url as it's now correctly formed, please review and merge and trigger a build when convenient

ntrogh (approved):

PR #9994, opened 2026-06-28
docs: add search performance troubleshooting for excessive rg processes

### LINKED ISSUE #9950: Add troubleshooting steps for the spawning of excessive rg processes
Add troubleshooting steps for the spawning of excessive rg processes.

Users still regularly report the an excessive number of `rg` (ripgrep) processes are spawned. Looking closely at these reports, almost all of them show `rg` being executed with the `--no-ignore` flag. When run with this option, `rg` searches through directories like `node_modules`, which drastically increases the execution time. I suspect this happens because users do not understand the meaning of the `Use Exclude Setting and Ignore Files` button in the VS Code Search UI. With the toggle off, `rg` is executed with the `--no-ignore` flag.

Image

Troubleshooting steps should advise users encountering this issue to:

1. Check the toggle button: Ensure that the "Use Exclude Settings and Ignore Files" button is enabled in the Search UI.
2. Verify your ignore settings: Confirm that `search.useIgnoreFiles` is set to `true` in your settings.
3. Disable symbolic link following: If the issue persists, setting `search.followSymlinks` to `false` may resolve it.
4. Set in `settings.json`:
~~~
{
  "search.exclude": {
    "**/node_modules/**": true,
    ".claude/worktrees/**": true
  }
}
~~~

Context:

- https://github.com/microsoft/vscode/issues/193927
- https://github.com/microsoft/vscode/issues/244376
- https://github.com/microsoft/vscode/issues/300132

- https://github.com/microsoft/vscode/issues/305544

1chooo:
## Summary

- Adds a new `### Search performance` subsection to the `## Search across files` section in `docs/editing/codebasics.md`, covering the four troubleshooting steps from issue #9950
- Explains that disabling the **Use Exclude Settings and Ignore Files** toggle causes `rg` to run with `--no-ignore`, scanning `node_modules` and other large directories
- Adds an explicit `search.exclude` JSON example and a `search.followSymlinks` tip for worktree/symlink-heavy workspaces
- Adds a Common question entry that links to the new section for FAQ discoverability

Closes microsoft/vscode-docs#9950

## Test plan

- [ ] Verify the new `### Search performance` section renders correctly on the docs site
- [ ] Confirm all `setting(...)` macros resolve properly
- [ ] Check the `#search-performance` anchor link in the Common questions entry works

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/editing/codebasics.md
+++ b/docs/editing/codebasics.md
@@ -224,6 +224,35 @@ In the **files to include** box, you can use the **Search only in changed files*
 > [!TIP]
 > From the Explorer, you can right-click on a folder and select **Find in Folder** to search inside a folder only.
 
+### Search performance
+
+If VS Code search is slow or you notice many background `rg` processes in your system's process monitor, the most common cause is misconfigured exclude settings that allow search to scan large directories such as `node_modules`.
+
+Use the following steps to resolve the issue:
+
+1. **Check the "Use Exclude Settings and Ignore Files" toggle**: In the Search view, ensure the **Use Exclude Settings and Ignore Files** button in the **files to exclude** box is enabled (the default). When this toggle is disabled, VS Code searches all files without respecting `.gitignore` or your `setting(search.exclude)` settings, which can dramatically slow down search in large workspaces.
+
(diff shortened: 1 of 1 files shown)


ntrogh:
Issue is already assigned.

1chooo:
Thanks for reviewing!!!

PR #10025, opened 2026-07-08
Replace improper feature names

jo-oikawa:
Replace the use of 'Multi-Chat' and 'Quick Chat' as proper feature names with generic descriptions of functionality, as we don't use these in our UI.


pierceboggan (approved):
LGTM

PR #10030, opened 2026-07-10
docs: fix OTLP tracing metadata

fhshaik:
Fixes the tracing article metadata to refer to OTLP instead of OLTP and clarifies that Foundry Toolkit hosts a local HTTP and gRPC server, matching the article content. Validation: git diff --check.


question429 (approved):

ntrogh (approved):
Thanks

PR #10037, opened 2026-07-14
doc-writer: Single-pane detail panel for the Agents window (Changes editor & docked tabs)

aeschli:
[#325498](https://github.com/microsoft/vscode/issues/325498)


aeschli:
Will not be enabled for stable -> move to 1.130

ntrogh:
Incorporated changes

PR #10041, opened 2026-07-14
Finalize release for version 1.129

kycutler:


release-notes/images/1_129/release-social-background.webp
rzhao271, line 1: @kycutler Is the version number supposed to be cut off here? It shows for 1.130 which should be on Insiders.

roblourens (approved):

release-notes/images/1_129/release-social-background.webp
kycutler: Not too sure. It's the same for previous releases; e.g. https://github.com/microsoft/vscode-docs/blob/main/release-notes/images/1_128/release-social-background.webp. So inclined to leave it as is since the 129 release-highlights.webp looks fine.

PR #10050, opened 2026-07-20
Revert "feat: add agent window keybindings"

rzhao271:
Reverts microsoft/vscode-docs#10032 because the contributed keybindings are all for macOS only.


build/keybindings/doc.keybindings.linux.json
```diff
@@ -284,6 +285,7 @@
 { "key": "enter",                 "command": "workbench.action.chat.steerWithMessage" },
 { "key": "alt+enter",             "command": "workbench.action.chat.steerWithMessage" },
 { "key": "ctrl+shift+up",         "command": "workbench.chat.action.focusLastFocused" },
+{ "key": "ctrl+shift+up",         "command": "workbench.chat.action.focusLastFocused" },
```
dmitrivMS, line 288: I don't have context, but how is this line different from the one above it?

rzhao271: It's not. I'm doing a clean revert though, so the diff does restore duplicated entries.

dmitrivMS (approved):

PR #10071, opened 2026-07-28
docs: fix pluralization of 'platforms' in README

denizguney:
Changed "platform" to the plural "platforms" since multiple operating systems (Linux, macOS, and Windows) are listed.


ntrogh:
Leaving the text as-is, also works grammatically. Thanks for contributing!

PR #10083, opened 2026-08-01
fix(docs): correct BYOK utility model default

ShiroKSH:
## Summary

- Correct the documented default utility model for BYOK main-agent models.
- Explain how to restore utility features when GitHub Copilot utility models are unavailable.

## Root cause

The setting defaults to GitHub Copilot, but the note incorrectly said that no utility model is used by default.

## Validation

- Checked the current VS Code setting registration and endpoint-resolution behavior.
- Ran whitespace checks and regenerated the documentation sidebars without extra changes.


ntrogh (approved):
Thanks @ShiroKSH

PR #10084, opened 2026-08-01
Replace "when the agent session ends" with detailed description it means agent execution stopping, not the chat session stopping

okko:


ntrogh (approved):
Thanks @okko

PR #10092, opened 2026-08-04
Document built-in dictation in VS Code

cwebster-99:
Built-in dictation replaces the extension-first experience on supported desktop platforms, so the voice documentation needs to reflect how speech input now works across VS Code. I opted for our documentation to reflect the built in support and we can link to the speech extension readme as a resource for the extension docs. 

This update:

* Reframes voice support around built-in dictation in chat, the Agents window, editors, and terminals.
* Documents the default on-device Nemotron model, first-use download, offline behavior, platform support, privacy, commands, and configuration.
* Keeps the VS Code Speech extension as the fallback recommendation for unsupported platforms.
* Adds the current dictation settings to the AI settings reference and corrects the terminal speech timeout default and scope.

Todo: 
- Add screenshots


meganrogge (dismissed):
Thank you!

ntrogh (approved):
Thanks for adding this @cwebster-99 !

PR #10104, opened 2026-08-06
Document Agent Plugins as an open standard

digitarald:
## Summary

- present Agent Plugins 1.0 as an open, cross-client standard, consistent with the Agent Skills documentation
- document the canonical manifest fields, fixed component locations, path variables, and client extension namespaces
- distinguish the portable v1 core from VS Code-specific agents, hooks, commands, and legacy plugin formats
- remove current Preview labels while preserving the historical Experimental label in the v1.110 release notes

## Validation

- `git diff --check`
- verified new Agent Plugins documentation links return HTTP 200
- cross-checked manifest fields against the canonical v1 JSON schema and VS Code parser
- cross-checked format detection and fixed component paths against `microsoft/vscode`

### DIFF
--- a/docs/agent-customization/agent-plugins.md
+++ b/docs/agent-customization/agent-plugins.md
@@ -16,9 +16,10 @@ Keywords:
 ---
 # Agent plugins in VS Code
 
-Agent plugins are prepackaged bundles of agent customizations that you can discover and install from plugin marketplaces in Visual Studio Code. Agent Plugins is an [open standard](https://agent-plugins.org/) for packaging [agent skills](/docs/agent-customization/agent-skills.md) and [MCP servers](/docs/agent-customization/mcp-servers.md) that works across multiple AI agents, including GitHub Copilot in VS Code, GitHub Copilot CLI, and the GitHub Copilot app.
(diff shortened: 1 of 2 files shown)


digitarald:
@ntrogh @connor4312 — could you please review the Agent Plugins 1.0 format and compatibility details here? In particular, I would appreciate a check of the portable-vs-client-specific boundary and the legacy format wording.

ntrogh:
@digitarald I've created an [add-on PR](https://github.com/microsoft/vscode-docs/pull/10105) with my feedback, as it involved some restructuring. Feel free to apply the changes you like.

Beyond the standards work, should we move the plugin manifest and configuration information towards the end of the article and move browsing and installing plugins to the top? What is the most common use case for users to come to this content: creating plugins or browsing/installing? WDYT?

digitarald:
Agreed on browsing/installing. But understanding that plugins are a spec and not "Claude Plugins" is a key clarification. Moving it further down LGTM .

ntrogh (approved):

ntrogh:
@digitarald LGTM. Merge whenever you're good

PR #10120, opened 2026-08-10
Fix two links that 404 on code.visualstudio.com

kutsibalci:
## What

Two links in published topics that return **404 on code.visualstudio.com**. `+2/-2`, two files.

| file | link | after |
|---|---|---|
| `docs/languages/rust.md:266` | `/docs/debugtest/debugging.md/#launch-configurations` | `/docs/debugtest/debugging.md#launch-configurations` |
| `docs/intelligentapps/finetune.md:399` | `/uwp/api/windows.applicationmodel.limitedaccessfeatures` | `https://learn.microsoft.com/uwp/api/...` |

## Verified against the live site, not the Markdown

Per the note at the top of CONTRIBUTING.md, I resolved every candidate against the production site rather than trusting my reading of the source. Both of these are `404` there today; the fixed forms are `200`.

That check is what removed most of the sweep. Of 577 distinct site-relative link targets, 69 had no matching file in the repository, but only 11 actually fail on the site — `/updates/v1_74.md` is served from `release-notes/`, `/api/references/vscode-api` is generated, `/assets/**` comes from a separate pipeline, and a dozen more resolve through `redirection.json` or a site-level redirect (`/docs/getstarted/keybindings.md` → `/docs/configure/keybindings`, `/docs/azure/docker.md` → `/docs/containers/overview`, and so on). Those all work and are untouched here.

## Why each is an anomaly rather than a convention

**The stray slash.** `.md/#anchor` occurs exactly **once** in the repository. The correct `.md#anchor` form occurs **840** times.

**The Microsoft Learn path.** `docs/intelligentapps/finetune.md` links to the `LimitedAccessFeatures` class with a site-relative `/uwp/...` path, which the site resolves against `code.visualstudio.com` and cannot find. Every other Microsoft Learn reference in this repository — **394** of them — is written as a full `https://learn.microsoft.com/...` URL. This is the only one that is not. The target itself is fine: `https://learn.microsoft.com/uwp/api/windows.applicationmodel.limitedaccessfeatures` returns `200`.

## Deliberately left alone

- `/issues` and eight `/Issues/Detail/NNNNN` links. They 404, but they are in `release-notes/v0_3_0.md` through `v0_8_0.md` — a dated record of releases from 2015, pointing at an issue tracker that no longer exists. Rewriting history seemed wrong; happy to include them if you would rather they were fixed.
- `/docs/copilot/getting-started.md`, which 404s but appears only in `.github/skills/`, not in published content.

## AI disclosure

AI-assisted (Claude Code). The tool ran the link sweep and the live verification and drafted this description. I checked the findings before opening: each surviving link was requested against code.visualstudio.com, the replacement URLs were requested too, and the two anomaly counts above were measured across the repository. I have reviewed and understood the change and take responsibility for it.

### DIFF
--- a/docs/intelligentapps/finetune.md
+++ b/docs/intelligentapps/finetune.md
@@ -396,7 +396,7 @@ If you want to modify the Azure subscription and resource group that were previo
 ## Inference with the Phi Silica LoRA adapter
 
 > [!IMPORTANT]
-> The Phi Silica APIs are part of a Limited Access Feature (see [LimitedAccessFeatures class](/uwp/api/windows.applicationmodel.limitedaccessfeatures)). For more information or to request an unlock token, please use the [LAF Access Token Request Form](https://go.microsoft.com/fwlink/?linkid=2271232&c1cid=04x409).
+> The Phi Silica APIs are part of a Limited Access Feature (see [LimitedAccessFeatures class](https://learn.microsoft.com/uwp/api/windows.applicationmodel.limitedaccessfeatures)). For more information or to request an unlock token, please use the [LAF Access Token Request Form](https://go.microsoft.com/fwlink/?linkid=2271232&c1cid=04x409).
 
 > [!NOTE]
 > Inference with the Phi Silica LoRA adapter is currently supported only on Copilot+ PCs with ARM processors.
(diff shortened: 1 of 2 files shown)


kutsibalci:

ntrogh (approved):
Thanks!

ntrogh (approved):

rzhao271 (approved):

PR #10124, opened 2026-08-10
enterprise: clarify managed settings policy precedence

digitarald:
## Summary

- correct managed-settings precedence to describe per-key resolution across native MDM, server, and file channels
- document that Copilot managed settings replace overlapping VS Code device policy values rather than merging with them
- add MCP allowlist, denylist, and managed-only policy mappings, version requirements, and a deployment example
- clarify what Policy Diagnostics reports and recommend a safe device-specific override path

## Why

The existing guidance said that one managed-settings channel wins outright. Current VS Code resolves managed settings per key. It also did not explain that an overlapping GitHub-delivered managed setting takes precedence over the corresponding VS Code ADMX or configuration-profile policy. This can make a device-level MCP allowlist appear to stop working.

## Validation

- checked the documentation diff with `git diff --check`
- verified policy names, managed-setting keys, schemas, precedence behavior, and minimum versions against the current VS Code source
- completed an independent technical and writing-guideline review

### DIFF
--- a/docs/enterprise/ai-settings.md
+++ b/docs/enterprise/ai-settings.md
@@ -33,7 +33,7 @@ All three channels use the same managed setting keys and values. For the list of
 > [!NOTE]
 > Precedence is enforced starting in VS Code version 1.128.
 
-When the same setting is available from more than one channel, VS Code uses a single authoritative channel rather than merging the channels. The channel with the highest precedence that provides any managed settings wins outright, and the other channels are ignored.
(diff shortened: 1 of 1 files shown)


ntrogh (approved):
@digitarald Thanks for the updates

joshspicer (approved):

shams833:
#10024

shams833:
#10162

PR #10127, opened 2026-08-11
Clean uninstall does not work, VS Code keeps my old data at ~/.vscode-shared, which need to be documented!

d41bf92:
Please update the instructions because this is very confusing and frustrating.


ntrogh (approved):

d41bf92:
a

```%USERPROFILE%.vscode-shared```
I'm not a Windows expert but it looks weird

dbaeumer (approved):

PR #5718, opened 2022-10-18
Adds relative link fix for several videos.

anthonyisensee:
I don't know exactly how your website build occurs but based on other /assets/ links present in the repository I believe this should fix the broken video links and point to the files that are actually there. May fix #5717.

Issue #5717, opened 2022-10-18
Links to Languages/JS videos broken


anthonyisensee:
Videos at https://code.visualstudio.com/docs/languages/javascript lead to broken links. Have tested on multiple browsers and networks to ensure it's not a me problem.

For example, both of the broken links are rendered to the page:

- /assets/docs/languages/docs/languages/javascript/intellisense.mp4
- /assets/docs/languages/docs/languages/javascript/snippets.mp4

Based on examining `.gif` files on the R page it appears they should lead to these links (boom, there the files are!):

- /assets/docs/languages/javascript/intellisense.mp4
- /assets/docs/languages/javascript/snippets.mp4

In other words, the extra "docs/languages/" must be removed from the links.

anthonyisensee:
https://raw.githubusercontent.com/microsoft/vscode-docs/main/docs/languages/javascript.md currently contains the following:

- /docs/languages/javascript/intellisense.mp4
- /docs/languages/javascript/snippets.mp4

anthonyisensee:
Based on the assumption that something is adding `/assets/docs/languages` when it only sees `/docs/languages/javascript/file.mp4` I've created [a pull request](https://github.com/microsoft/vscode-docs/pull/5718) to use absolute links (as explained [here](https://github.com/microsoft/vscode-docs/blob/c41175e8775cde38ae641153848a6ce26a0dae9c/CONTRIBUTING.md#links) in contribution guidelines). I hope I am not misunderstanding how website build works, but I may be.

gregvanl:
Hi @anthonyisensee Thanks for pointing this out. This seems odd and I had a look at previous versions of the website via the WayBack Machine website and those paths had been working back in 2020. Not sure yet what changed.
Also odd is that some topics have no problem rendering similar video paths. For example, https://code.visualstudio.com/docs/languages/java uses:

<video autoplay loop muted playsinline controls>
  <source src="/docs/languages/java/package-viewer.mp4" type="video/mp4">
</video>

anthonyisensee:
Indeed, that is odd. 

Both

- https://code.visualstudio.com/assets/docs/languages/java/package-viewer.mp4
and
- https://code.visualstudio.com/docs/languages/java/package-viewer.mp4

lead to the same file despite a completely different path. The same behavior also exists with the javascript files.

Now that I'm examining the HTML a bit more closely, there may be a major HTML difference here.

javascript page:

```html
[...]

PR #10136, opened 2026-08-13
Fix broken video links on Java and Python language pages

tha2015:
## Problem

The videos on [Languages / Java](https://code.visualstudio.com/docs/languages/java) and [Languages / Python](https://code.visualstudio.com/docs/languages/python) render as blank players. All six `<source>` elements 404 on the live site, e.g.:

- `https://code.visualstudio.com/docs/languages/java/package-viewer.mp4` → **404**

## Root cause

The pages use absolute `src="/docs/languages/<lang>/*.mp4"` paths. The website build only rewrites media paths that start with `images/` (into `/assets/docs/languages/...`); absolute `/docs/...` paths are shipped unchanged and 404. The video files themselves are present and deployed — they live under `docs/languages/images/<lang>/` and are served at `/assets/docs/languages/<lang>/*.mp4` (200).

This is the same issue as #5717, fixed by #5718 for the JavaScript/TypeScript pages. The Java and Python pages were not included in that fix (Java's videos were later relocated into `images/java/`, which broke the old absolute path). A code search shows these are the only two pages still using the absolute form.

## Fix

Switch the six video `src` values to the relative `images/<lang>/` convention (the same convention the images on these pages already use), so the build rewrites them to the correct `/assets/` URLs.

**`docs/languages/java.md`** (4)
- `package-viewer.mp4`, `Intellisense.mp4` (matches the actual filename casing), `resolve-main.mp4`, `testng.mp4`

**`docs/languages/python.md`** (2)
- `python-linting.mp4`, `python-debugging.mp4`

All six target files are confirmed present under `docs/languages/images/<lang>/`.

Refs #5717, #5718.


tha2015:

ntrogh (approved):
Thanks!

PR #10138, opened 2026-08-16
docs: fix JavaScript parameter inlay hints setting

lurantys:
Fix the JavaScript docs to use the correct `js/ts.inlayHints.parameterNames.enabled` setting.

The old setting name does not exist in VS Code.


ntrogh (approved):
Thanks!

PR #10167, opened 2026-08-25
Added information on how to manually install an extension

realy8v:
These commits are good because devs testing their own VS Code extension can actually try it. There is no reason for this to not be merged.


ntrogh:
Discussed this with engineering team and we choose not to document this approach. Referring to the API docs for extension developers that want to test their extension: https://code.visualstudio.com/api/working-with-extensions/testing-extension

Appreciate your contribution.

realy8v:
Think about the people wanting to install an extension from source because the VS Code marketplace doesn't have it at that moment (On that note, please put Kristal's VS Code extension back up)

